PR #3, opened 2013-04-18
Optimized snowball code for tamil stemmer

rdamodharan:
I have replaced simple `or` conditions with `among()` command as suggested by Martin. I tried to use `among()` which consisted of some `test()` commands inside and they didnt work as expected so have left them as it is. I tested the modification against the data in https://github.com/rdamodharan/snowball-data/tree/master/tamil and the results were same.

### DIFF
--- a/algorithms/tamil/stem_Unicode.sbl
+++ b/algorithms/tamil/stem_Unicode.sbl
@@ -90,8 +90,6 @@ routines (
  has_min_length
 )
 
-integers (mark mark2)
-
 externals ( stem )
 
 groupings (
@@ -107,11 +105,9 @@ groupings (
 )
 
 booleans (
- found_vallinam_doubling
  found_a_match
  found_vetrumai_urupu
  found_wrong_ending
- was_stripped
 )
 
 integers (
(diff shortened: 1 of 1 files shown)


ojwb:
I'm trying to tidy up the licensing situation for snowball (see snowballstem/snowball#10 for more details).

Are you OK to license this patch under the 3-clause BSD licence shown here:

http://opensource.org/licenses/BSD-3-Clause

ojwb:
The cause of the inconsistent results was a variable in the tamil snowball stemmer not being reset between runs (see snowballstem/snowball-data#1 for details).

ojwb:
Currently waiting to hear about licensing.

Also waiting to hear if the altered output when the `found_vetrumai_urupu` variable is consistently cleared before each word is OK - 987 words in the testdata stem differently after this fix, but I am not familiar with Tamil so I can't tell if they're now handled better, worse, or similarly well.

rdamodharan:
@ojwb I am extremely sorry for the delay in the response. 

Regarding the licensing I am fine with licensing the patch under 3-clause BSD. 

I have done some preliminary tests on the change you suggested and it looked fine to me. I will look more into it this weekend and confirm to you.

ojwb:
No need to be sorry for taking time - it took us long enough to track down the issue that was causing the inconsistent stemming results!

Let me know if you've any questions I might be able to help with.

rdamodharan:
I have checked some more outputs and I think the change proposed (unsetting found_vetrumai_urupu) gives more correct results.

ojwb:
Thanks for checking - will get this merged soon.

ojwb:
I rebased f5fc814937e6761c0688611de7b4a8845532c9eb onto the other branch so it was just a whitespace cleanup, and have merged the result.  Thanks.

Was there a reason to keep all the commented-out unused groupings in the code?

rdamodharan:
Thank you Olly. I added it initially thinking I might need it but have not
used it. There is no reason to keep it now.

Damu

ojwb:
Thanks for confirming.  All merged, so closing.

PR #8, opened 2014-10-07
Add cmake build script.

vstakhov:
Here is a simple cmake build script to build `snowball` compiler, `libstemmer` library and the example `stemwords` executable. This build system is tested to be working on Ubuntu Linux.


ojwb:
Having two different build systems is a maintenance headache, so we really need to pick one. Richard says he lacks the time, and I don't really understand cmake, plus we have PR #24 which updates the makefiles but not the cmake build system.

So while your contributions are appreciated, I'm proposing to back them out.  Richard suggested to open a PR for discussion of this, so I've opened #25.

Issue #10, opened 2014-12-01
Insert licence into source code


omron93:
Could you please inset licence into distributed source code?

Because BSD licence requires including its text with all derivative works. So without including a proper full license text it is making difficult or impossible for anyone to comply with desired license terms.

There was discussion on old mailing list - http://thread.gmane.org/gmane.comp.search.snowball/1463

Thanks,
Marek

omron93:
Is this acceptable?

ojwb:
Unfortunately it's more complex than that.  It's something I'd like to resolve, but I've been very busy since taking over snowball maintenance and have not had time to work on it yet.

http://snowball.tartarus.org/license.php says:

```
 All the software given out on this Snowball site is covered by the BSD License (see http://www.opensource.org/licenses/bsd-license.html ), with Copyright (c) 2001, Dr Martin Porter, and (for the Java developments) Copyright (c) 2002, Richard Boulton.
```

There are actually more copyright holders than that (for example, my contributions are substantial enough to be subject to copyright), but people who contributed code to be made available from the Snowball site can probably be assumed to be happy with it being made available under this advertised licence (I am intending to check this assumption is legally valid).

Unfortunately, there's also code contributed to the repos on github directly (for example, the cmake build system PR) which has never been made available from the Snowball website.  Since there's currently no licence specified in the github repo, so we don't even have the implied licence we do for contributions made to the version on the Snowball site.

So we need to check for such contributions, and ask for an explicit license, or else remove them if we can't contact the author, or if they aren't willing or able to grant such a license.

We should also make sure to explicitly ask for a licence grant for new contributions before merging any more non-trivial PRs.

If you want to help move this forwards, checking the PR list and commit log for such contributions and contacting the contributors would be very useful.

omron93:
And what should this "licence grant" consist of? Is there any document contributors should fill,...? (I don't understand similar licensing process much)

stefanor:
As a stopgap, you could maybe say "Copyright Dr Martin Porter, Richard Boulton and contributors"? That's what many projects do, to avoid listing every contributor.

Of course, it'd still be ideal to list every contributor (probably in a separate file, to avoid unnecessary duplication)

ojwb:
The main issue isn't with exhaustively listing contributors or not, but that it isn't clear that all the contributors have actually licensed their contributions under the appropriate licence.

[...]

PR #12, opened 2014-12-16
/usr/env perl instead of /usr/bin/perl

jbaum98:
This fixes an issue where it doesn't work if `perl` is in a nonstandard location.

### DIFF
--- a/libstemmer/mkmodules.pl
+++ b/libstemmer/mkmodules.pl
@@ -1,4 +1,4 @@
-#!/usr/bin/perl -w
+#!/usr/bin/env perl
 use strict;
 use Getopt::Std;
 


ojwb:
I can't seem to reopen this PR for some reason, but the reason I hadn't simply applied it already is that the proposed change disables warnings.  Needs `use warnings;` and ideally a suitable Perl version requirement adding.

jbaum98:
I deleted my copy of the repo and the branch because I thought it was all done, maybe that's the problem? Sorry.

ojwb:
Fixed in 1354a7d.

ojwb:
Hi @jbaum98,

I'm trying to tidy up the licensing situation for snowball (see #10 for more details).

Are you OK to license this patch under the 2-clause BSD licence shown here:

http://opensource.org/licenses/bsd-license.html

jbaum98:
Fine by me. Is there anything I have to do?

ojwb:
No, replying should be enough.  Thanks.

PR #14, opened 2014-12-26
Fix path for snowball-data in travis setup

rboulton:


ojwb:
Sorry - I'd made this fix locally, but failed to push it before the holidays hit.

PR #17, opened 2015-02-08
Support for C# code generation

cesarsouza:
Hello,

I am sending a pull request, if anyone is interested, that adds a new -cs flag to the snowball compiler in order to generate C# files from algorithm definitions.  A companion base library is also available, as well as automatic tests for all languages included in Snowball. The generate code works on Mono and can be build by xbuild.

I have included Linux/Bash scripts to generate code, compile and test against all languages available in the Snowball-data repository.

Please let me know if you would like to know more about anything in this pull request. I can give more examples on how to run, compile and test if necessary!

Best regards,
Cesar


cesarsouza:
Hi Olly,

Thanks for catching up with the Snowball backlog! 

Answering your previous question about the licensing, I am perfectly fine to share my work under the BSD license. I will also update this pull request to solve the recently added conflicts and re-run the tests to check if everything is still working.

Regards,
Cesar

compiler/driver.c
```diff
@@ -149,32 +164,55 @@ static void read_options(struct options * o, int argc, char * argv[]) {
...
+				check_lim(i, argc);
+				o->string_class = argv[i++];
+				continue;
+			}
+			if (eq(s, "-a") || eq(s, "-amongclass")) {
+				check_lim(i, argc);
+				o->among_class = argv[i++];
+				continue;
+			}
 #endif
```
ojwb: Please avoid whitespace changes like these - they make patches harder to review, and make tools like `git blame` less useful.  Currently the codebase is mostly tab-free (which I think should probably be the policy) but where tabs the width is 8.

compiler/generator.c
```diff
@@ -586,7 +586,8 @@ static void generate_do(struct generator * g, struct node * p) {
     if (g->label_used)
         wsetl(g, g->failure_label);
     if (keep_c) {
-        w(g, "~M"); wrestore(g, p, keep_c);
+        w(g, "~M");
+		wrestore(g, p, keep_c);
         w(g, "~N~}");
```
ojwb: This sort of unrelated change isn't helpful either.

Perhaps we should change the code to not have multiple statements per line, but if so that should be done as a separate PR.

csharp/Snowball/Algorithms/DanishStemmer.generated.cs
```diff
@@ -0,0 +1,464 @@
+// This file was generated automatically by the Snowball to C# compiler
```
ojwb: We don't include the generated files in the repo.

ojwb:
I can't comment directly on the added file `compiler/generator_csharp.c` it seems, but the comment it generates in the output still says "Java" rather than "C#".

cesarsouza:
After almost two years...

Thanks for the helpful comments, OIly. I have updated my branch to be compatible with the latest development version, solved the issues you mentioned, and also added proper commands in the makefile to generate the C# stemmers, build a C# version of stemwords and make C# distribution package.

Regards,
Cesar

ojwb:
Sorry for taking a while to review.

`csharp/Unit Tests/Externals/nunit.framework.dll` seems to be a source-less binary which isn't something I want in the tree for multiple reasons.  However, it's also referred to by build files, so presumably just removing it will break stuff.

Are these "Unit Tests" actually useful?  They seem to be redundant with `make check_csharp` (or if they aren't, the testcases they contain could easily be added to the data used by that check.

ojwb:
You seem to have removed the machinery to generate the `copy_from()` method, but any code using `$` on a string variable will generate a call to `copy_from()`, and fail to compile (this feature isn't used by any of the current stemming algorithms).

ojwb:
No feedback for over a year, so I've merged a squashed version of this branch after stripping out the source-less binary and associated files and adding a simple rule to the makefile to build using mono.

The `$` on a string issue still exists, but that's actually broken for several of the existing backends (#62) so blocking the merge on fixing this doesn't seem helpful overall - at least with the current C# backend the user will get a compilation failure for a stemmer that uses this feature, rather than us generating code with incorrect behaviour.

PR #21, opened 2015-06-09
Add license file

omron93:
Add BSD license text - see #10


ojwb:
This is actually the wrong licence text - see http://article.gmane.org/gmane.comp.search.snowball/1558 for why.

I'll add an explicit licence text once I resolve the licensing of contributions since the move to github, but this PR isn't useful so closing it.

PR #22, opened 2015-07-14
Snowball version of Porter stemmer for Lithuanian language

dainiusjocas:
Hello,

I've been working on a Snowball stemmer for Lithuanian language and I'd like to contribute it to a wider community. By contributing my work I hope that community can have some benefit. 

Please let me know if you would like to know more about anything in this pull request. If there are any problem with my code or there are more things to do in order to merge the code, I'm more than willing put effort to fix it.

Best wishes,
Dainius

### DIFF
--- a/algorithms/lithuanian/stem_ISO_8859_1.sbl
+++ b/algorithms/lithuanian/stem_ISO_8859_1.sbl
@@ -1,0 +1,380 @@
+externals ( stem )
+/* Special characters in Unicode Latin-1 and Latin Extended-A */
+// ' nosine
+stringdef a'   decimal '261'  // ą a + ogonek
+stringdef e'   decimal '281'  // ę e + ogonek
+stringdef i'   decimal '303'  // į i + ogonek
+stringdef u'   decimal '371'  // ų u + ogonek
+
+// . taskas
+stringdef e.   decimal '279'  // ė e + dot
+
+// - ilgoji
+stringdef u-   decimal '363'  // ū u + macron
+
+// * varnele
+stringdef c*   decimal '269'  // č c + caron (haček)
+stringdef s*   decimal '353'  // š s + caron (haček)
+stringdef z*   decimal '382'  // ž z + caron (haček)
+
+// [C](VC)^m[V|C]
+// definitions of variables for
+// p1 - position of m = 0
+// p2 - position of m = 1
+integers ( p1 p2 s)
+
+// booleans - to be commented
+// CHANGE
+booleans ( CHANGE )
+
+// escape symbols for substituting lithuanian characters
+stringescapes { }
+
+// groupings
+// v - lithuanian vowels
+groupings ( v )
+
+// v - all lithuanian vowels
+define v 'aeiyou{a'}{e'}{i'}{u'}{e.}{u-}'
+
+// all lithuanian stemmer routines: 4 steps
+routines (
+   step2 R1 step1 fix_chdz fix_gd fix_conflicts
+)
+
+backwardmode(
+
+  define R1 as $p1 <= cursor
+  define step1 as (
+	 setlimit tomark p1 for ([substring]) R1  among(
+      // Daiktavardžiai
+      // I linksniuotė
+      'as' 'ias' 'is' 'ys'        // vyras, kelias, brolis, gaidys
+      'o'    'io'                 // vyro, kelio
+      'ui'   'iui'                // vyrui, keliui
+      '{a'}' 'i{a'}' '{i'}'       // vyrą, kelią, brolį
+      'u'    'iu'                 // vyru, keliu
+      'e'    'yje'                // vyre, kelyje
+      'y'     'au'   'i'          // kely, brolau, broli,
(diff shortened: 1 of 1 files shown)


ojwb:
I don't know Lithuanian, so I can't usefully review the algorithm.  Code-wise, aside from resolving the comment above it looks good to me.

It could really do with a test vocabulary along with the corresponding expected stemmed output to go here:

https://github.com/snowballstem/snowball-data

Also are you happy to license this as 2-clause BSD: http://opensource.org/licenses/bsd-license.html

ojwb:
The snowball compiler issues a couple of warnings:

```
algorithms/lithuanian/stem_Unicode.sbl:49: among(...) has repeated string 'omės'
algorithms/lithuanian/stem_Unicode.sbl:49: among(...) has repeated string 'otės'
```

The line number reference is just to the start of the `among`, but looking through that it's right:

```
      'ome' 'omės'                // mokome, bijomės
```

vs:

```
      'om{e.}s'                   // mokydavomės
```

And:

```
      'ot' 'ote' 'otės'           // mokot, mokote, bijotės
```

vs:

```
      'ot{e.}s'                   // tikėjotės
```

Since `{e.}` is a stringdef for `ė`:

```
stringdef e.   decimal '279'  // ė e + dot
```

algorithms/lithuanian/stem_ISO_8859_1.sbl
```diff
@@ -0,0 +1,380 @@
...
+
+// [C](VC)^m[V|C]
+// definitions of variables for
+// p1 - position of m = 0
+// p2 - position of m = 1
+integers ( p1 p2 s)
+
+// booleans - to be commented
+// CHANGE
+booleans ( CHANGE )
```
ojwb: Although `CHANGE` is set in several places, its value is never actually used anywhere.

The snowball compiler ought to be able to detect such cases and issue a warning, but it doesn't currently.

ojwb:
@dainiusjocas Ping?

Clarifying the licensing is the key thing, as without that we can't merge the code.

It would be good to resolve the other points, but if the licensing is straight, somebody else could potentially do that.  As things are, nobody else can usefully help at all.

dainiusjocas:
Hi,

I have no problems with licensing this as 3-clause BSD.

dainiusjocas:
This issue with repeated strings should be solved by using {e.} instead of ė everywhere. I was just going through the grammar book and collecting all variations of word endings. My mistake.

dainiusjocas:
Also, I was using a Lithuanian vocabulary, but it is of a relatively poor quality. Therefore, I'm not entirely sure if it is wise to publish it among other vocabularies.

ojwb:
Thanks for responding.

Is the `CHANGE` variable something that should be used, or did you decide it wasn't useful after all?

We really want some sort of wordlist, to give us confidence that the algorithm performs as intended, and continues to do so as the compiler evolves.  The quality isn't critical, but it's obviously most useful to test with common words in the language, since those are the most important cases in actual use.

There are some CC-BY-SA 3.0 licensed wordlists here, including one for Latvian:

https://invokeit.wordpress.com/frequency-word-lists/

Can you take a look and see if that's better quality?

ojwb:
Rather than stalling this indefinitely, I'm going to assume that `CHANGE` can just be stripped out, and that the word list linked to above is suitable (or if it's clearly not, I'll create one from a dump from the Lithuanian version of wikipedia).

ojwb:
That "invokeit" wordlist is rather poor quality - AIUI it's generated from subtitles, and seems to contain a lot of English words.

I'll try making a list from wikipedia.  I did this for Tamil, but that was easier as the alphabet is non-Latin so we can ignore all the wiki markup by only selecting words made from character in the appropriate Unicode range - for languages using a Latin alphabet we need to actually parse the markup, but there are existing tools to so that's OK.

I've also mostly written a new warning for "set but not used" that would have caught the `CHANGE` boolean never being read (it only catches cases where a variable is never read, rather than spotting dead stores individually, but it's better than never warning).  This new warning also reports that `p2` is set but not used, which indeed seems to be the case!  Is that intentional?

algorithms/lithuanian/stem_ISO_8859_1.sbl
```diff
@@ -0,0 +1,380 @@
...
+    )
+  )
+
+)
+
+define stem as (
+
+  $p1 = limit
+  $p2 = limit
+  $s = size
```
ojwb: `size` measures in Unicode characters for single byte encodings and wide characters, but in bytes for UTF-8.  This unfortunately means that the stemmer's behaviour is different for UTF-8 and non-UTF-8 versions.  To address this issue, we added `len` which always gives the length in Unicode characters, but that happened after you opened this PR so that wasn't available when you wrote this code.  I'm proposing to change this line to `$s = len` to get consistent behaviour regardless of the encoding in use.

ojwb:
I've opened PR #76 which updates this patch, hooks it into the build system, and fixes the various issues noted above, so closing this PR in favour of that one.

PR #24, opened 2015-07-29
Add Python generator

mitya57:
I am not the original author of these generators; @shibukawa is. However I needed to use the Python generator, so I have fixed some compilation errors in the code, rebased it against the current snowball code, and submitted this pull request.

The original code by Yoshiki can be found at [shibukawa/snowball](https://github.com/shibukawa/snowball) (in `snowball` directory).

Python needs no introduction. <s>JSX is a statically typed JavaScript-like language that compiles to JavaScript, the compiler can be found at [jsx/JSX](https://github.com/jsx/JSX/) (requires node.js).</s>

Update: I excluded the JSX generator from this pull request

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,3 +13,4 @@ before_install:
   - git clone --depth=1 -b "$TRAVIS_BRANCH" https://github.com:"${TRAVIS_REPO_SLUG%%/*}"/snowball-data.git || git clone --depth=1 -b "$TRAVIS_BRANCH" https://github.com/snowballstem/snowball-data.git || git clone --depth=1 https://github.com/snowballstem/snowball-data.git
 script:
   - make && make check STEMMING_DATA=snowball-data
+  - make check_python python=python2.7 STEMMING_DATA=snowball-data
--- a/doc/libstemmer_python_README
+++ b/doc/libstemmer_python_README
@@ -1,0 +1,93 @@
+Snowball stemming library collection for Python
+===============================================
+
+How to use library
+------------------
+
+The ``snowballstemmer`` module has two functions.
+
+The ``snowballstemmer.algorithms`` function returns a list of available algorithm names.
+
+The ``snowballstemmer.stemmer`` function accepts algorithm name and returns ``Stemmer`` objects.
+
+``Stemmer`` objects have ``Stemmer.stemWord(word)`` method and ``Stemmer.stemWords(word[])`` method.
+
+.. code-block:: python
+
+   import snowballstemmer
+
+   stemmer = snowballstemmer.stemmer('english');
+   print(stemmer.stemWords("We are the world".split()));
+
+``Stemmer`` objects have ``Stemmer.maxCacheSize`` property. They cache result within the value. Default is ``10000``.
+
+Accelerates Stemming
+--------------------
+
+if **PyStemmer** is installed, ``snowballstemmer.stemmer`` returns ``PyStemmer``\ 's ``Stemmer`` objects. This ``Stemmer`` object has same methods (``Stemmer.stemWord()``, ``Stemmer.stemWords()``).
+
+**PyStemmer** is a Snowball's ``libstemmer_c`` wrapper module and it returns 100% compatible result with **snowballstemmer**.
+
+**PyStemmer** has faster speed because it uses C-lang module, and **snowballstemmer** has higher usability because it is pure Python module.
+
+* `PyStemmer <http://pypi.python.org/pypi/PyStemmer/>`_
+
+Benchmark
+~~~~~~~~~
+
+Test Case: Snowball stemmer check data (16 algorithms, total 582560 words, cache hit 0%)
+Computer: MacBook Pro 3rd Gen Corei7 2.3GHz
+
+* Python 2.7 + **snowballstemmer** : 2m 30s
+* PyPy 1.9 + **snowballstemmer** : 45s
+* Python 2.7 + **PyStemmer** : 5s
+
+This test case is much harder than usual usecases!
+
+The TestApp example
+-------------------
+
+The ``testapp.py`` example program allows you to run any of the stemmers
+on a sample vocabulary.
+
+Usage::
+
+   testapp.py <algorithm> "sentences ... "
+
+.. code-block:: bash
+
+   $ python testapp.py English "sentences... "
+
+License
+-------
+
+It is a BSD licensed library.
+
+Copyright (c) 2013, Yoshiki Shibukawa
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+   Redistributions of source code must retain the above copyright notice,
+   this list of conditions and the following disclaimer.
+
+   Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+
+   Neither the name of the <ORGANIZATION> nor the names of its contributors
+   may be used to endorse or promote products derived from this software
+   without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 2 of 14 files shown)


mitya57:
@stefanor I would like you to also review this as you maintain snowball in Debian.

With `make dist_libstemmer_python` you get a tarball, after unpacking which you can use `setup.py` to build/install the Python module. If you remove the last line in `dist_libstemmer_python` code in GNUmakefile, then you'll get the unpacked directory without the need to deal with tarball.

I would like to see `python-snowballstemmer` and `python3-snowballstemmer` Debian packages built this way. I need these packages because they are dependencies for Sphinx 1.3.

Note that the Python part has a different version number, I have not changed this to not break reverse dependencies which may expect it to be as it is.

mitya57:
@ojwb Ping?

mitya57:
@ojwb To simplify things a bit, I removed JSX stuff from this pull request. For the Python part, I addressed all your comments.

If someone (me?) needs the JSX part in the future, I will be happy to propose another pull request to add it.

.travis.yml
```diff
@@ -1,4 +1,9 @@
-language: c
-compiler: gcc
```
ojwb: I don't know much about travis, but don't we still want to specify the compiler for C code?

doc/libstemmer_python_README
```diff
@@ -0,0 +1,84 @@
...
+Copyright (c) 2013, Yoshiki Shibukawa
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided
+that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this list of conditions and
+  the following disclaimer.
+* Redistributions in binary form must reproduce the above copyright notice, this list of conditions
+  and the following disclaimer in the documentation and/or other materials provided with the distribution.
```
ojwb: I believe BSD 3-clause is the correct licence here - for the gory details see:
http://article.gmane.org/gmane.comp.search.snowball/1558

I'm guessing @shibukawa just took this from the linked OSI page after OSI had unhelpfully changed the content...

python/setup.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python
+
+from distutils.core import setup
+
+setup(name='snowballstemmer',
+      version='1.1.0',
+      description='This package provides 16 stemmer algorithms (15 + Poerter English stemmer) generated from Snowball algorithms.',
```
ojwb: We should take care to spell Martin's surname correctly - it's "Porter".

python/stemwords.py
```diff
@@ -0,0 +1,103 @@
...
+import re
+import codecs
+import snowballstemmer
+
+def usage():
+    print('''usage: %s [-l <language>] [-i <input file>] [-o <output file>] [-c <character encoding>] [-p[2]] [-h]
+
+The input file consists of a list of words to be stemmed, one per
+line. Words should be in lower case, but (for English) A-Z letters
+are mapped to their a-z equivalents anyway. If omitted, stdin is
```
ojwb: I don't see any code below to lowercase the input word (as there is in the C version).

doc/libstemmer_python_README
```diff
@@ -0,0 +1,84 @@
+Snowball stemming library collection for Python
+===============================================
+
+How to use library
+------------------
+
+The ``snowballstemmer`` module has two functions.
+
+The ``snowballstemmer.algorithms`` function returns a list of avilable algorithm name' string.
```
ojwb: `avilable` -> `available`.

Also I think some text has been lost or damaged around `name' string` - I guess it's trying to say something like `returns a list of available algorithm names.`

ojwb:
> To simplify things a bit, I removed JSX stuff from this pull request.

It will probably be less painful to do two separate merges, so that seems a good plan, though it's annoying github seems to have now just thrown away my comments on that code.  It really does have a terrible patch reviewing workflow.

mitya57:
I will look at your last comments a bit later, just…

> It will probably be less painful to do two separate merges, so that seems a good plan, though it's annoying github seems to have now just thrown away my comments on that code. It really does have a terrible patch reviewing workflow.

They aren't thrown away. Your first batch of comments is still available on mitya57@fb158c6aef6ae606.

.travis.yml
mitya57: Unfortunately Travis doesn't seem to support different `language`s. So you can either specify multiple compliers or multiple Python versions.

I wanted to test with both Python 2.7 and Python ≥ 3.3, but now as you committed 30e10d914ff10910 to specify multiple compilers, I reverted my change and it tests only Python 2.7 now.

If in future Travis moves to a newer OS that has Python ≥ 3.3 (Ubuntu 12.04 doesn't), we will be able to install it using `apt-get` and test with it.

doc/libstemmer_python_README
mitya57: As BSD-3-clause is stricter that BSD-2-clause, we can probably relicense it without asking Shibukawa (whom I attempted to contact multiple times and failed at that).

mitya57:
Updated the pull request, and made the Travis build finally pass.

.travis.yml
ojwb: On 5 September 2015 at 02:33, Dmitry Shachnev wrote:
> > @@ -1,4 +1,9 @@
> > -language: c
> > -compiler: gcc
>
> Unfortunately Travis doesn't seem to support different languages. So you can either specify multiple compliers or multiple Python versions.
>
> I wanted to test with both Python 2.7 and Python ≥ 3.3, but now as you committed 30e10d9 to specify multiple compilers, I reverted my change and it tests only Python 2.7 now.
>
> If in future Travis moves to a newer OS that has Python ≥ 3.3 (Ubuntu 12.04 doesn't), we will be able to install it using apt-get and test with it.

You should be able to achieve that using multiple entries for the same
compiler with different `env:` settings in each.

Something like this (inspired by swig's `.travis.yml`):

```
compiler:
  - clang
  - gcc
    env: PYTHON=python2.7
  - gcc
    env: PYTHON=python3.2
  - gcc
    env: PYTHON=python3.3
```

Cheers,
    Olly

ojwb:
I've set up builds with Python 2.7 and 3 in 8189e60, but the I couldn't seem to access your original commit adding the different builds to see which python interpreters you specified.

The default travis image doesn't see to have `python3` so that build currently fails - I guess we need to install it with apt if we aren't using the python-specific travis builder (that is what SWIG seems to do).

Could you let me know which Python versions you specified before (or send a PR to update `.travis.yml` suitably)?

ojwb:
> They aren't thrown away. Your first batch of comments is still available on mitya57@fb158c6.

Curiously they are now shown as "ojwb commented on an outdated diff" here, which I'm sure they weren't before.  Anyway, good that they are still accessible.

mitya57:
> The default travis image doesn't see to have python3 so that build currently fails - I guess we need to install it with apt if we aren't using the python-specific travis builder (that is what SWIG seems to do).
>
> Could you let me know which Python versions you specified before (or send a PR to update .travis.yml suitably)?

Travis uses Ubuntu 12.04 where 2.7 is available out-of-the-box and 3.2 is in the archive. This code doesn't work with 3.2 and works only with ≥ 3.3, so we can't use apt-get for now.

When one specifies `language: python` and `python: [version-list]`, then the needed Python versions are installed not via apt-get, but via some hidden internal methods, which are not available to us when `language` is `c`.

ojwb:
Hmm, what a pain.

They really should have just put the extra python versions in a package repo and then they'd be trivial to mix and match.  Having to pick a language-specific VM base image only really works for single-language projects.

This will only get worse as more backends are added to Snowball - there's Javascript and C# on the cards.  Not sure I see a good way to handle this, short of having branches with a different `.travis.yml`.

mitya57:
> This will only get worse as more backends are added to Snowball - there's Javascript and C# on the cards. Not sure I see a good way to handle this, short of having branches with a different .travis.yml.

If it will support versions of node.js and mono in Ubuntu 12.04, then it'll be possible to test it after installing them via apt-get.

By javascript, do you mean the JSX generator, or something else? That said, do you want me to propose the pull request for JSX (I don't need it much myself)?

ojwb:
> If it will support versions of node.js and mono in Ubuntu 12.04, then it'll be possible to test it after installing them via apt-get.

The Python bindings work with versions of Python in Ubuntu 12.04, but we can't test with all the versions we'd like to (and which we could in a travis python VM image).

I gather node.js is a pretty fast moving target, so it seems likely the node.js travis VM image has custom installed newer versions of node and/or related software which we can't simply install onto a different image with apt-get.  We might be able to test with the packaged versions, but that isn't ideal.

To handle Python and C, maybe starting from the Python image and installing the relevant packages for C compilation would work, but as soon as we want the custom-installed versions from more than one VM image that trick stops working.

I'm not sure it's worth putting energy into working around these deficiencies in travis - hopefully they'll move on to a newer Linux version soon and this won't be such an issue.

I think it would be useful to get the JSX generator merged too.  The main issue seemed to be that the patch contains a hash function apparently taken from a random forum posting which lacked any licensing information, but simply replacing that is clearly a feasible fix if it comes to it.

ojwb:
travis-ci now have trusty available in beta, and there they've scrapped the array of different images - there's just a minimal one, and one with everything.  So I've updated to use that and now we're testing with python2, python3, and pypy (since that was documented as working).

mitya57:
@ojwb: Thanks, and nice, it even works with Python 2.6 and PyPy!

PR #25, opened 2015-09-01
Remove the cmake build system

ojwb:
Having two different build systems is a maintenance headache, so we
really need to pick one.  Richard says he lacks the time, and I don't
understand cmake, plus we have PR #24 which updates the makefiles but
not the cmake build system.


vstakhov:
As I've stated previously, I've suggested to add cmake system to integrate snowball with `rspamd` spam filtering system which uses cmake so far. In general, I can do the necessary updates if needed. Or you can obviously remove cmake support completely if you'd like (and I'll continue to use my fork).

ojwb:
Sorry, I missed the motivation behind this change in the previous discussion.

Adding support for every build system out there so people can embed snowball into their projects just isn't going to scale, even if there's someone willing to maintain each of them.  I think we should really be working towards making snowball more usable as an installed dependency, rather than making it easier for people to embed a copy of it in their projects.

If we have two build systems in the main repo, we're going to have to find out which people are using each time there's a report of a bug report where that might be relevant.  Sometimes they'll actually think to tell us, or it'll be obvious from the error, but experience tells me we'll have to ask a lot of the time.

So if you really want to be able to embed snowball as you are, I think it's more sensible to maintain that separately.  It's likely to be less work for you that way - e.g. you don't need to add support for the jsx or python backends (#24) if you aren't using them in rspamd.  It also means that your "snowball with cmake" tree can always be in a consistent state, which is more helpful to anyone else wanting to use it.  If cmake support is in the main repo, then it'll be out of step for a bit after any change to the makefiles.

PR #26, opened 2015-09-11
Add JSX generator and fix other minor issues

mitya57:
As discussed with @ojwb in #24, here is the JSX generator for snowball.

I know almost nothing about JSX, so I couldn't even test it properly. However, [Sphinx](https://github.com/sphinx-doc/sphinx) already uses JS files produced with this generator and JSX compiler, and it seems to work fine.

While I was at it, I also fixed some minor issues. You can safely pull 083dab56 into master, even if you don't yet want to merge the JSX generator.


ojwb:
Thanks.  I've merged the first two commits, and am looking at the final one.

compiler/generator_jsx.c
```diff
@@ -0,0 +1,1495 @@
...
+    g->S[0] = p->mode == m_forward ? "" : "_b";
+    g->I[0] = SIZE(b);
+    g->L[0] = b;
+    write_failure_if(g, "!(this.eq_s~S0(~I0, ~L0))", p);
+}
+
+static void generate_define(struct generator * g, struct node * p) {
+
+    struct name * q = p->name;
+    symbol stem[] = {'s', 't', 'e', 'm'};
```
ojwb: This seems to be special casing functions named `stem`, which is clearly not right.  It seems to mark them as `override`.  Not entirely sure what this is all for, but probably it should instead check if the function is `external`.

ojwb:
I've tried changing all of those and managed to get the jsx compiler installed, and the tests pass.  Will push a branch shortly.

I'm not entirely sold on generating JSX being the right route to generating Javascript from Snowball, particularly as the JSX compiler doesn't seem very actively maintained (last commits over a year ago and even then the last ones were just merging PRs for typos).

It seems to me that Snowball itself could just generate code with the optimisations that going via JSX could give us already applied, which would be rather simpler to use.

Plus the name 'JSX' is also used by the rather more active React project (http://facebook.github.io/react/) so we're bound to end up with confused questions about it.

But I'm happy to get this merged and then see where we go from there.

mitya57:
> I've tried changing all of those and managed to get the jsx compiler installed, and the tests pass. Will push a branch shortly.

Thanks!

> I'm not entirely sold on generating JSX being the right route to generating Javascript from Snowball, particularly as the JSX compiler doesn't seem very actively maintained (last commits over a year ago and even then the last ones were just merging PRs for typos).

I share your concerts. JSX doesn't seem to be popular amongst JS developers, too. As I said, I don't need this particular code myself (unlike the Python one), so it's up to you to decide whether it is something wanted.

ojwb:
> it's up to you to decide whether it is something wanted.

I think it's worth merging - the JSX language is pretty similar to Javascript, so it shouldn't be hard to adjust the JSX backend to generate Javascript directly - much simpler than starting from scratch anyway.

And being able to generate stemmers in Javascript seems useful.  It would be a neat way to implement the stemming demo on the website too.

ojwb:
OK, I've merged the JSX backend.  I squashed a change on top to discard some Java-specific things, including the `hash()` function for which the licensing wasn't clear, so we avoid having that code in the git history.

Also committed a few small improvements to it.  Thanks for all your work on this.

mitya57:
Not that it was my work, so thanks go to you :)

PR #27, opened 2015-09-14
Added Java stemmers

magic890:
Added Java stemmers: Danish, Dutch, English, Finnish, French, German, Hungarian, Italian, Norwegian, Porter, Portuguese, Romanian, Russian, Spanish, Swedish, Turkish.

Here an [example](http://stackoverflow.com/a/13381739/2211721) of their usages.


ojwb:
You appear to be proposing to commit the generated output of the snowball compiler - I don't think it's useful to have these files under version control in addition to the sources for them and the sources for the compiler to generate them.

Or am I missing something?

magic890:
Sorry, my fault. However it could be nice to write a README on how to compile and do the other tasks.

PR #29, opened 2015-11-25
Fix Java generation

dscorbett:
In generator_java.c: `eq_s` and `eq_s_b` take the literal string, not the string’s length.

In SnowballProgram.java: `CharSequence#length` is a method.


ojwb:
Thanks - I made that change on a machine without Java installed, and kept an eye on the travis autobuild to make sure it worked.  I'd forgotten we haven't yet hooked up the Java tests in the travis build...

PR #32, opened 2016-01-12
Arabic Light Stemmer

assem-ch:
Based on  the algorithm described here http://ojuba.org/wiki/thawab/stemming, which is inspired from the research paper: http://www.informatik.uni-trier.de/~ley/db/conf/sigir/sigir2002.html#LarkeyBC02. 

We added many other tweaks to the algorithm and we included normalization before stemming. 

Also we suggested a fix for the size function inside snowball in order to calculate the size of unicode strings rightfully.


ojwb:
I've added `len` and `lenof` to snowball git master.

assem-ch:
I tested the new `len`, it seems that `len` is above expected lenght with `+1`.

To check:
`ثثة` should give `ثثة`
`ثثثة` should give `ثثث`

Those are not real arabic words but they are good for testing length since they obey only one rule with size >= 4, last letter will be stripped.

ojwb:
Hmm, it seems that `SIZE(z->p)` reports one too high too, so we scan an extra byte.  Thanks for spotting - investigating...

ojwb:
Fixed that - the underlying bug is actually a very long standing one, but none of the existing stemmers tickle it.

assem-ch:
I tested it , it works

GibreelAbdullah:
Shouldn't the algorithm strip the all the vowel signs?

I tried كَتَبَ in the demo page of https://snowballstem.org/demo.html#Arabic, but it returns كَتَب (stripping only the last vowel sign). Is this intended?

assem-ch:
@GibreelAbdullah we do only necessary normalization, the ones needed for stemming, you are expected to do normalization before stemming

PR #33, opened 2016-02-18
Add Arabic to languages supported in setup.py of python dist package

assem-ch:


ojwb:
The CI failure seems to be a timeout unrelated to this change, so merging.

Thanks.

PR #37, opened 2016-03-08
Add a Makefile for windows, old English

tofutim:
A few things I would like to contribute. Thanks!

Tim

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 /dist
 /jsx_out
 /python_out
+
+# mac files
+.DS_Store
--- a/Makefile.msc
+++ b/Makefile.msc
@@ -1,0 +1,75 @@
+# Makefile for snowball using Microsoft (Visual) C
+
+# The toplevel directory of the source tree.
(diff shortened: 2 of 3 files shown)


ojwb:
mitya57 is right on both points.

We definitely don't want to change the English stemmer in this way, though adding the modified stemmer as a variant would be reasonable I think (c.f. german and german2).

It's also unhelpful to have mixed 3 essentially unrelated changes in one PR.

FWIW, I'm not terribly enthusiastic about adding multiple build systems.  I've seen this tried on other projects, and it doesn't play out well.  Unless someone is really committed to keeping the extra ones up to date, all but the main one will inevitably bitrot.  Or else you require contributors to update all of them and add an extra obstacle to contributing (especially true when one build system can only be tested with proprietary tools).  The last time this came up, I suggested the additional build system be maintained in a forked repo, which allows its maintainer to merge in changes and update the build system for them on his or her own schedule.

Closing this PR, but feel free to submit new PRs.

Issue #34, opened 2016-02-21
Snowball as a shared library


Sufrostico:
Hi there,

Is there a way to compile snowball to a shared library?

Generating something like libsnowball.so ? or libsnowball.dll?

Sufrostico:
Hi there, any news on this?

ojwb:
Please have patience, it's barely been a week since you opened this.

Sufrostico:
oks, thanks for the answer

mitya57:
FWIW Debian also ships a patch to use shared library.

Sufrostico:
I'll take a look into that

Sufrostico:
@mitya57 can you provide a link, can't find it in alioth

mitya57:
@Sufrostico It is [here](http://anonscm.debian.org/cgit/collab-maint/snowball.git/plain/debian/patches/shared-library.diff). It doesn't apply cleanly to the latest snowball git, but in September I have made an attempt to rebase it against the current git master. My attempt is [here](http://anonscm.debian.org/cgit/users/mitya57/snowball.git/plain/debian/patches/shared-library.diff) — maybe not thoroughly tested.

Sufrostico:
@mitya57  Nice!

I also made [one](https://github.com/snowballstem/snowball/pull/35) a few days ago, but yours looks way better.

ojwb:
There are really two things that need doing to have snowball as a good quality shared library.

The obvious one is to make the build system build a shared library.

The other (which hasn't been addressed here) is that the C code snowball currently generates contains a lot of relocations due to the many arrays of string literals.  These result in a load time penalty (as the dynamic loader has to perform these relocations), and a memory overhead (as the pages containing these relocations can't be shared between different processes which have the snowball library loaded).  And the problem gets worse for each new stemmer that is contributed.

For more details of these relocations, see section 2.4.3 of this document:

https://software.intel.com/sites/default/files/m/a/1/e/dsohowto.pdf

I actually solved this in a somewhat primitive but effective way many years ago for the fork of snowball we use in Xapian (the reason why we ended up with a fork was that it was sometimes hard to get Martin to accept patches, especially as he was often very busy with other things).

Since I took over as maintainer I've been gradually merging changes from the fork back in to the main tree, with the hope that eventually Xapian could just use upstream snowball, but this one is still to do.

emirotin:
Looks like this is related: https://github.com/abiliojr/fts5-snowball/issues/4
[...]

PR #42, opened 2016-06-30
Change makefile to support shared library compilation

Sufrostico:
Change the GNUmakefile to support the creation of a shared library based on the
patch by Dmitry Shachnev (mitya57) in his debian package
http://anonscm.debian.org/cgit/users/mitya57/snowball.git/

Comments on the issue https://github.com/snowballstem/snowball/issues/34


ojwb:
We have both this and #160 so closing this (slightly arbitrarily, but #160 is much newer so presumably has fewer conflicts).

PR #45, opened 2016-08-05
Strip diaeresis from "gü" and "guë" in French

dscorbett:
`residual_suffix` was broken: deleting "ë" after "gu" did not work, so "aiguë" was stemmed to "aiguë" instead of the intended "aigu". While I was at it, I normalized "gü" to "gu" before a vowel, as in "aigüe" and "ambigüité".

### DIFF
--- a/algorithms/french/stem_ISO_8859_1.sbl
+++ b/algorithms/french/stem_ISO_8859_1.sbl
@@ -30,6 +30,7 @@ stringdef e`   hex 'E8'  // e-grave
 stringdef i"   hex 'EF'  // i-diaeresis
 stringdef i^   hex 'EE'  // i-circumflex
 stringdef o^   hex 'F4'  // o-circumflex
+stringdef u"   hex 'FC'  // u-diaeresis
 stringdef u^   hex 'FB'  // u-circumflex
 stringdef u`   hex 'F9'  // u-grave
 
@@ -45,6 +46,8 @@ define prelude as repeat goto (
     (  ['y'] v <- 'Y' )
     or
     (  'q' ['u'] <- 'U' )
+    or
+    ( 'g' ['{u"}'] v <- 'u' )
 )
 
 define mark_regions as (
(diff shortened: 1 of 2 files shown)


ojwb:
The current behaviour for `aiguë` matches the description of the algorithm (https://snowballstem.org/algorithms/french/stemmer.html) so I think "intended" isn't quite the right word.  Neither of these are due to bugs in the implementation of the algorithm, but rather arguable shortcomings of the algorithm itself.

Considering the "guë" change first, moving that test up to where you have has two effects:

* The test is no longer limited to the `RV` region (_"If the word begins with two vowels, RV is the region after the third letter, otherwise the region after the first vowel not at the beginning of the word, or the end of the word if these positions cannot be found. (Exceptionally, par, col or tap, at the beginning of a word is also taken to define RV as the region to their right.)_").  In the case of "aiguë" this indeed seems desirable, but careful checking is needed with a large French word list to ensure that it doesn't harm other cases.  Overstemming (mapping two non-equivalent words the same output) is much more problematic than understemming (failing to map two equivalent words to the same output) for the intended domain of use of these stemmers (Information Retrieval).
* The other replacements in "Step 4" which are still confined to `RV` will now be considered even if the "guë" -> "gu" replacement happens.  Unless I'm missing something, none of them can match as the string now ends "gu", but we could optimise by avoiding trying them at all - e.g. `( 'gu' ['{e"}'] delete ) or setlimit tomark pV for ( /* ... */ )`.

As for the "gü" to "gu" change, is that needed due to the 1990 orthography change mentioned here https://en.wikipedia.org/wiki/French_orthography#Diacritics (_"the diaeresis in words containing guë (such as aiguë or ciguë) may be moved onto the u: aigüe, cigüe, and by analogy may be used in verbs such as j'argüe"_)?  If so, what's the logic in applying it for any following vowel rather than just to "güe"?  And why change it to "gue" rather than to "guë"?

Again this change really needs to be tested against a large word list to ensure it doesn't harm other cases.

We also need additions to the French test vocabulary which exercise these changes (the first CI build passed which means nothing in the current vocabulary is affected by these changes; the others failed but only because the Python stemmers are very slow and take more than 10 minutes for the complex arabic stemmer and its very large wordlist, even when only testing 1/3 of the list).

Also the documentation of the algorithm needs updating to reflect the changes (ideally in such a way that it's clear what parts are additions/changes and when they were made).

dscorbett:
Superseded by #78.

PR #48, opened 2016-11-01
add my Irish stemmer

jimregan:


ojwb:
Do you have a suitable test vocabulary which we can add to snowball-data so there's test coverage for this?

ojwb:
Also, "irish" should be added before "italian" to preserve the sorted order of the lists.

ojwb:
Thanks for the test data PR - I've merged that and forced a rerun of the CI jobs which were failing due to lack of corresponding test data.

Most tests now pass, but the JSX tests fail: 
https://travis-ci.org/snowballstem/snowball/jobs/226897553#L646

That's presumably a bug in the JSX backend or support code, but it's kind of a blocker to actually merging this.  I'll take a look when I get a chance, but if someone else wants to prod at it, go for it.

jimregan:
From the diff, it looks like the English stemmer is being run

ojwb:
> From the diff, it looks like the English stemmer is being run

Good spot - `stemwords.jsx` hard codes all the known stemmers, and if
the language specified by `-l` is unknown, silently just uses the
English stemmer.

I'll fix both the hard coding and the silent fall back to English.

ojwb:
Merged by hand.  Thanks for you work and your patience.

I couldn't see how to exit with an error in JSX, so I just added Irish to its hard-coded list for now.

It looks like this JSX (not to be confused with the other rather more well known other JSX) is dead (https://github.com/jsx/JSX/issues/336) so I my plan is to replace support for it with native Javascript support (which I already have mostly working, based on the JSX support - this is how the live demo on the website works).

jimregan:
Cool, thanks

jimregan:
Ok, I see it in 5137019d68befd633ce8b1cd48065f41e77ed43e -- while I'm at it, I'd also contributed a conversion of a Czech stemmer, and I should have a Slovak stemmer lying around somewhere. I'll wait for the JSX stuff to resolve, but would there be an interest in having them?

ojwb:
@jimregan I don't seem to be finding time to make much progress with the JSX stuff recently - feel free to open PRs if you have more stemmers to add.

jimregan:
Ok, cool

ojwb:
@jimregan I've now removed the JSX backend and there's a Javascript backend in its place.  I've also cleaned things up so there's now just a single file (`libstemmer/modules.txt`) to have to update to add a new stemmer.

I can just add the Czech stemmer from the website (unless you have a newer version?)  I now have a script to generate a wordlist from a wikipedia dump I can use unless you have a better wordlist already.

And please do contribute the Slovak stemmer (and any others you might have!)

PR #49, opened 2016-11-17
[Italian] "divano" overstemmed

cesco69:
this pull request fix the issue with word divani and similar *ani suffix words.

An example is "divano" is stemmed correctly to "div", but it's pluaral "divani" is stemmed in "divan"

### DIFF
--- a/algorithms/italian/stem_ISO_8859_1.sbl
+++ b/algorithms/italian/stem_ISO_8859_1.sbl
@@ -147,7 +147,7 @@ backwardmode (
 
     define verb_suffix as setlimit tomark pV for (
         [substring] among(
-            'ammo' 'ando' 'ano' 'are' 'arono' 'asse' 'assero' 'assi'
+            'ammo' 'ando' 'ano' 'ani' 'are' 'arono' 'asse' 'assero' 'assi'
             'assimo' 'ata' 'ate' 'ati' 'ato' 'ava' 'avamo' 'avano' 'avate'
             'avi' 'avo' 'emmo' 'enda' 'ende' 'endi' 'endo' 'er{a`}' 'erai'
             'eranno' 'ere' 'erebbe' 'erebbero' 'erei' 'eremmo' 'eremo'


ojwb:
The proposed patch changes the results for quite a lot of words in the Italian test vocabulary (in the snowball data module):

https://travis-ci.org/snowballstem/snowball/jobs/176628305#L385

I don't really know Italian, so I can't easily judge if those changes are improvements (or at least harmless), or if this change results in the stemmer conflating cases which it shouldn't.

Note that overstemming (mapping two non-equivalent words the same output) is much more problematic than understemming (failing to map two equivalent words to the same output) for the intended domain of use of these stemmers (Information Retrieval).

cesco69:
> The proposed patch changes the results for quite a lot of words in the Italian

`ani` is a suffix for many italian words with the suffix `ano` and is used for the pluaral version, eg.:

```
mano -> mani
nano -> nani
vano -> vani
brano -> brani
piano -> piani
umano -> umani
arcano -> arcani
```

and so many others... is normal if changes the results a lot of words in the Italian test vocabulary, but i can't judge if my PR is totally right, i'm italian but not a expert in italian language.

ojwb:
It would be good to resolve this, but as I already said I unfortunately don't know much about Italian.  However, I've been looking into it as best I can.

Looking at the changes to how the vocabulary stems, the proposed change seems problematic.  For example all of the following currently stem to `allontan`:

```
allontana
allontanando
allontanandoli
allontanandosi
allontanar
allontanare
allontanarlo
allontanarne
allontanarono
allontanarsi
allontanasse
allontanata
allontanati
allontanato
allontanava
allontanavano
allontani
allontanò
```

But with the proposed change, `allontani` would instead stem to `allont`.

There are more cases that look to be harmed in a similar way, but also cases which are probably helped in addition to `divano`/`divani`.  And some I really can't tell.

The examples given in the latest comment above as motivation all actually stem to the same form as each other already (the -o or -i suffix is removed by the stemmer), so don't really make any sort of case for this change.

> An example is "divano" is stemmed correctly to "div", but it's pluaral "divani" is stemmed in "divan"

Both "divano" and "divani" seem to be nouns, and it looks to me like the real issue is that the noun `divano` gets mistaken for a verb with suffix `-ano` which gets removed.  I think `divan` is probably a more appropriate stem than `div`, and adding a new verb ending `-ani` is really not the right answer, since the problem is with nouns not verbs, and `-ani` doesn't actually appear to be an Italian verb ending.

There does seem to be a flaw here, but the proposed solution doesn't seem the appropriate way to address it.

ojwb:
> I think `divan` is probably a more appropriate stem than `div`

To back this up, `diva` and `dive` both also stem to `div` but translate to "diva" in English, whereas "divano" translates to "sofa".

PR #50, opened 2016-12-22
Change algorithm to not remove punctuation marks

assem-ch:


ojwb:
Rebased and pushed as a75f87cc7775af7de66a81c7d4590dbc0ef91358.

Thanks, and sorry for taking a while to get to this.

As a bonus, this seems to speed up the stemmer (I measured ~5% for C on the test data).

PR #51, opened 2017-02-13
Rust Backend

JDemler:
Hey,

for my own needs I hacked together a rust backend for the snowball compiler.
It's mostly a literal translation of the python and java backends.

The current state of it is sufficient for my needs but not enough to merge it upstream.
As far as I see it, following things would need to be done:

- [x]  Add a library similar to the ones residing in java/ and python/
- [x]  Test the implementation against all stemmers (currently only tested on eight)
- [x]  Integrate these tests into travis
- [x]  Yield warning-free rust code
- [x]  Implement missing features 

Before tackeling these tasks I just wanted to know if pull request to that repository are still accepted/merged.


compiler/generator_rust.c
```diff
@@ -0,0 +1,1355 @@
...
+    const char * s;
+    switch (p->type) {
+        case c_name:
+            write_varref(g, p->name); break;
+        case c_number:
+            write_int(g, p->number); break;
+        case c_maxint:
+            write_string(g, "sys.maxsize"); break;
+        case c_minint:
+            write_string(g, "(~sys.maxsize)"); break;
```
dscorbett: `maxint` and `minint` should be translated to `usize::MAX` and `usize::MIN`.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1355 @@
...
+
+extern void generate_program_rust(struct generator * g) {
+
+    g->outbuf = str_new();
+    g->failure_str = str_new();
+
+    generate_start_comment(g);
+    if (g->analyser->int_limits_used) {
+        /* sys.maxsize is used in the code generated for maxint and minint */
+        w(g, "import sys~N~N");
```
dscorbett: This is Python syntax. It should be `use std::usize;` in Rust.

ojwb:
PRs are accepted, though there's currently a bit of a backlog in dealing with them.

There aren't currently any formally documented requirements for a new backend, but I think the minimum requirements are approximately that it works with all the existing stemmers, and that you can run the testsuite for it (hooked up to CI if at all possible).

The current stemmers don't actually exercise all of the language features between them (e.g. `size` and `sizeof` are currently unused, and were actually reporting one too high for a long time), so it's not entirely trivial to ensure a new backend supports everything correctly.  Ideally we'd have a language feature test suite to explicitly exercise everything and ensure it works as expected.

I wouldn't worry overly about generating idiomatic code - it's not meant for human consumption, unless you're debugging the code generator, which most people won't be.  Warning-free generated code is certainly convenient, but also not a blocker for merging I feel.

compiler/driver.c
```diff
@@ -22,10 +22,13 @@ static void print_arglist(void) {
                     "             [-c++]\n"
 #ifndef DISABLE_PYTHON
                     "             [-py[thon]]\n"
-#endif
+#endif
```
ojwb: Please avoid unrelated whitespace changes.

compiler/header.h
```diff
@@ -309,7 +309,7 @@ struct options {
     FILE * output_h;
     byte syntax_tree;
     byte widechars;
-    enum { LANG_JAVA, LANG_C, LANG_CPLUSPLUS, LANG_PYTHON, LANG_JSX } make_lang;
+  enum { LANG_JAVA, LANG_C, LANG_CPLUSPLUS, LANG_PYTHON, LANG_JSX, LANG_RUST } make_lang;
```
ojwb: Again, whitespace.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1355 @@
...
+        switch (q->type) {
+            case t_string:
+            case t_integer:
+            case t_boolean:
+                w(g, "~M~V0 = other.~W0~N");
+                break;
+        }
+    }
+    /* For Python 3, this can just be super(). */
+    w(g, "~Msuper(~n, self).copy_from(other)~N~-");
```
ojwb: Python remnant.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1355 @@
...
+
+
+enum special_labels {
+    x_return = -1
+};
+
+static int new_label(struct generator * g) {
+
+    int next_label = g->next_label++;
+    g->max_label = (next_label > g->max_label) ? next_label : g->max_label;
```
ojwb: You don't seem to ever use `max_label` (it's needed for Python because there `goto` is emulated by throwing exceptions of different classes and we need to know how many different classes are used so we can declare them).  So you can probably just `return g->next_label++;` here.

JDemler:
Thanks for the reply and the code comments.

> The current stemmers don't actually exercise all of the language features between them (e.g. size and sizeof are currently unused, and were actually reporting one too high for a long time), so it's not entirely trivial to ensure a new backend supports everything correctly. Ideally we'd have a language feature test suite to explicitly exercise everything and ensure it works as expected.

This makes me really worried. 
Is there a list of features not covered by the standard tests?
Is there any known snowball-code using these features (preferably with tests)?

ojwb:
There isn't such a list.  I'm really only aware there such features because the arabic stemmer originally used `size` and `sizeof` (which is OK for wide characters, but goes wrong for UTF-8) and Assem encountered this bug.  Then I added `len` and `lenof` (which are the same as `size` and `sizeof` except in the UTF-8 case) so now `size` and `sizeof` aren't actually used.

If I knew of any existing code which exercised them, I'd have put it in the testsuite already.  I hope at some point to find time to go through the manual and write code to test everything that's documented there, but that's likely to be a significant effort.  Perhaps going through and checking what features are entirely unused by current stemmers and adding tests for those would be more tractable (that doesn't test that the used features work in ways which the current stemmers don't exercise, but would be a step forward).

This situation certainly is not ideal, though the worst issue I can see with it is that it's a latent trap waiting to catch the first person to use such a feature.

I feel it's not really your problem as a new backend author, though if you wanted to work out which features are unused and write a dummy stemmer which used them all I'd certainly be grateful.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1355 @@
...
+        case c_sizeof:
+            g->V[0] = p->name;
+            w(g, "~V0.len()");
+            break;
+        case c_len: 
+            w(g, "env.current.chars().count()");
+            break;
+        case c_size:
+            w(g, "env.current.len()");
+            break;
```
ojwb: You set `o->widechars = true;` so `lenof` should be same as `sizeof` and `len` the same as `size` (and return the number of wide characters) - see "12 Character representation" in http://snowballstem.org/compiler/snowman.html for details.

dscorbett: What if `-utf8` is specified?

JDemler: Actually I'm a bit confused about widechars and utf-8.
Rust strings are always utf-8 encoded. So string.len() is its number of bytes, string.chars().count() the number of characters.
I set widechars to true, because char literals are declared as widechars. 
For example "\u{00DF}".

JDemler:
Would you say it is better to fail hard on untested features, maybe even at compile time, or just try to implement them with the best of my knowledge?

compiler/generator_rust.c
ojwb: Sounds like you probably want to force `o->widechars = false; o->utf8 = true;` and handle converting UTF-8 sequences for non-ASCII codepoints into the `\u{xxxx}` format in `write_literal_string()`.

Or maybe the compiler should offer both wide and UTF-8 versions of the literal and the language backend can make use of whichever is more convenient.

> What if `-utf8` is specified?

Currently there's nothing to stop the user specifying `-utf8` for languages which are inherently wide character, but the resultant code is unlikely to be useful.

ojwb:
(The CI failures are timeouts on the Python tests, so presumably not your fault.)

Failing hard at runtime seems unhelpful, as that could happen in production if the feature is only used in rare cases (and words triggering those cases aren't in the test data) - if you're going to fail hard, I think it has to be at compile time (either snowball compile time, or generate rust code which won't compile, assuming rust is a compiled language).

Perhaps it's more helpful to try to implement them though - at least then when/if we have a compliance test suite, we won't have to find someone to implement them from scratch.

FYI, another language feature not used by any current stemmer is `$` on a string variable (which is what the `copyfrom()` mechanism is for in most of the backends - the C backend handles this differently though).

compiler/generator_rust.c
JDemler: > Or maybe the compiler should offer both wide and UTF-8 versions of the literal and the language backend can make use of whichever is more convenient.

Seems to be the best solution IMO but also a lot of work. 
As far as I can grasp it, that would need deep changes in structures, tokenizer and analyzer?
Like, storing all literal strings twice (wide and utf8)? Or is there a much simpler solution I am missing?
For example the amongvec struct would have b and b_wc.

The utf8-byte to codepoint conversion is probably a lot simpler and will suffice.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1348 @@
...
+        g->I[0] = replab;
+        w(g, "~Mcontinue 'replab~I0;~N");
+    }
+    w(g, "~-~M}~N");
+    g->unreachable = false;
+
+    if (keep_c) write_restorecursor(g, p, savevar);
+
+    g->I[0] = replab;
+    w(g, "~Mbreak 'replab~I0;~N~M}");
```
dscorbett: The generated source would look better with `~N` after the `}`.

JDemler: True. Thank you!

compiler/generator_rust.c
```diff
@@ -0,0 +1,1348 @@
...
+}
+
+static void write_restorecursor(struct generator * g, struct node * p,
+                                struct str * savevar) {
+
+    struct str * temp = str_new();
+    write_margin(g);
+    restore_string(p, temp, savevar);
+    write_str(g, temp);
+    w(g, ";");
```
dscorbett: `restore_string` already appends a semicolon, so this line is unnecessary.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1348 @@
...
+            s = " - "; goto label0;
+        case c_divide:
+            s = " / ";
+        label0:
+            write_char(g, '('); generate_AE(g, p->left);
+            write_string(g, s); generate_AE(g, p->right); write_char(g, ')'); break;
+        case c_cursor:
+            w(g, "env.cursor"); break;
+        case c_limit:
+            w(g, p->mode == m_forward ? "env.limit;" : "env.limit_backward;"); break;
```
dscorbett: `generate_AE` is supposed to generate an expression, so semicolons should not be generated here.

JDemler: Thanks!

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+static void generate_dollar(struct generator * g, struct node * p) {
+
+    struct str * savevar = vars_newname(g);
+    write_comment(g, p);
+    g->V[0] = p->name;
+
+    ++g->copy_from_count;
+    str_assign(g->failure_str, "env.copy_from(");
+    str_append(g->failure_str, savevar);
+    str_append_string(g->failure_str, ")");
```
dscorbett: This statement needs a semicolon.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+            str_append_string(g->failure_str, ";");
+        }
+        
+        
+        generate(g, p->aux);
+
+        if (!g->unreachable) {
+            write_margin(g);
+            write_str(g, g->failure_str);
+            w(g, ";");
```
dscorbett: This semicolon is redundant because both failure strings a few lines above already end with semicolons.

compiler/generator_rust.c
dscorbett: This statement needs a semicolon.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+    struct str * savevar = vars_newname(g);
+    write_comment(g, p);
+    g->V[0] = p->name;
+
+    ++g->copy_from_count;
+    str_assign(g->failure_str, "env.copy_from(");
+    str_append(g->failure_str, savevar);
+    str_append_string(g->failure_str, ")");
+    g->B[0] = str_data(savevar);
+    writef(g, "~M~B0 = ~n();~N"
```
dscorbett: `~B0` is a new variable so it needs `let` before it.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+            w(g, "~M break 'lab~I0;~N");
+            g->unreachable = true;
+    }
+}
+
+static void write_failure_if(struct generator * g, char * s, struct node * p) {
+
+    writef(g, "~Mif ", p);
+    writef(g, s, p);
+    write_block_start(g);
```
dscorbett: It would be better style to `writef(g, " ", p);` before the start of the block.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+    g->S[0] = p->mode == m_forward ? "0" : "env.limit_backward";
+
+    write_failure_if(g, "~S0 as i32 > c || c > env.limit as i32", p);
+    writef(g, "~Menv.cursor = c as usize;~N", p);
+}
+
+static void generate_delete(struct generator * g, struct node * p) {
+
+    write_comment(g, p);
+    writef(g, "~Mif !env.slice_del(){~N"
```
dscorbett: It would be better style to change `){` to `) {`. The same is true in `generate_slicefrom`.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+    }
+    switch (g->failure_label)
+    {
+        case x_return:
+            w(g, "~Mreturn false;~N");
+            g->unreachable = true;
+            break;
+        default:
+            g->I[0] = g->failure_label;
+            w(g, "~M break 'lab~I0;~N");
```
dscorbett: There is an extra space before the `break`.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+        int i;
+        for (i = 0; i < x->literalstring_count; i++)
+        {
+            g->I[0] = i;
+            g->I[1] = v->i;
+            g->I[2] = v->result;
+            g->L[0] = v->b;
+            g->S[0] = i < x->literalstring_count - 1 ? "," : "";
+
+            w(g, "~MAmong(~L0, ~I1, ~I2, None");
```
dscorbett: If the fourth field is always `None`, it is unnecessary and should be removed.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+static void generate_grouping_table(struct generator * g, struct grouping * q) {
+
+    int range = q->largest_ch - q->smallest_ch + 1;
+    int size = (range + 7)/ 8;  /* assume 8 bits per symbol */
+    symbol * b = q->b;
+    symbol * map = create_b(size);
+    int i;
+    for (i = 0; i < size; i++) map[i] = 0;
+
+    /* Using unicode would require revision here */
```
dscorbett: Is this copied-pasted comment true for the Rust generator?

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+    g->V[0] = p->name;
+    writef(g, "~M~V0 = env.assign_to(~V0)~N", p);
+}
+
+static void generate_sliceto(struct generator * g, struct node * p) {
+
+    write_comment(g, p);
+    g->V[0] = p->name;
+    writef(g, "~M~V0 = env.slice_to(~V0);~N"
+              "~Mif ~V0 == ''{~N"
```
dscorbett: This is invalid Rust syntax.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+
+    write_comment(g, p);
+    if (keep_c) writef(g, "~Mc = env.cursor;~N", p);
+    if (p->mode == m_forward) {
+        writef(g, "~Menv.insert(env.cursor, env.limit, ", p);
+    } else {
+        writef(g, "~Menv.insert(env.limit_backward, env.cursor, ", p);
+    }
+    generate_address(g, p);
+    writef(g, ")~N", p);
```
dscorbett: This is a statement and needs a trailing semicolon.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+    write_comment(g, p);
+    write_failure(g);
+}
+
+static void generate_debug(struct generator * g, struct node * p) {
+
+    write_comment(g, p);
+    g->I[0] = g->debug_count++;
+    g->I[1] = p->line_number;
+    writef(g, "~Menv.debug(~I0, ~I1)~N", p);
```
dscorbett: This needs a semicolon.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+            if (v->function != 0)
+            {
+                w(g, ", \"");
+                if (v->function->type == t_routine) {
+                    /* Need to use mangled version of private name here. */
+                    w(g, "_~n");
+                }
+                write_varname(g, v->function);
+                w(g, "\"");
+            }
```
dscorbett: Representing functions with mangled strings is unnecessary in Rust. It would be more idiomatic to replace the contents of this block with:
```c
                w(g, ", ");
                write_varname(g, v->function);
```

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+}
+
+extern void generate_program_rust(struct generator * g) {
+
+    g->outbuf = str_new();
+    g->failure_str = str_new();
+
+    generate_start_comment(g);
+    if (g->analyser->int_limits_used) {
+        /* sys.maxsize is used in the code generated for maxint and minint */
```
dscorbett: This comment needs to be updated for Rust. It is still referring to Python.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+
+static void generate_start_comment(struct generator * g) {
+
+    w(g, "//! This file was generated automatically by the Snowball to Rust compiler~N");
+    w(g, "//! http://snowballstem.org/~N~N");
+}
+
+static void generate_class_begin(struct generator * g) {
+  w(g, "use snowball::SnowballEnv;~N");
+  w(g, "use snowball::Among;~N~N");
```
dscorbett: Where are these defined? Did you forget to commit something?

JDemler: Hey, thank you for all your hard work on this. I'm failing to keep up :)

These structs are not yet part of the backend. 
I used them for my purposes but I will use them in the test application of the backend.

Until then you can find them in this crate: https://github.com/JDemler/rust-stemmers

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+    w(g, "~Mstatic A_~I0: &'static [Among; ~I1] = &[~N~+");
+    {
+        int i;
+        for (i = 0; i < x->literalstring_count; i++)
+        {
+            g->I[0] = i;
+            g->I[1] = v->i;
+            g->I[2] = v->result;
+            g->L[0] = v->b;
+            g->S[0] = i < x->literalstring_count - 1 ? "," : "";
```
dscorbett: It is better style to include the comma even after the final item in a multiline list.

compiler/generator_rust.c
dscorbett: Now that I have read your crate, I see that this field should not be removed. Instead, it should be either `None` or <code>Some(Box::new(<var>v->function</var>))</code> depending on `v->function`.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+}
+
+static void generate_loop(struct generator * g, struct node * p) {
+
+    struct str * loopvar = vars_newname(g);
+    write_comment(g, p);
+    w(g, "~Mfor _ in 0..");
+    generate_AE(g, p->AE);
+    writef(g, "~N", p);
+    writef(g, "~M{", p);
```
dscorbett: It is more consistent with the rest of the generated code for the `{` to be on the same line as the `for`.

Also, the `{` should have a new line after it.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+    if (x->command_count == 0 && x->starter == 0) return;
+
+    if (x->starter != 0) generate(g, x->starter);
+
+    p = p->left;
+    if (p != 0 && p->type != c_literalstring) p = p->right;
+    w(g, "~Mif among_var == 0{~N~+");
+    write_failure(g);
+    g->unreachable = false;
+    w(g, "~-~M}~N~N");
```
dscorbett: There is an extra new line here.

JDemler:
I have a question about the `copy_from` methods.

In the java and python backends, they should be created if `g->copy_from_count != 0` [here](https://github.com/snowballstem/snowball/blob/master/compiler/generator_java.c#L1346).

`g->copy_from_count` is incremented in the `generate dollar` method. But even when running through `generate_dollar`, the `copy_from`-method is not generated. 
Is this a bug or am I doing something wrong? 

As far as I understand, the `copy_from` method in the base classes only copies the fields of the stemmers (limit, cursor, current etc.) but the generated method also copies the local variables.
Is this necessary?

compiler/generator_rust.c
```diff
@@ -0,0 +1,1348 @@
...
+    if (p != 0 && p->type != c_literalstring) p = p->right;
+    w(g, "~Mif among_var == 0 {~N~+");
+    write_failure(g);
+    g->unreachable = false;
+    w(g, "~-~M}~N~N");
+
+    while (p != 0) {
+        if (p->type == c_bra && p->left != 0) {
+            g->I[0] = case_number++;
+            w(g, "~Melse if among_var == ~I0{~N~+");
```
dscorbett: It would be better style to put a space before the `{`.

dscorbett:
Try moving the call to `generate_copyfrom` after the call to `generate_methods`.

`copy_from` makes a copy of a stemmer, i.e. it copies all a stemmer’s fields. Therefore, the generated function must copy all the generated stemmer’s fields as well as the base fields. A more Rusty equivalent would be for the stemmer to `impl Clone`, which should be automatically derivable.

ojwb:
The copyfrom stuff is needed when `$` is applied to a string variable.  I think it needs to copy all the variables in the program over, so that the temporary state has them.

Currently the machinery for this is duplicated in each backend which needs it, despite it being pretty much the same for everything except C (where we just do a `struct` copy).  The base class copies the members it holds, but can't copy members which are in the subclass.  The only slight trick is that is `copyfrom()` isn't needed for a particular snowball program, we avoid generating `copyfrom()` for the corresponding subclass.

I am attempting to factor out common code like this, as repeating so much code is dumb for hopefully obvious reasons.  But that means if it's not working, it's presumably down to the code in your backend!

compiler/generator_rust.c
ojwb: FYI, this function feature isn't used by most of the stemmers - if you want an example to prod, the finnish stemmer uses it.

compiler/generator_rust.c
ojwb: I think it would indeed requiring storing two versions of every literal string.

Looks like you can just call `get_utf8()` (in `space.c`) in a loop to get the Unicode codepoints from a UTF-8 string.

JDemler: Wow... now everything makes sense. Thanks. `get_utf8()` works like a charm!
Before I really had troubles implementing `find_among_b`. 
Because it operated on characerts not bytes. Now everything works fine!

compiler/generator_rust.c
JDemler: Thanks. Amongs with routines in them are working now.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+    write_failure(g);
+    g->unreachable = false;
+    w(g, "~-~M} ");
+
+    while (p != 0) {
+        if (p->type == c_bra && p->left != 0) {
+            g->I[0] = case_number++;
+            w(g, "else if among_var == ~I0 {~N~+");
+            generate(g, p);
+            w(g, "~-~M} ");
```
dscorbett: This leaves a line-final space after the last among clause. Instead of adding spaces after the braces here and on line 1074, add a space before the `else` on line 1079.

JDemler:
Sorry. Can't get travis to work at the moment. 
Will be offline for two days but I will fix it on sunday!

rust/src/snowball/snowball_env.rs
```diff
@@ -0,0 +1,404 @@
...
+    pub fn get_next_char_boundry(data: &str, mut index: usize) -> usize {
+        loop {
+            if index > data.len() || data.is_char_boundary(index) {
+                return index;
+            }
+            index += 1;
+        }
+    }
+
+    pub fn get_next_char_boundry_b(data: &str, mut index: usize) -> usize {
```
dscorbett: In these two functions’ names, `boundry` should be `boundary`.

rust/src/snowball/snowball_env.rs
```diff
@@ -0,0 +1,404 @@
...
+                }
+            }
+            return res as i32;
+        } else {
+            return self.cursor as i32;
+        }
+    }
+
+    /// Check if the char the cursor points to is in the grouping
+    /// This is determined by weird magic stuff. I have no idea how it works
```
dscorbett: Here’s how the weird magic works. A grouping is represented by a minimum code point, a maximum code point, and a bitfield of which code points in that range are in the grouping. For example, in english.sbl, `valid_LI` is `'cdeghkmnrt'`. The minimum and maximum code points are 99 and 116, so every time one of these grouping functions is called for `g_valid_LI`, `min` must be 99 and `max` must be 116. There are 18 code points within that range (inclusive) so the grouping is represented with 18 bits, plus 6 bits of padding:
```
cdefghij klmnopqr st
11101100 10110001 01000000
```
The first bit is the least significant. Those three bytes become `&[0b00110111, 0b10001101, 0b00000010]`, which is `&[55, 141, 2]`, which is how `g_valid_LI` is defined in english.rs.

rust/src/main.rs
```diff
@@ -0,0 +1,117 @@
...
+            "norwegian" => Stemmer { stemmer: Box::new(algorithms::norwegian_stemmer::stem) },
+            "porter" => Stemmer { stemmer: Box::new(algorithms::porter_stemmer::stem) },
+            "portuguese" => Stemmer { stemmer: Box::new(algorithms::portuguese_stemmer::stem) },
+            "romanian" => Stemmer { stemmer: Box::new(algorithms::romanian_stemmer::stem) },
+            "russian" => Stemmer { stemmer: Box::new(algorithms::russian_stemmer::stem) },
+            "spanish" => Stemmer { stemmer: Box::new(algorithms::spanish_stemmer::stem) },
+            "swedish" => Stemmer { stemmer: Box::new(algorithms::swedish_stemmer::stem) },
+            "tamil" => Stemmer { stemmer: Box::new(algorithms::tamil_stemmer::stem) },
+            "turkish" => Stemmer { stemmer: Box::new(algorithms::turkish_stemmer::stem) },
+            x => panic!("Unkown algorithm '{}'", x)
```
dscorbett: Instead of hard-coding this mapping, generate it automatically. See for example how python/create_init.py creates the equivalent mapping for Python.

Also, “Unkown” should be “Unknown”.

rust/src/snowball/snowball_env.rs
dscorbett: What are these functions for? Can they be removed?

compiler/generator_rust.c
```diff
@@ -0,0 +1,1347 @@
...
+
+    write_comment(g, p);
+
+    g->S[0] = p->mode == m_forward ? "" : "_b";
+    g->I[0] = x->number;
+
+    if (x->command_count == 0 && x->starter == 0) {
+        write_failure_if(g, "env.find_among~S0(~A_~I0, context) == 0", p);
+    } else {
+        writef(g, "~Mlet among_var = env.find_among~S0(~A_~I0, context);~N", p);
```
dscorbett: `among_var` does not live long enough in the Rust source generated from
```snowball
externals (stem)
define stem as (
    [gopast substring]
    among('x' (<- 'X'))
)
```
See `generate_define` in generator.c for how to use `amongvar_needed` to fix this.

rust/src/snowball/snowball_env.rs
JDemler: Thank you for explaining. Is it ok if I copy this explanation into the code comments?

dscorbett: Yes, it is okay.

ojwb:
I found a couple of small test files lying around, which you may find useful for testing:

`size.sbl` which should leave everything alone, except for the empty string which should be changed to (literally) `<empty>`.

```
externals ( stem )

integers ( zero )

define stem as (
    $zero = 0
    $zero == size
    <- '<empty>'
)
```

`limits.sbl` which checks that valid code is generated for `minint` and `maxint`:

```
externals (foo)
integers (x)
define foo as ( $x = maxint $x = minint )
```

compiler/generator_rust.c
```diff
@@ -1188,14 +1188,23 @@ static void generate(struct generator * g, struct node * p) {
 }
 
 static void generate_start_comment(struct generator * g) {
-
```
ojwb: Please try to avoid reformatting the code copied from other generator files like this.  Martin's code formatting is quite unusual in places, but there's a lot of code shared between the files which would be much better factored out, and keeping the source code matching where it's really the same will make it significantly easier to do that refactoring in the future.

JDemler: Sorry, I didn't realize these newlines were intentional. Fixed it by matching `generator_python.c`s newline pattern.

JDemler:
Thanks for the test cases. Both succeed. 

Automatic binding to the algorithms is still missing as [David suggested](https://github.com/snowballstem/snowball/pull/51#pullrequestreview-22550715),  but other to that, I feel pretty confident about the current state of the backend.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1376 @@
...
+    write_comment(g, p);
+
+    g->S[0] = p->mode == m_forward ? "" : "_b";
+    g->I[0] = x->number;
+
+    if (x->command_count == 0 && x->starter == 0) {
+        write_failure_if(g, "env.find_among~S0(~A_~I0, context) == 0", p);
+    } else {
+        writef(g, "~Mamong_var = env.find_among~S0(~A_~I0, context);~N", p);
+        //  writef(g, "~Mlet among_var = env.find_among~S0(~A_~I0, context);~N", p);
```
ojwb: This commented out line should probably be removed now.

rust/src/main.rs
```diff
@@ -0,0 +1,117 @@
...
+            let stdin = std::io::stdin();
+            for line in stdin.lock().lines() {
+                writeln!(&mut output, "{}", stemmer.stem(&line.unwrap())).unwrap();
+            }
+        }        
+    }
+}
+
+
+/// Wrapps a usable interface around the actual stemmer implementation
```
ojwb: "Wraps"

compiler/generator_rust.c
ojwb: Still present!

compiler/generator_rust.c
```diff
@@ -0,0 +1,1376 @@
...
+static void generate_start_comment(struct generator * g) {
+
+    w(g, "//! This file was generated automatically by the Snowball to Rust compiler~N");
+    w(g, "//! http://snowballstem.org/~N~N");
+}
+
+// rustc emmits warnings if variables don't match the style guide
+// (i.e. upper-case for globals, snake case for fields etc.)
+// To allow warning free compilation of generated code and
+// consistency with snowball variable namings we allow some kind of warnings here
```
ojwb: Currently the intent is that the compiler code is compatible with ISO C90, though I've been wondering if we should give up on that and say a C99 compiler is needed - the main benefit I see is being able to declare variables at the point of initialisation.  But we should probably explicitly make that decision, rather than just slide into it by using `//` comments instead of `/*` ... `*/`.  There are a few other `//` comments in this file.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1376 @@
...
+
+    g->V[0] = q;
+
+    if (q->type == t_routine) {
+        w(g, "~N~Mfn ~W0(env: &mut SnowballEnv, context: &mut Context) -> bool {~+~N");
+    } else {
+        w(g, "~N~Mpub fn ~W0(env: &mut SnowballEnv) -> bool {~+~N");
+        generate_setup_context(g);
+    }
+    // w(g, "~Mprintln!(\"~W0: \\t\\t\\t{:?}\", env);~N");
```
ojwb: Looks like a remnant from debugging which should be removed.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1376 @@
...
+    w(g, "~M~V0 ~S0 "); generate_AE(g, p->AE); w(g, ";~N");
+}
+
+static void generate_integer_test(struct generator * g, struct node * p, char * s) {
+
+    g->V[0] = p->name;
+    g->S[0] = s;
+    w(g, "~Mif !(~V0 ~S0 ");
+    generate_AE(g, p->AE);
+        w(g, ")");
```
ojwb: Misindent.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1376 @@
...
+}
+
+static void write_literal_string(struct generator * g, symbol * p) {
+
+    int i = 0;
+    write_string(g, "\"");
+    while (i < SIZE(p)) {
+        int ch;
+        i += get_utf8(p + i, &ch);
+        if (32 <= ch && ch <= 127) {
```
ojwb: Should be `ch < 127` (this flaw is in all the other backends except C, but I'm about to commit a fix for those).

compiler/generator_rust.c
```diff
@@ -0,0 +1,1376 @@
...
+            w(g, "env.current.chars().count()");
+            break;
+        case c_size:
+            w(g, "env.current.len()");
+            break;
+    }
+}
+
+/* K_needed() tests to see if we really need to keep c. Not true when the
+   the command does not touch the cursor. This and repeat_score() could be
```
ojwb: Extra `the` here (already fixed in other backends, but after you took a copy).

compiler/generator_rust.c
```diff
@@ -0,0 +1,1376 @@
...
+    g->unreachable = false;
+    w(g, "~-~M}~N");
+    if (keep_c) write_restorecursor(g, p, savevar);
+
+    g->failure_label = a0;
+    str_delete(g->failure_str);
+    g->failure_str = a1;
+
+    write_check_limit(g, p);
+    write_inc_cursor(g, p);
```
ojwb: Spaces at end of line.

compiler/generator_rust.c
```diff
@@ -0,0 +1,1376 @@
...
+    generate(g, p->left);
+
+    w(g, "~-~M}~N");
+    str_delete(loopvar);
+    g->unreachable = false;
+}
+
+static void generate_repeat(struct generator * g, struct node * p, struct str * loopvar) {
+
+     struct str * savevar = vars_newname(g);
```
ojwb: Extra space in indent.

rust/src/main.rs
```diff
@@ -83,28 +83,13 @@ pub struct Stemmer {
...
-            "spanish" => Stemmer { stemmer: Box::new(algorithms::spanish_stemmer::stem) },
-            "swedish" => Stemmer { stemmer: Box::new(algorithms::swedish_stemmer::stem) },
-            "tamil" => Stemmer { stemmer: Box::new(algorithms::tamil_stemmer::stem) },
-            "turkish" => Stemmer { stemmer: Box::new(algorithms::turkish_stemmer::stem) },
-            x => panic!("Unknown algorithm '{}'", x)
-        }
+        // Have a look at ../build.rs
+        // There we generate a file that is rust code for a closure that returns a stemmer.
+        // We match against all the algorithms in src/snowball/algoritms/ folder.
+        // Alas, this canp not be included as a match statment or function because of rusts
```
ojwb: "can", "rust's"

rust/build.rs
```diff
@@ -0,0 +1,55 @@
+use std::env;
+use std::fs;
+use std::fs::{OpenOptions};
+use std::io::Write;
+use std::path::Path;
+
+
+// This build script makes the code independend from the algorithms declared
```
ojwb: "independent"

compiler/generator_rust.c
```diff
@@ -0,0 +1,1376 @@
...
+    g->failure_str = a1;
+}
+
+static void generate_start_comment(struct generator * g) {
+
+    w(g, "//! This file was generated automatically by the Snowball to Rust compiler~N");
+    w(g, "//! http://snowballstem.org/~N~N");
+}
+
+// rustc emmits warnings if variables don't match the style guide
```
ojwb: "emits"

rust/src/main.rs
dscorbett: Actually, “canp not” should be “cannot”, “rusts” should be “Rust's”, and “statment” should be “statement”.

compiler/generator_rust.c
JDemler: I solved this now by both cloning `env` and `context` in `generate_dollar`. 
My testcase compiled. Don't know if it behaves as intended, though.
This is now similar to the c solution AFAIK.

.travis.yml
```diff
@@ -24,11 +24,11 @@ matrix:
...
-  - [ -z "$RUST" ] ||  curl https://sh.rustup.rs -sSf -o rustup.sh
-  - [ -z "$RUST" ] || chmod ugo+x rustup.sh
-  - [ -z "$RUST" ] || ./rustup.sh -y
-  - [ -z "$RUST" ] || source "$HOME/.cargo/env"
-  - [ -z "$RUST" ] || rm rustup.sh
+  - test -z "$RUST" || curl https://sh.rustup.rs -sSf -o rustup.sh
+  - test -z "$RUST" || chmod ugo+x rustup.sh
+  - test -z "$RUST" || ./rustup.sh -y
+  - test -z "$RUST" || source "$HOME/.cargo/env"
+  - test -z "$RUST" || rm rustup.sh
```
ojwb: Oh sorry, I should have remembered as I've hit essentially the same issue before elsewhere.

compiler/generator_rust.c
```diff
@@ -922,15 +922,21 @@ static void generate_setlimit(struct generator * g, struct node * p) {
  * current string */
 static void generate_dollar(struct generator * g, struct node * p) {
 
-    struct str * savevar = vars_newname(g);
+    struct str * savevar_env = vars_newname(g);
+    struct str * savevar_context = vars_newname(g);
     write_comment(g, p);
     g->V[0] = p->name;
     ++g->copy_from_count;
```
ojwb: If `copy_from` is gone, it's pointless to increment the counter for it...

JDemler:
This should be it.

ojwb:
Will check things over when I get a chance, but might be a few days.

Meanwhile, here's a testcase for `$` on strings which requires `copy_from` in Java:

```
strings (x y)
routines (setvars doit)
externals (stem)
define setvars as ( $x = 'ingly' $y = 'ing' )
backwardmode (
define doit as ( [x] <- y )
)
define stem as (
    setvars
    backwards doit
)
```

This should change "-ingly" endings to "-ing" (e.g. "knowingly" -> "knowing").  It works for the C backend, but Java doesn't work correctly currently (because when it "copies" the object it just creates a reference to the same object).

(It's dumb that string assignment invokes this mechanism, but this is a simple test of it as things currently stand).

JDemler:
I adjusted the behaviour of the backend to meet `c`'s. 
One thing I am not sure about though:

This is the code for one assignment. `$x = 'ingly'`
```c
/* z is copied. Is this a deep copy or a shallow copy? */
/* In other words: is z->S[0] copied?*/
struct SN_env env1 = * z; /* $ x, line 4 */
       ...
       /*Set z->S[0]*/
        z->S[0] = z->p;
       /* Reset z->S[0] to original value?*/
        * z = env1;
```
So, apparently, string variables are not copied but only its pointers are.
What about integer and boolean variables? 

For example:
```
strings (x y z)
integers ( p1 )
routines ( setvars checkit doit )
externals (stem)

define setvars as ( $z = 'ing' $x = 'ingly' $y = z $z checkit)

define checkit as ($p1 = limit)

backwardmode (define doit as ( [x] <- y ))

define stem as (
    setvars
    backwards doit
)
```
Is `p1` meant to be set or not?
The current behaviour of the rust backend is, that variables set in `generate_dollar` are not copied and then restored. 
So `p1` is set to the limit of `z`.

JDemler:
To make the testing easier I tried to build something to run and change tests:

https://github.com/JDemler/snowball/tree/fb-test_cases

This could probably be a lot simpler. But it should ease your testing.
Just use `make check_rust` and all the tests in `tests/` will run.

ojwb:
Sadly the manual doesn't really talk about the nitty-gritty of this in much detail.  Conceptually `$x` where `x` is a string seems to be rather like `with` in Pascal - it gives an implied context to a group of operations.  But it's unclear from the manual if it's a lexical group or if this implied context needs to be carried into subroutines (currently it is for C, and that's probably more sensible, though perhaps harder to implement for some backends).

> What about integer and boolean variables?

In C those aren't restored - here's `struct SN_env`'s definition:

```
struct SN_env {
    symbol * p;
    int c; int l; int lb; int bra; int ket;
    symbol * * S;
    int * I;
    unsigned char * B;
};
```

The shallow copies which happen before and after the group of operations will (needlessly) copy the pointers `S`, `I`, and `B`, but those pointers don't ever change during execution of the snowball program so that won't change any of the snowball variables.  The part which is important is that struct members `p` to `ket` are copied and restored.  We could optimise this slightly to just copy the start of the struct (on x86-84 we'd copy 28 instead of 56 bytes; on x86, 24 instead of 36 bytes).

So for Java a correct implementation would be approximately:

```
SnowballProgram v_1 = new SnowballProgram();
v_1.copy_from((SnowballProgram) this);
current = S_x;

// ...

((SnowballProgram) this).copy_from(v_1);
```

This relies on the fact that Java object variables are really references, so after `current = S_x;` modifying `current` also modifies `S_x`.

Except that in the Java backend `current` is a `StringBuffer` and `S_x` a `StringBuilder` (or compatible class the user specifies) so you can't just assign them like this, and also I'm not sure quite how this is meant to interact with the failure mechanism.

JDemler:
Thanks for the clarification.
The rust backend is compatible to c. It restores `p` to `ket` but doesn't restore the local variables.

compiler/generator_rust.c
```diff
@@ -939,6 +945,7 @@ static void generate_dollar(struct generator * g, struct node * p) {
               "~Menv.limit = env.current.len();~N", p);
     generate(g, p->left);
     if (!g->unreachable) {
+        g->V[0] = p->name;
         /* Update string variable. */
         w(g, "~M~V0 = env.current.clone().into_owned();~N");
```
ojwb: Hmm, so does this make a copy of the contents of the variable and then copy it back afterwards?

If so, that's different to the C implementation because during `$s` `s` will maintain its old value (and only be updated at the end) - for example, consider:

```
externals ( stem )
strings ( s t )
define stem as (
  $s = 'foo'
  $s ( = 'bar' $t = s )
)
```

With the C backend, both `s` and `t` will be set to `bar`; if the value of `s` is only updated at the end of the `$` section then `t` would instead be set to `foo`.

JDemler: Then I don't understand the generated c code correctly.
Tried to annotate to my understanding:
```c
        //Copy env
        struct SN_env env2 = * z; /* $ s, line 5 */
        // Set current value to S[0] = 'foo'
        z->p = z->S[0];
        z->lb = z->c = 0;
        z->l = SIZE(z->p);
        {   int ret;
            {   int saved_c = z->c;
                // Set current value to 'bar'
                ret =insert_s(z, z->c, z->l, 3, s_1); /* =, line 5 */
                z->c = saved_c;
            }
            if (ret < 0) return ret;
        }
        {
            //Copy env
            struct SN_env env3 = * z; /* $ t, line 5 */
            //Set current value to S[1] = '1'
            z->p = z->S[1];
            z->lb = z->c = 0;
            z->l = SIZE(z->p);
            {   int ret;
                {   int saved_c = z->c;
                    // Set current value to S[0] (currently still 'foo')
                    ret =insert_v(z, z->c, z->l, z->S[0]); /* = s, line 5 */
                    z->c = saved_c;
                }
                if (ret < 0) return ret;
            }
            failure = 0; /* mark success */
            // Set S[1] = current value = 'foo'
            z->S[1] = z->p;
            // Reset env and thus current value to 'bar'
            * z = env3;
            if (failure) goto lab1;
        }
        failure = 0; /* mark success */
    lab1:
        //Set S[0] to current value 'bar'
        z->S[0] = z->p;
        //Reset env
        * z = env2;
```

What I probably get wrong is how pointers are handled.
```c
 z->p = z->S[0];
```
Makes `z->p` point to `z->S[0]`. So if `z->p` is changed `z->S[0]` is changed too, correct?
But then, the reset would be useless:
```c
z->S[0] = z->p;
```

Possible solutions:

1. One way to get the intended behaviour is to special case string variable assignments (this would involve work in the tokenizer and analyzer and probably the other backend which is probably not feasable).

2. Another way would be to hack up the rust implementation to use mutable pointers like the C implemenation (IMO we should avoid that).

3. Or we keep a field in the generator to what string variable the env is currently set.  
Then in `generate_assignfrom` not only the `env` could be assigned to but also the string variable.
(This would involve adding a new field to the generator struct)

I personally favor option 3 but am not sure if I understand the consequences of introducing a new field to the generator struct.

ojwb: Oh, it's a bit more complex than I realised.

So in C, the `z->p` is set to `z->S[0]` so in Snowball the current string now aliases `s` (as I said), **but** if the current string is modified such that it needs to be reallocated (i.e. it gets longer than the length of the space allocated) then `z->p` will get updated to point to the reallocated block, but `z->S[0]` will not be changed, and so the aliasing is broken.  And in fact you can't safely access the string variable `s` then (until after the `$` has ended), as `z->S[0]` will point to a block that's been released.

So actually the generated C code doesn't reliably have the semantics I suggested, so the Rust backend's approach seems OK.  This restriction on not being allowed to look at the currently "active" string ought to at least be documented, and perhaps the C backend should `NULL` out the variable to help avoid accidental uses.  We could also warn about situations where this could happen at compile time fairly easily I think (basically "does any code reachable from inside this `$` touch the variable that the `$` is over.  There are already some similar checks - for example, if any code reachable from _X_ could modify the cursor.)

We could fix things so the variable either always aliases (e.g. via a double pointer indirection, or the "current active string variable" idea) or so it always keeps its original value (e.g. copy the value, or copy on write) but I'm not sure either is worth the overhead it would add to the generated code (or the effort to implement).  This is a corner case of a feature that's not used in any of the current stemmers after all.

JDemler: Ok, that makes sense.
I aggree, that the plan to add documentation and throw a compiler error in this case would be the best solution for now.
What about the Java, Python and the JSX backends? String variables with literal assignments seem to be completly broken for them. Are there still active maintainers? Should we throw a compiler error when using string variables with these backends?

compiler/generator_rust.c
```diff
@@ -0,0 +1,1369 @@
...
+    w(g, "#[derive(Clone)]~N");
+    w(g, "struct Context {~+~N");
+    for (q = g->analyser->names; q; q = q->next) {
+        g->V[0] = q;
+        switch (q->type) {
+            case t_string:
+                w(g, "~M~W0: String,~N");
+                break;
+            case t_integer:
+                w(g, "~M~W0: usize,~N");
```
ojwb: @JDemler Sorry, I failed to spot this before but the Rust backend maps Snowball integers to an unsigned type, but [the manual](http://snowballstem.org/compiler/snowman.html) clearly says that they should be "signed integers".  I think it's probably simplest to change this globally (so the cursor and limits become signed too) because of commands like `atlimit`, `tolimit`, the ability to use `cursor` in numeric expressions, etc but I'm not sure if there are consequences to making this change as I've never programmed in Rust.  Also not sure what Rust type is most appropriate - I guess `i32` or `isize` from looking at https://doc.rust-lang.org/book/second-edition/ch03-02-data-types.html (in practice Snowball strings are at most a few hundred characters so a 32-bit limit is not problematic).

JDemler: Let me see if I understand the problem correctly:

1. Using AEs one could possibly create negative integer values
2.  `minint` is defined as the minimum negative number

The rust backend is not able to handle both cases corretly.
1. first would probably result in a runtime panic.
2. `minint` returns `0` in generated code

Beyond AEs though there is no chance to ever encounter a negative number, as they are filtered:
``` rust
   let c = env.byte_index_for_hop(3);
    if 0 as i32 > c || c > env.limit as i32 {
        return false;
    }
    env.cursor = c as usize;
```

Do I understand the problem correctly?
Anyways... the simplest solution would probably be to change everything to `isize`.
Not only the local variables, but also `cursor`, `limit`, `limit_backward`, `bra` and `ket` inside `SnowballEnv`

Are there any real live examples of AEs resulting in negative values?

ojwb: I think your summary is accurate.

You can also set an integer variable without involving an AE by using `setmark x` but that should never give a negative value, so I think any negative value must come from an AE.

> Are there any real live examples of AEs resulting in negative values?

I noticed it while looking at the code in `generator_rust.c` (I'm working on a compiler optimisation to "localise" variables - i.e. map a Snowball variable to a function-local variable in the target language when appropriate).

I don't think there aren't any examples in the current stemmers we ship.  There could be cases in Snowball code that's not been submitted to us, but realistically it's probably a theoretical issue at the moment.  However, I think it is worth addressing this as negative integers are something one might reasonably want to use - we're not in the domain of weird programs written only to prove a pedantic point.

JDemler: All right. I'll try to whip up a PR next week.

PR #52, opened 2017-02-16
Avoid generating `--` from two `-` operators

dscorbett:
The C, Java, and JSX files generated from this source:
```snowball
externals (stem)
define stem as ([hop --1] delete)
```
contain the invalid syntax `--1`. This commit adds a space after the unary `-` operator in those three languages. Python and Rust don’t have the `--` operator, so their generators don’t need changing. C# does, so its generator will.

### DIFF
--- a/compiler/generator.c
+++ b/compiler/generator.c
@@ -313,7 +313,7 @@ static void generate_AE(struct generator * g, struct node * p) {
         case c_minint:
             write_string(g, "MININT"); break;
         case c_neg:
-            write_char(g, '-'); generate_AE(g, p->right); break;
+            write_string(g, "- "); generate_AE(g, p->right); break;
         case c_multiply:
             s = " * "; goto label0;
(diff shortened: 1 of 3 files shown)


ojwb:
Good spot.

While this is a very artificial case, I wonder if it's cleaner to just optimise away the double negation at the Snowball level, like so:

```
--- a/compiler/generator.c
+++ b/compiler/generator.c
@@ -313,6 +313,10 @@ static void generate_AE(struct generator * g, struct node * p) {
         case c_minint:
             write_string(g, "MININT"); break;
         case c_neg:
+            if (p->right->type == c_neg) {
+                generate_AE(g, p->right->right);
+                break;
+            }
             write_char(g, '-'); generate_AE(g, p->right); break;
         case c_multiply:
             s = " * "; goto label0;
```

Though it'd be better to do this in generic code than each backend.

PR #56, opened 2017-05-04
fix potential memory leak (found by clang static analyzer)

patperry:
Minor fix to silence Xcode warning: if `stemmer->close` is `NULL`, then `sb_stemmer_delete` leaks memory.


ojwb:
Thanks

PR #57, opened 2017-05-07
add Go generator

mschoch:
initial implementation ported from rust generator
primarily focused on getting functionality working
not on the best or most performant Go code

seems to work, based on:
make check_go
and other ad-hoc testing

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -22,6 +22,12 @@ cargoflags ?=
 rust_src_main_dir = rust/src
 rust_src_dir = $(rust_src_main_dir)/snowball/algorithms
 
+go ?= go
+goflags ?= stemwords/algorithms.go stemwords/main.go
+gofmt ?= gofmt
+go_src_main_dir = go
+go_src_dir = $(go_src_main_dir)/algorithms
+
 ICONV = iconv
 #ICONV = python ./iconv.py
 
@@ -48,7 +54,8 @@ COMPILER_SOURCES = compiler/space.c \
 		   compiler/generator_java.c \
 		   compiler/generator_jsx.c \
 		   compiler/generator_python.c \
-		   compiler/generator_rust.c
+		   compiler/generator_rust.c \
+		   compiler/generator_go.c
 
 COMPILER_HEADERS = compiler/header.h \
 		   compiler/syswords.h \
@@ -108,6 +115,8 @@ PYTHON_SOURCES = $(libstemmer_algorithms:%=$(python_output_dir)/%_stemmer.py) \
 		 $(python_output_dir)/__init__.py
 JSX_SOURCES = $(libstemmer_algorithms:%=$(jsx_output_dir)/%-stemmer.jsx)
 RUST_SOURCES = $(libstemmer_algorithms:%=$(rust_src_dir)/%_stemmer.rs)
+GO_SOURCES = $(libstemmer_algorithms:%=$(go_src_dir)/%_stemmer.go) \
+	$(go_src_main_dir)/stemwords/algorithms.go
 
 COMPILER_OBJECTS=$(COMPILER_SOURCES:.c=.o)
 RUNTIME_OBJECTS=$(RUNTIME_SOURCES:.c=.o)
@@ -230,6 +239,23 @@ $(rust_src_dir)/%_stemmer.rs: algorithms/%/stem_Unicode.sbl snowball
 	o="$(rust_src_dir)/$${l}_stemmer"; \
 	echo "./snowball $< -rust -o $${o}"; \
 	./snowball $< -rust -o $${o}
+
+$(go_src_main_dir)/stemwords/algorithms.go:
+	@echo "Generating algorithms.go"
+	@cd go/stemwords && go generate && cd ../..
+
+$(go_src_dir)/%_stemmer.go: algorithms/%/stem_Unicode.sbl snowball
+	@l=`echo "$<" | sed 's!\(.*\)/stem_Unicode.sbl$$!\1!;s!^.*/!!'`; \
+	o="$(go_src_dir)/$${l}/$${l}_stemmer"; \
+	mkdir -p $(go_src_dir)/$${l}
+	@l=`echo "$<" | sed 's!\(.*\)/stem_Unicode.sbl$$!\1!;s!^.*/!!'`; \
+	o="$(go_src_dir)/$${l}/$${l}_stemmer"; \
+	echo "./snowball $< -go -o $${o}"; \
+	./snowball $< -go -o $${o} -gop $${l}
+	@l=`echo "$<" | sed 's!\(.*\)/stem_Unicode.sbl$$!\1!;s!^.*/!!'`; \
+	o="$(go_src_dir)/$${l}/$${l}_stemmer"; \
+	echo "$(gofmt) -s -w $(go_src_dir)/$${l}/$${l}_stemmer.go"; \
+	$(gofmt) -s -w $(go_src_dir)/$${l}/$${l}_stemmer.go
 
 $(jsx_output_dir)/%-stemmer.jsx: algorithms/%/stem_Unicode.sbl snowball
 	@mkdir -p $(jsx_output_dir)
@@ -466,6 +492,24 @@ check_rust_%: $(STEMMING_DATA_ABS)/%
 	  rm tmp.in; \
 	else \
 	  $(cargo) run $(cargoflags) -- -l `echo $<|sed 's!.*/!!'` -i $</voc.txt -o $(PWD)/tmp.txt; \
+	fi
+	@if test -f '$</output.txt.gz' ; then \
+	  gzip -dc '$</output.txt.gz'|$(THIN_TEST_DATA)|diff -u - tmp.txt; \
+	else \
+	  diff -u $</output.txt tmp.txt; \
+	fi
+	@rm tmp.txt
+
+check_go: $(GO_SOURCES) $(libstemmer_algorithms:%=check_go_%)
+
+check_go_%: $(STEMMING_DATA_ABS)/%
+	@echo "Checking output of `echo $<|sed 's!.*/!!'` stemmer for Go"
+	@cd go && if test -f '$</voc.txt.gz' ; then \
+	  gzip -dc '$</voc.txt.gz'|$(THIN_TEST_DATA) > tmp.in; \
+	  $(go) run $(goflags) -l `echo $<|sed 's!.*/!!'` -i tmp.in -o $(PWD)/tmp.txt; \
+	  rm tmp.in; \
+	else \
+	  $(go) run $(goflags) -l `echo $<|sed 's!.*/!!'` -i $</voc.txt -o $(PWD)/tmp.txt; \
 	fi
 	@if test -f '$</output.txt.gz' ; then \
 	  gzip -dc '$</output.txt.gz'|$(THIN_TEST_DATA)|diff -u - tmp.txt; \
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -242,7 +242,7 @@ $(rust_src_dir)/%_stemmer.rs: algorithms/%/stem_Unicode.sbl snowball
 
 $(go_src_main_dir)/stemwords/algorithms.go:
 	@echo "Generating algorithms.go"
-	@cd go/stemwords && go generate && cd ../..
+	@cd go/stemwords && go generate
 
 $(go_src_dir)/%_stemmer.go: algorithms/%/stem_Unicode.sbl snowball
 	@l=`echo "$<" | sed 's!\(.*\)/stem_Unicode.sbl$$!\1!;s!^.*/!!'`; \
--- a/compiler/driver.c
+++ b/compiler/driver.c
@@ -7,6 +7,8 @@
 #define DEFAULT_BASE_CLASS "org.tartarus.snowball.SnowballProgram"
 #define DEFAULT_AMONG_CLASS "org.tartarus.snowball.Among"
 #define DEFAULT_STRING_CLASS "java.lang.StringBuilder"
+#define DEFAULT_GO_PACKAGE "snowball"
+#define DEFAULT_GO_SNOWBALL_RUNTIME "github.com/snowballstem/snowball/go"
 
 static int eq(const char * s1, const char * s2) {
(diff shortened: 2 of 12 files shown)


compiler/generator_go.c
```diff
@@ -0,0 +1,1253 @@
...
+
+static void generate_AE(struct generator * g, struct node * p) {
+    const char * s;
+    switch (p->type) {
+        case c_name:
+            write_varref(g, p->name); break;
+        case c_number:
+            write_int(g, p->number); break;
+        case c_maxint:
+            write_string(g, "1<<32 - 1"); break;
```
dscorbett: Shouldn’t this expression be parenthesized?

compiler/generator_go.c
```diff
@@ -0,0 +1,1253 @@
...
+
+static void generate_assignfrom(struct generator * g, struct node * p) {
+
+    int keep_c = p->mode == m_forward; /* like 'attach' */
+
+    write_comment(g, p);
+    if (keep_c) writef(g, "~Mvar c = env.Cursor~N", p);
+    /* Copying limits and cursors is necessary here because the rust
+     * borrowchecker does not like taking something from someone you are about
+     * to mutate... */
```
dscorbett: This comment is about Rust.

compiler/generator_go.c
mschoch: The << has higher precedence than -, so in isolation it's fine.  But, I guess I don't fully know what other code might be generated in a surrounding expression.  So it would be safer to have the whole thing inside parens.

compiler/generator_go.c
mschoch: Yep, that is copy/paste leftover from the rust version.  I'll remove this and any similar I find.

mschoch:
What is the preferred style of addressing review comments for this project?  Would you like separate commits that can be squashed later?  Or force push into single commit now?

compiler/generator_go.c
ojwb: The operator precedence in Go appears to be different to other languages, so explicit parentheses seem wise if only to avoid confusing people who aren't familiar with Go.

But the expression seems wrong anyway - snowball integers should map to a signed integer type, so this should be the largest value of that signed integer type, but the current expression would be the largest value of an unsigned integer type.

ojwb:
fixup commits is probably most helpful.

compiler/generator_go.c
mschoch: Thanks for pointing that out.  It looks to be a problem in the Rust version as well, they use usize for integer variables and operations.  Making the switch isn't hard, but I need to review how that affects other code, as I suspect these values can then later show up as slice offsets later.

mschoch: OK, so I'm trying to work through the correct way to fix this.  In this implementation, like the Rust version, cursor and limit are unsigned, which seems to be supported by the line, "The cursor, c, (and the limit, l) can be thought of as having a numeric value, from zero upwards".

However, integer values are signed integers according to the manual as well.

These can interact.  One example I've come up with is:

```
integers ( p1 )

externals ( stem )

define stem as (
backwards(
  $p1 = -3
  tomark p1
  )
)
```

In forwards mode, it seems clear that tomark on a negative value would always fail, and the cursor (unsigned) would always be ahead of that position.  But, in backwards mode, what is the expected behavior?  Should it fail because the argument is negative?

Are there other operations that intermingle program integer variables with the cursor/limit?

compiler/generator_go.c
```diff
@@ -0,0 +1,1253 @@
...
+    write_block_end(g);
+    g->unreachable = false;
+}
+
+static void generate_hop(struct generator * g, struct node * p) {
+
+    write_comment(g, p);
+    g->S[0] = p->mode == m_forward ? "" : "-";
+
+    w(g, "~Mvar c = env.ByteIndexForHop(~S0");
```
dscorbett: This Snowball code:
```snowball
externals (stem)
define stem as backwards (hop -1)
```
generates this line of Go:
```go
    var c = env.ByteIndexForHop(--1)
```
which is a syntax error.

mschoch: Any plans to turn these into a test suite?

compiler/generator_go.c
mschoch: Looking at the c code generated, I see that there cursor/limit/etc are all signed as well.  Does the runtime guard against them becoming negative at any point?  Or is that just considered an error in the program?

mschoch:
Addressed some of the initial problems.  Next I'll be making the larger change to support signed integers throughout.

mschoch:
I've now updated the implementation to support signed integers (both as a datatype, and internally for tracking cursor/limit/etc as is done in the C impl)

I identified one example program where the cursor can be set negative, and I updated the runtime to behave like the C version (it's not clear if this is well specified or not).

The example I considered is this:
```
integers ( p1 )

externals ( stem )

define stem as (
backwards(
  tomark -2
  = 'bobo'
  )
)
```

There may be other operations that need to be updated to behave correctly if the cursor becomes negative (or perhaps this is just considered undefined or poor snowball programming).

compiler/generator_go.c
```diff
@@ -0,0 +1,1241 @@
...
+
+static void generate_AE(struct generator * g, struct node * p) {
+    const char * s;
+    switch (p->type) {
+        case c_name:
+            write_varref(g, p->name); break;
+        case c_number:
+            write_int(g, p->number); break;
+        case c_maxint:
+            write_string(g, "(1<<31) - 1"); break;
```
dscorbett: The parentheses need to go around the _whole_ expression. Currently, `-maxint` becomes `-(1<<31) - 1` but it should become `-(1<<31 - 1)`. The same goes for `minint`.

ojwb:
Or perhaps better, use constants `MaxInt32` and `MinInt32` from the `math` package instead: https://golang.org/pkg/math/#pkg-constants

ojwb:
Snowball has the concept of undefined operations (e.g. _The effect of obeying `among` when the preceding `substring` is not obeyed is undefined_), but the manual defines `tomark` as _tomark `AE` moves `c` forward to the position given by `AE`_, and when `AE` is negative we can't move `c` **forward** to it so that should probably be a no-op.

That's also how the C backend (which is essentially the reference implementation as it was the original backend) seems to handle it.

ojwb:
Oh, I see you mean in `backwards` mode when trying to position `c` before the start of the string.  I think that should signal `f` - _In the case of `tomark AE` , a similar fail condition occurs as with `hop AE`. If `c` is already beyond `AE`, or if position `l` is before position `AE`, the signal is `f`._ and _String commands have been described with `c` to the left of `l` and moving right. But the process can be reversed. `backwards C` `c` and `l` are swapped over, and `c` moves left towards `l`_

compiler/generator_go.c
ojwb: I suspect it's simpler to make the cursor and limit signed, because there are operations that compare them with and copy their values to/from integer variables, which are definitely meant to be signed.

I'm not aware of a valid way to make the cursor or limit negative (nor to set the cursor past the limit).

ojwb: > Are there other operations that intermingle program integer variables with the cursor/limit?

This can happen in at least:

* integer commands such as `$p1 <= cursor` and `$p1 = limit`
* integer expressions such as `$p1 + cursor`
* `hop`, `setmark`, `tomark`, `atmark`

compiler/generator_go.c
ojwb: It would be great to have one, but currently I'm lacking the time to sort it out myself.

`hop -1` is equivalent to `false` so rather an artificial example, but `hop -$x` is something you might more realistically encounter.

mschoch:
Sure, I can use the constants in the math package, but then it complicates the generation because they won't always be imported, and imported-but-not-used packages are an error in Go.

I already avoid conditionally needing the utf8 package by wrapping the method I need in the runtime package.  I could do the same for the math symbols.  But you seem have something really specific you want.

mschoch:
Can you explain when you mean when you say, "I'm not aware of a valid way to make the cursor or limit negative"?

In this program:

```
integers ( p1 )

externals ( stem )

define stem as (
backwards(
  tomark -2
  = 'bobo'
  )
)
```

My undrstanding is that `tomark` sets the cursor to -2.

When I look at the C code I see the line:

```
z->c = -2; /* tomark, line 7 */
```

ojwb:
```
z->c = -2; /* tomark, line 7 */
```

That looks like a bug in the C backend to me - thanks for pointing it out.  There's a gating check generated, but what it checks doesn't match what the docs say:

```
    if (z->c < -2) return 0;
    z->c = -2; /* tomark, line 7 */
```

It's more clearly wrong when not in backwards mode:

```
    if (z->c > 1000000) return 0;
    z->c = 1000000; /* tomark, line 10 */
```

That checks _if `c` is already beyond `AE`_, but there's no check for _if position `l` is before position `AE`_.

mschoch:
@ojwb thanks for the info on tomark and hop.  So, after re-reading the spec, at a high level, both hop and tomark should have checks ensuring that the new cursor position is valid.  And I think if we get this part right, my concerns about signed integers affecting cursor/limit locations should go away.

mschoch:
I spent some more time looking at hop, the specification and the generated C code.

First, the spec says "moves c AE character positions towards l, but if AE is negative, or if there are less than AE characters between c and l the signal is f"

Sample program:
```
externals ( stem )

define stem as (
  hop 2
  hop -1
  = 'b'
)
```

On the input string 'beautiful', this produces 'bb'.

So, first it seems the negative AE is allowed, despite what the spec says.  

It does move c "closer" to l by one character, when you realize that closer means farther in the negative direction.

Looking at the generated C code, it appears that if AE is positive, you can't exceed l.  And if AE is negative, you can't go less than 0.

Do you agree with the C implementation here?  It seems to be safe and generally useful, and the spec just needs slightly different wording.

mschoch:
I believe I've addressed the issues you've raised so far, and I'm committed to continue working on the `tomark` and `hop` issues we've identified, once we agree on what the C version should do.

I've also added a travis CI job as well.

Are there any remaining barriers to getting this merged?

ojwb:
Thanks for hooking up the CI.

I don't think there are any blockers - I just need to actually sit down and give it one last check over, then merge assuming it looks OK.

Extending the spec to allow `hop` with a negative argument indeed seems reasonable.

go/README.md
```diff
@@ -0,0 +1,55 @@
+# Go Target for Snowball
+
+The initial implementation was built as a port of the Rust target.  The initial focus has been on getting it to function, and making it work correctly.  No attempt has been made to beautify the implementation, generate code, or address performance issues.
+
+## Usage
+
+Until this PR is accepted, don't use `go get` to get the runtime library in place.  Instead, clone the mschoch repo into `$GOPATH/src/github.com/snowballestem/snowball`.
```
sokolovstas: this is typo 
> snowballestem 
must be
snowballstem

GNUmakefile
```diff
@@ -231,6 +240,23 @@ $(rust_src_dir)/%_stemmer.rs: algorithms/%/stem_Unicode.sbl snowball
 	echo "./snowball $< -rust -o $${o}"; \
 	./snowball $< -rust -o $${o}
 
+$(go_src_main_dir)/stemwords/algorithms.go:
+	@echo "Generating algorithms.go"
+	@cd go/stemwords && go generate && cd ../..
```
ojwb: `cd ../..` isn't useful here - each command in a make rule is run in its own shell (or effectively is).

go/README.md
```diff
@@ -0,0 +1,55 @@
+# Go Target for Snowball
+
+The initial implementation was built as a port of the Rust target.  The initial focus has been on getting it to function, and making it work correctly.  No attempt has been made to beautify the implementation, generate code, or address performance issues.
```
ojwb: ITYM "generated code" there.

go/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Usage
+
+Until this PR is accepted, don't use `go get` to get the runtime library in place.  Instead, clone the mschoch repo into `$GOPATH/src/github.com/snowballstem/snowball`.
```
ojwb: This note doesn't make sense to merge into the main repo - it would be wrong the moment it was merged.

go/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Using the Generated Stemmers
+
+Assuming you generated a stemmer, put that code in a package which is imported by this code as `english`.
+
+```
+env := snowball.NewEnv("beautiful")
+english.Stem(env)
+fmt.Printf("stemmed word is: %s", env.Current())
+```
```
ojwb: When using the C stemmer at least, you'd reuse `env` when stemming a lot of words (to avoid the overhead of allocating a new `env` for each word).  I'd guess that's the case for Go too, so would be useful to steer users towards here.

GNUmakefile
mschoch: Fixed

go/README.md
mschoch: Fixed

go/README.md
mschoch: Removed

go/README.md
mschoch: We didn't have a method to cleanly reuse, so I added a Reset() method to accomplish this.

ojwb: That's what `SetCurrent()` ought to do - just changing the string but leaving the cursor pointing potentially past the end of it isn't a useful operation.

Looks like the Rust code you based this on doesn't handle this properly...

mschoch: Yeah that makes sense.  Pushed another commit to switch/fix this.

compiler/driver.c
```diff
@@ -115,6 +126,14 @@ static void read_options(struct options * o, int argc, char * argv[]) {
                 continue;
             }
 #endif
+#ifndef DISABLE_Go
```
avsej: @mschoch will it work with lowercase `o`?

ojwb: It's not consistent with `DISABLE_GO` elsewhere so you'd have to define both, but I don't think anything actually makes use of all the ability to disable backends from the build.  It doesn't seem all that useful as none of the backends drag in extra dependencies.  Certainly nothing in the repo defines these; I've been meaning to check on the mailing list if anyone is making use of them in a wider context and if not to drop them.

PR #60, opened 2017-06-10
Use ~M for indent consistently

ojwb:
Fixes occasional lines with 2 space indent instead of 4.


ojwb:
@mschoch Does this look sensible?

PR #61, opened 2017-06-10
fix typo DISABLE_Go to match DISABLE_GO elsewhere

mschoch:


mschoch:
Thanks for finding this @avsej

PR #84, opened 2018-06-28
Add missing CFLAGS to allow for example address sanitizer to compile

pointhi:
Currently, it is not possible to compile snowball with address sanitizer, because ```CFLAGS``` and ```CPPFLAGS``` are not handed over to all compilation steps.

For example, the following command does not finish correctly:

```
make CFLAGS="-fsanitize=address"
```

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -171,7 +171,7 @@ clean:
 	rmdir $(js_output_dir) || true
 
 snowball: $(COMPILER_OBJECTS)
-	$(CC) -o $@ $^
+	$(CC) $(CFLAGS) $(CPPFLAGS) -o $@ $^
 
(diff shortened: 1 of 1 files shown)


GNUmakefile
```diff
@@ -171,7 +171,7 @@ clean:
 	rmdir $(js_output_dir) || true
 
 snowball: $(COMPILER_OBJECTS)
-	$(CC) -o $@ $^
+	$(CC) $(CFLAGS) $(CPPFLAGS) -o $@ $^
```
ojwb: Thanks for the patch.

I agree `$(CFLAGS)` should be passed here, but `$(CPPFLAGS)` shouldn't AFAICS.  The C preprocessor isn't being invoked here.

I checked automake and that passes `$(CFLAGS)` but not `$(CPPFLAGS)` when linking.

PR #90, opened 2018-11-13
c++ compatibility for SN_set_current using extern

vvarma:


ojwb:
This change doesn't make sense to me - it's not useful to call `SN_set_current()` without having called `SN_create_env()` first, so this change doesn't actually allow compiling snowball with a C compiler and using it from C++ code.

If we're going to add `extern "C"` it should be for all functions in the API.

vvarma:
Please correct me if I have misunderstood the issue.
`SN_create_env` is not invoked directly, the generated code invokes this method. And the generated methods needed to create the env have `extern "C"` . https://github.com/snowballstem/snowball/blob/master/compiler/generator.c#L1542

The only api which is used directly from `api.h` is SN_set_current, all other methods used for invoking the stemmer are from the generated files which are compatible with c++.

ojwb:
Ah yes, I see.

I think it's cleaner to make all these functions `extern "C"` so I'll push a commit to do that instead of your proposed patch.

Thanks for reporting this, and for the clarification.

vvarma:
Sure that works. Though I don't see the point of marking the other APIs as extern since they can't be used independent of the language specific implementation unlike the set current method.
Having said that, the change you mentioned will ensure compatibility with c++ as I faced.

ojwb:
They were already `extern` - a C function prototype is `extern` by default unless it's marked `static`.

What `extern "C"` does is to specify that a function should use C calling conventions and name mangling.

If you compile the stemmer as C code and the calling code as C++ (which is probably the most natural way to build) then where the other two functions are actually declared and used they will have C linkage (because they are compiled with a C compiler which always uses C linkage) so it seems a bad idea to declare them as having C++ linkage here even if they aren't actually currently used from C++ code.

PR #94, opened 2019-06-24
Update Python packaging

mitya57:
- Fixed `NameError` in demo script when input contains uppercase words.
- Updated PyPI links to point to the new domain.
- Updated metadata (email and classifiers) to satisfy PyPI checks.
- Bumped version to 1.9.0.

I have tested this on [test.pypi.org](https://test.pypi.org/project/snowballstemmer/), when you merge this I will upload it to the production server.


python/stemwords.py
```diff
@@ -81,7 +81,7 @@ def stemming(lang, input, output, encoding, pretty):
     for original in codecs.open(input, "r", encoding).readlines():
         original = original.strip()
         # Convert only ASCII-letters to lowercase, to match C behavior
-        original = ''.join((lower_(c) if 'A' <= c <= 'Z' else c for c in original))
+        original = ''.join((c.lower() if 'A' <= c <= 'Z' else c for c in original))
```
ojwb: I was curious where this came from and if it ever worked so looked back to where it was introduced.  It was added in 2015 (e7e180d8561af193aeb777115ecea6a539080d9f) by you it seems!  I didn't find an answer as to whether it worked back then - the testsuite doesn't pass upper-case ASCII so quite possibly not.

mitya57: Indeed it was my commit, but I just imported @shibukawa's code so this originated there. Maybe it was copied from some other language than Python.

PR #95, opened 2019-07-07
Document supported Python versions through trove classifiers

jdufresne:
Helps potential users know, at a quick glance, if the library is
compatible with a given project.


mitya57:
Can you also add Python 3.7 to the Travis config, to make sure we really support it?

Here, before 3.6:
https://github.com/snowballstem/snowball/blob/533602101f963eeb0c38343d94c428ceef740c0c/.travis.yml#L18-L24

jdufresne:
Done :slightly_smiling_face:

PR #101, opened 2019-07-10
Clarify that we are using the 3-Clause BSD license

### LINKED ISSUE #100: PyPI licence tags
@mitya57 This was held for moderation by the mailing list, but it didn't seem relevant for most subscribers (and I'm not sure if you're even subscribed to the list) so putting it in an issue here instead:

>Hello!
>
>You have recently released a new version of snowballstemmer. Thanks! It will
>probably make a lot of people happy. There is a tiny issue with your
>package on PyPI though: it is a bit difficult to guess the exact
>license you use.
>
>
>It is very clear that you use one of the BSD licenses. PyPI allows you
>to use the following classifier:
>
>    License :: OSI Approved :: BSD License
>
>But it is not very precise. You can also use the "license" field in
>setuptools, but there is no "fixed" format, so you could write "BSD2",
>"BSD-2", "BSD-2-Clause", etc. This makes it hard for automated tools
>to properly guess which BSD license you are using.
>
>May I suggest using the aforementioned classifier *and* an SPDX
>identifier (https://spdx.org/licenses/) in the "license" field?
>
>
>Thanks,
>Cyril Roelandt
>
>(This email was automatically generated, but you can answer it! A real
>human being will read your answer.)

(If you think it makes more sense to list your address as maintainer for PyPI rather than the mailing list, please do).

mitya57:
Fixes #100, reported by Cyril Roelandt.


ojwb:
Thanks.

PR #104, opened 2019-08-08
adding catalan and basque modules

iolalla:
Hi Team,

Today I got people asking about the catalan stemmer, so this is the reason I forked your git and added the files, catalan and basque in the algorithm folders and added those to libstemmer/modules.txt 

I've tested it with go and works smoothly.

I hope this helps.

Cheers.

### DIFF
--- a/libstemmer/modules.txt
+++ b/libstemmer/modules.txt
@@ -28,6 +28,8 @@ portuguese      UTF_8,ISO_8859_1        portuguese,pt,por
 romanian        UTF_8,ISO_8859_2        romanian,ro,rum,ron
 russian         UTF_8,KOI8_R            russian,ru,rus
 spanish         UTF_8,ISO_8859_1        spanish,es,esl,spa
+catalan         UTF_8,ISO_8859_1        catalan,ca,cat
+basque          UTF_8,ISO_8859_1        basque,ba,bas,eus
 swedish         UTF_8,ISO_8859_1        swedish,sv,swe
 tamil           UTF_8                   tamil,ta,tam
 turkish         UTF_8                   turkish,tr,tur
--- a/libstemmer/modules.txt
+++ b/libstemmer/modules.txt
@@ -29,7 +29,7 @@ romanian        UTF_8,ISO_8859_2        romanian,ro,rum,ron
 russian         UTF_8,KOI8_R            russian,ru,rus
 spanish         UTF_8,ISO_8859_1        spanish,es,esl,spa
 catalan         UTF_8,ISO_8859_1        catalan,ca,cat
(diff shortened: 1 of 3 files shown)


ojwb:
Thanks.

We also really need to add a vocabulary list adding to the snowball-data repo for each new stemmer (that's why the tests are failing under CI), and the website also needs updating - I've documented the steps here:

https://github.com/snowballstem/snowball/blob/master/CONTRIBUTING.rst#adding-a-new-stemming-algorithm

There are already pages for both Basque and Catalan on the website, but it would be good to add `algorithm_vocab` to those pages so we show a sample of the stemmer at work like we do for most of the other languages (e.g. see `algorithms/french/stemmer.tt`).

iolalla:
Hey @ojwb,

 I've added the data pull request ( [#10](https://github.com/snowballstem/snowball-data/pull/10)) , if you agree and  everything looks good, the next step will be to improve the  web, as you proposed.

Cheers.

ojwb:
I prodded the CI to rerun using the snowball-data PR, but all the languages are failing except Go and Rust due to a difference in the stemming for Basque.

Looking at the code, I think the problem is that it the source code in `basque.sbl` uses literal UTF-8 for ñ - that will work in UTF-8, but not in other encodings.

I've pushed a commit to this PR to replace the two `.sbl` files with the versions from the snowball-website repo which have already been updated to use the newer `{U+xxxx}` notation for representing Unicode characters.  That should solve this (I've tested locally for C and Java, which covers ISO-8859-1 and wide character Unicode).

libstemmer/modules.txt
```diff
@@ -28,6 +28,8 @@ portuguese      UTF_8,ISO_8859_1        portuguese,pt,por
 romanian        UTF_8,ISO_8859_2        romanian,ro,rum,ron
 russian         UTF_8,KOI8_R            russian,ru,rus
 spanish         UTF_8,ISO_8859_1        spanish,es,esl,spa
+catalan         UTF_8,ISO_8859_1        catalan,ca,cat
+basque          UTF_8,ISO_8859_1        basque,eu,eus,baq
```
mitya57: Did you intentionally list these languages after Spanish?

This makes some sense from geographical point of view, but in my opinion it’s better to have this list sorted alphabetically.

ojwb: Indeed, thanks for spotting - I've pushed 2aa2809f67332b7f5c3494a741cd0390a6b40cff to fix that and 0e0d307b082c0c5758ea7f89bb9b3d3dd9b215e1 to note this in the instructions.

iolalla: Thanks for the correction!

PR #108, opened 2019-09-04
Add estonian.sbl and changes to modules.txt

Lindafr:


libstemmer/modules.txt
```diff
@@ -34,6 +34,7 @@ basque          UTF_8,ISO_8859_1        basque,eu,eus,baq
 swedish         UTF_8,ISO_8859_1        swedish,sv,swe
 tamil           UTF_8                   tamil,ta,tam
 turkish         UTF_8                   turkish,tr,tur
+estonian		UTF_8,ISO_8859_1		estonian,et,est
```
mitya57: Please keep the sorting order, i.e. insert this between English and Finnish. See 0e0d307b082c0c5758ea7f89bb9b3d3dd9b215e1.

ojwb:
The output from the stemmer doesn't seem to match what's expected, e.g.:

https://travis-ci.org/snowballstem/snowball/jobs/580651042#L614

algorithms/estonian.sbl
```diff
@@ -0,0 +1,256 @@
...
+	substantive
+	verb_exceptions
+	verb
+	nu
+)
+
+stringdef a" hex 'E4' //a-umlaut ä
+stringdef o" hex 'F6' //o-umlaut ö
+stringdef o' hex 'F5' //o with tilde õ
+stringdef u" hex 'FC' //u-umlaut ü
```
ojwb: You should use the newer notation here - that may fix at least some of the test failures.  E.g.:

```
stringdef a"   '{U+00E4}'
```

algorithms/estonian.sbl
```diff
@@ -0,0 +1,256 @@
...
+		'sse' (RV or LONGV) //illative: saapa-sse, tegemisse -> tegemisse
+		'st' (RV or LONGV) //elative: saapa-st, rohelist -> rohelist
+		'le' (RV or LONGV) //allative: raamatu-le
+		'lt' (RV or LONGV) //ablative: raamatu-lt
+		'ga' (RV or LONGV) //komitatiive: õpetaja-ga
+		'ks' (RV or LONGV) //translative: õpetaja-ks
+		'ta' (RV or LONGV) //abessive and da-infinitive: õpetaja-ta and hüpa-ta
+		't' //partitiiv, raamatu-t and kapsas-t
+		's'  (RV or LONGV)//inessive and sg3pst: raamatu-s and sõiti-s
+		'l'  (RV or LONGV) //adessive: raamatu-l and kapsa-l.
```
ojwb: Are you aware the action applied here in **every** matching case will be the same?  I could see the logic in putting the action just once or putting it explicitly for every case, but putting it against all but one case makes me wonder if you've misunderstood this feature of `among` and are expecting the above to mean no action for `t` or something.

algorithms/estonian.sbl
```diff
@@ -0,0 +1,256 @@
...
+
+externals ( stem )
+booleans ( is_verb )
+integers ( p1 )
+groupings ( V1 RV KI GI)
+stringescapes {}
+
+define V1 'aeiou{o'}{a"}{o"}{u"}'
+define RV 'aeiuo'
+define KI 'kptgbdshfšzž'
```
ojwb: This uses literal UTF-8 accented characters, which will only work when the generated code is working in UTF-8.

You should use `stringdef` for these - add something like these above:

```
stringdef s* '{U+0161}' //s-caron š
stringdef z* '{U+017E}' //z-caron ž
```

And then `'kptgbdshf{s*}z{z*}'` here.

algorithms/estonian.sbl
```diff
@@ -0,0 +1,256 @@
...
+
+
+define stem as (
+	do mark_regions
+	not verb_exceptions
+	unset is_verb
+	backwards (
+	    do emphasis
+	    do verb
+	    try (not is_verb do substantive)
```
ojwb: Can't you use the signal from `verb` instead of communicating the same information via this `is_verb` boolean?

I'm suggesting replacing the two lines above with something like:

```
test verb or do substantive
```

and stripping out the rest of the `is_verb` code.

algorithms/estonian.sbl
```diff
@@ -0,0 +1,256 @@
...
+		'n{a"}en' 'n{a"}eb' 'n{a"}ed' 'n{a"}eme' 'n{a"}ete' 'n{a"}evad' (<-'n{a"}gi')
+		'n{a"}eksin' 'n{a"}eks' 'n{a"}eksid' 'n{a"}eksime' 'n{a"}eksite' (<-'n{a"}gi')
+		'n{a"}gemata' 'n{a"}hakse' 'n{a"}hti' 'n{a"}ha' 'n{a"}gema' (<-'n{a"}gi')
+	)
+)
+
+
+define stem as (
+	do mark_regions
+	not verb_exceptions
```
ojwb: It looks to me like currently the work `mark_regions` does is unused when `verb_exceptions` succeeds - if so we can swap these two lines and speed up the stemmer a little.

algorithms/estonian.sbl
ojwb: Reading the description of the algorithm in the PR for the website, it seems that `t` is meant to be different, and the example of `kapsas-t` seems to confirm that.  However that example will actually be handled as `kapsa-st` (since the longest suffix which matches is chosen).

Assuming you want `t` not to use the code following `s` then you can add `()` after `t` so it uses that empty code instead.  If it should have the `(RV or LONGV)` check then the description for the website needs adjusting.

Also, is `sg3pst` in the comment after `s` a typo?

ojwb: > Also, is `sg3pst` in the comment after `s` a typo?

Ah no, probably shorthand for "singular third-person past tense".

ojwb:
@Lindafr Did you see the comments here and on the website PR?  I'm wondering if the github notification emails got spam filtered - I've seen that happen before.

Some of the points raised really need input from you to determine what the intention is, so this PR is currently blocked waiting for you to respond.

ojwb:
@Lindafr Ping?

This is stalled on a response for you (or potentially someone else with familiarity with Estonian, though you have the benefit of having come up with the algorithm so probably have clearer insights on what you intended).

Please see the comments above, and also those on https://github.com/snowballstem/snowball-website/pull/15

ojwb:
In an attempt to un-stall this PR, I've created an `estonian-wip` branch (for both snowball and snowball-data) and have applied changes that seem clearly right (I'll mark those as resolved above).

libstemmer/modules.txt
ojwb: Fixed by eb1d16a2acc7b7e6e9c6770a53ec2811da65f127.

algorithms/estonian.sbl
ojwb: Fixed by 2ff0d924943f484995c74157672ab413a1aa5774.

algorithms/estonian.sbl
ojwb: Fixed by 8a0ec827a6f7017d4c6dbde0aa82d5e719d09155.

ojwb:
> The output from the stemmer doesn't seem to match what's expected, e.g.: [...]

I realised that the proposed `modules.txt` entry was wrong:

> estonian        UTF_8,ISO_8859_1        estonian,et,est

The s-caron and z-caron characters aren't present in ISO-8859-1 so that's clearly not suitable and I've removed `,ISO_8859_1` from here in ca4faabf1c1d6fa0a98ea6832259c8ec27a4c8fe.

Wikipedia says ISO-8859-4 and ISO-8895-10 were designed to cover Estonian so we could support one or both of those if they're useful.  It needs `charsets/ISO-8859-4.sbl` and/or `charsets/ISO-8859-10.sbl` creating (which can be automated) and them hooking up in the build system.

ojwb:
We're still not getting output matching the snowball-data PR's `output.txt` but we're at least down to a small number of differences (though there are less than thousand words in the test vocabulary, which seems much too small to really provide good coverage) - for C in UTF-8:

```
$ make check_utf8_estonian|grep '^[-+]'
--- ../snowball-data/estonian/output.txt	2023-09-17 12:27:49.424593568 +1200
+++ tmp.txt	2023-09-17 13:50:59.013673661 +1200
-puusärk
+puusärki
-masinavärk
+masinavärki
-Luba
+luba
make: *** [GNUmakefile:528: check_utf8_estonian] Error 1
```

For Python (wide character Unicode):

```
$ make check_python_estonian|grep '^[-+]'
--- /home/olly/git/snowball-data/estonian/output.txt	2023-09-17 12:27:49.424593568 +1200
+++ tmp.txt	2023-09-17 13:51:33.729694512 +1200
-puusärk
+puusärki
-masinavärk
+masinavärki
-nädala 
+nädala
-Luba
+luba
-laulma 
+laul
make: *** [GNUmakefile:691: check_python_estonian] Error 1
```

The `Luba` case looks easy - I think `output.txt` was prepared using the stemmer without lower-casing the input (which `stemwords` does for ASCII A-Z), and changing the `L` to `l` eliminates that so I'll just fix that in the test data.

The stemmer should behave the same whether working in UTF-8 or wide characters.  Failing to is typically caused by using `size` or `sizeof`, or treating cursor or limit values as numbers of characters but `estonian.sbl` doesn't do appear to do any of those, and it's currently unclear to me why this is happening in this case.

dscorbett:
The “nädala” and “laulma” lines end with spaces. Is the C version stripping spaces before passing the string to the stemmer?

ojwb:
> The “nädala” and “laulma” lines end with spaces.

Oh, good spot.  I'd checked for invisible Unicode even, but not invisible ASCII.  Thanks.

> Is the C version stripping spaces before passing the string to the stemmer?

Actually `stemwords.c` passes the spaces intact to the algorithm - it's `stemwords.py` which strips them by doing `original = original.strip()` (which is how it removes the newline character).

Removing those trailing spaces (which I've pushed as snowballstem/snowball-data@d9b64f74997ea1134181cdef6ae694a36f7c41e2) changes the mismatching results to be the same for C and Python (and ada, csharp, go, java, js, rust):

```
-puusärk
+puusärki
-masinavärk
+masinavärki
-laulma
+laul
```

ojwb:
Looking at the `laulama` case I tried all the `laul` prefixed words in `voc.txt`:

```
$ grep ^laul estonian/voc.txt | ../snowball/stemwords -l estonian -p2
laulmaks                      laul
laulma                        laul
laulda                        laulda
laulan                        laula
laulad                        laula
laulab                        laula
laulame                       laula
laulate                       laula
laulavad                      laula
laulsin                       laul
laulis                        lauli
laulgu                        laulgu
laulev                        laulev
laulnud                       laul
lauldakse                     laul
lauldud                       laul
laulvale                      laul
```

These seem to all be forms of the verb "to sing" (https://en.wiktionary.org/wiki/laulma for everything except for `laulvale`; `laulvale` seems to be related to singing too as best I can make out).  Ideally they'd all give the same stem but it seems better that `laulma` gets conflated with at least some other forms by stemming to `laul`, so I propose we adjust `output.txt` for this case.

ojwb:
I found the repo where this was developed and there was a rule present in an earlier version of the algorithm with would have given `puusärk` and `masinavärk` but it was removed before submission to us:

https://github.com/texta-tk/est-snowball/commit/a1cf0654ded4beead22499bc2a7649a77b83c038#diff-2b5cd3254584cdfe27ac5ebcbdbf1881eeb43a10612fe91d9aeeb19c4e4501f0L54-R55

So it seems that `output.txt` simply wasn't updated to match this change in the algorithm and so this should be fixed by updating it to match the output actually given.

ojwb:
I've updated `output.txt` and CI now passes (python's still running but it's good locally).

Re ISO-8859-10 and/or ISO-8859-4 versions I think we skip these for now as they can cleanly be added later if there's demand.

Remaining things to resolve:

* The `-t` suffix removal is documented as unconditional but actually ends up with a `(RV or LONGV)` condition due to Snowball "falling through" to the next action in an `among` (https://github.com/snowballstem/snowball/pull/108#pullrequestreview-283982649).  I need to investigate this more but a simple test of removing this condition didn't look like an improvement.

* Create a larger test vocabulary - I've downloaded an Estonian wikipedia dump and it's currently crunching so I should have one later today.

* Eliminate `is_verb` flag (https://github.com/snowballstem/snowball/pull/108#pullrequestreview-284015848) which I'll look at once we have better test coverage.

* `mark_regions` minor optimisation (https://github.com/snowballstem/snowball/pull/108#pullrequestreview-284016944) which I'll look at once we have better test coverage.

* There are some typos and other issues with the submitted website documentation PR which I'll try to resolve.

ojwb:
>     * The `-t` suffix removal is documented as unconditional but actually ends up with a `(RV or LONGV)` condition due to Snowball "falling through" to the next action in an `among` ([Add estonian.sbl and changes to modules.txt #108 (review)](https://github.com/snowballstem/snowball/pull/108#pullrequestreview-283982649)).  I need to investigate this more but a simple test of removing this condition didn't look like an improvement.

A side issue is this `t` rule gives two examples:

```
't' //partitiiv, raamatu-t and kapsas-t
```

However, in fact "kapsast" matches the `'st' (RV or LONGV)` case in this among (because the `among` picks the longest matching string, and the "a" before matches `RV`) to stems to "kapsa".  Looking at https://en.wiktionary.org/wiki/kapsas#Estonian many other forms also stem to "kapsa" so this seems OK and I'm going to adjust the comments to reflect the reality here.

I've generated a list of Estonian words from a wikipedia dump and tested the as-implemented vs as-documented "-t" suffix removal and it does seem the as-documented is a little better (mostly in cases where it results in conflation with a form with a "-te" suffix which gets removed).  The only cases I've spotted which it seemed to make worse were shorter words, so I'm proposing to treat "-t" as documented, but add an "at least 4 characters" check.  This is a bit arbitrary, but the same threshold is used in a couple of places in the stemmer already, and it seems to work well here too.

Issue #115, opened 2019-09-23
release a python wheel


graingert:

mitya57:
A wheel is now available here: <https://pypi.org/project/snowballstemmer/#files>.

mitya57:
@Mogost The wheel is built from the same commit as the tarball (6169ef2456afeab15d53748627c139d7d022149e).

Mogost:
@mitya57 thank you

graingert:
@mitya57 can you release a .post0 build with the new wheel?

mitya57:
@graingert I think I will do 1.9.2 instead, a bit later.

ojwb:
Changes merged, but actually uploading a wheel is needed.

> I think I will do 1.9.2 instead, a bit later.

FWIW, I'm attempting a push to actually get 2.0.0 out, though something urgent may crop up elsewhere needing my attention.

graingert:
It would be useful to get .post versions with wheels of all the past
releases

mitya57:
> It would be useful to get .post versions with wheels of all the past releases

Why would you need it?

mitya57:
I have just uploaded 2.0.0 release with a wheel.

@ojwb Please close this issue.

PR #116, opened 2019-09-23
configure python universal wheels Refs #115

graingert:


graingert:
this will also create a .whl file which will need to be uploaded to pypi

GNUmakefile
```diff
@@ -432,7 +432,7 @@ dist_libstemmer_python: $(PYTHON_SOURCES)
 	cp -a $(PYTHON_SAMPLE_SOURCES) $${dest}/src/$(python_sample_dir) && \
 	cp -a $(PYTHON_RUNTIME_SOURCES) $${dest}/src/$(python_runtime_dir) && \
 	cp -a $(PYTHON_PACKAGE_FILES) $${dest} && \
-	(cd $${dest} && $(python) setup.py sdist && cp dist/*.tar.gz ..) && \
+	(cd $${dest} && $(python) setup.py sdist bdist_wheel && cp dist/*.{tar.gz,whl} ..) && \
```
graingert:
```suggestion
	(cd $${dest} && $(python) setup.py sdist bdist_wheel && cp dist/snowballstemmer-*{.tar.gz,-py2.py3-none-any.whl} ..) && \
```

graingert: might be better to be more specific with the dist filenames

graingert:
@ojwb

GNUmakefile
mitya57: The syntax with curly brackets is a bashism. It won't work when make uses another shell (e.g. dash).

```suggestion
	(cd $${dest} && $(python) setup.py sdist bdist_wheel && cp dist/*.tar.gz dist/*.whl ..) && \
```

python/setup.cfg
```diff
@@ -0,0 +1,2 @@
+[bdist_wheel]
+universal=1
```
mitya57: You need to also add this file to `PYTHON_PACKAGE_FILES` in GNUmakefile.

python/setup.py
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env python
 
-from distutils.core import setup
+try:
+    from setuptools import setup
+except ImportError:
+    from distutils.core import setup
```
mitya57: Is this fallback really needed? Most Python modules just use setuptools nowadays.

graingert: @mitya57 almost certainly not needed anymore, but most people who still use `distutils.core.setup` are ideologically opposed to removing it as a fallback

mitya57: I think distutils was here only for historical reasons.

ojwb:
> @ojwb

Contentless pings are just annoying.  If you really want my attention, at least go to the effort of telling me why.

@mitya57 Are you happy with this now?

mitya57:
Yes, looks good to me now.

PR #117, opened 2019-10-01
[python] Add changelog entry for 1.9.2 PyPI release

mitya57:
I am going to use this changelog entry for a new upload to PyPI that I will do <s>in a few hours</s> tomorrow, 2019-10-02.

### DIFF
--- a/python/CHANGES.rst
+++ b/python/CHANGES.rst
@@ -1,3 +1,12 @@
+snowballstemmer 2.0.0, 2019-10-02
+=================================
+
+* Simplified generated code for ``repeat`` and ``atleast`` commands.
+* Implemented “go grouping” optimisation.
+* Removed caching layer (#114).
+* Enabled building wheels (#115).
+* Updated package README.
+
 snowballstemmer 1.9.1, 2019-09-05
 =================================
 


mitya57:
Thinking more about this, maybe it’s better to not merge this at all, and instead make Python use the “global” NEWS file that you created in 7c776ff114f76767.

If you want, I can make a separate PR for that, or update this one.

ojwb:
I'm very close to releasing 2.0.0 (hopefully today, but almost certainly this week if not) so maybe waiting for that makes more sense.

Currently the global NEWS collapses changes made to backends and languages added since the last snowball.tartarus.org release (so it has "Add Python generator (#24)" but doesn't list all the Python enhancements and fixes since that) because it would be even more unreadably long otherwise, but that is perhaps unhelpful here.  I could explicitly include the Python changes since 1.9.1, or you could make this 1.9.2 release (which would be pretty much the same as 2.0.0), or include a Python-specific changelog when you package 2.0.0.

ojwb:
@mitya57 2.0.0 is now tagged (v2.0.0) and pushed.  I need to update the website and announce it, etc but you should be able to update PyPI with it.

mitya57:
> 2.0.0 is now tagged (v2.0.0) and pushed.

Wow, thanks a lot!

What do you think about including the Python-specific changelog in 2.0.0 PyPI upload, but replacing it with the global changelog for next release?

Also maybe we should use libstemmer_python_README as package description on PyPI?

ojwb:
> What do you think about including the Python-specific changelog in 2.0.0 PyPI upload, but replacing it with the global changelog for next release?

Sounds OK to me.

> Also maybe we should use libstemmer_python_README as package description on PyPI?

Sure, if it's not too long for what's expected.

BTW, is the "pass through" approach to pystemmer compatibility implemented here how Python users would expect something like that to work?

It occurred to me that maybe it would be better to instead suggest people wanting to support use of either module write their import code like this:

```
try:
    import Stemmer
except ImportError:
    import snowballstemmer as Stemmer
```

Then users only need to install either pystemmer or snowballstemmer, whereas currently you have to install both if you want pystemmer to be used.

Some slight tweaks would also be needed to how the API compatibility is implemented as well, but nothing major.

mitya57:
> BTW, is the "pass through" approach to pystemmer compatibility implemented here how Python users would expect something like that to work?
>
> It occurred to me that maybe it would be better to instead suggest people wanting to support use of either module write their import code like this:
> ```python
> try:
>     import Stemmer
> except ImportError:
>     import snowballstemmer as Stemmer
> ```
>
> Then users only need to install either pystemmer or snowballstemmer, whereas currently you have to install both if you want pystemmer to be used.

I have no hard preference, but your idea makes sense to me — installing just one library would be better than two.

However Python metadata does not allow to specify alternative dependencies (unlike e.g. Debian) — so a PyPI package will have to depend on something anyway, and I think most developers will choose snowballstemmer because it is always available.

So maybe we should *both* suggest to write code like above, and keep the “pass through” approach?

PR #119, opened 2019-10-03
Python packaging updates

mitya57:
I have already uploaded 2.0.0, so this is material for the next release.

- Python changelog is dropped, the global NEWS file is included instead.

- The contents of `doc/libstemmer_python_README` are used as package long description.
  I have uploaded it to the test server so you can see [here](https://test.pypi.org/project/snowballstemmer/) how it looks like.

- The license text is rendered incorrectly (as blockquote) so I have fixed it to render as a list.
  However, maybe we can drop the license text completely? The Python tarballs already include a COPYING file and proper metadata. What do you think?


ojwb:
> However, maybe we can drop the license text completely? The Python tarballs already include a COPYING file and proper metadata. What do you think?

Good idea - I've moved Yoshiki Shibukawa's copyright statement to COPYING instead.

PR #123, opened 2019-11-20
Javascript fixes

forivall:
This allows the javascript to work in a way that is slightly more in line with how node.js developers expect, with minimal changes. (although I do have other disagreements with the structure of the generated javascript code, but I can leave that to contributing a generator for assemblyscript)

Upon reading the contributing guidelines, i notice the suggestion about not changing existing code's formatting - i'll reduce this PR to just commit 73bcf12 if that's desired.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@
 /stemwords
 /dist
 /java/org/tartarus/snowball/ext/
+/javascript/*-stemmer.js
+!/javascript/base-stemmer.js
 /js_out
 /python_check
 /python_out
--- a/compiler/generator_js.c
+++ b/compiler/generator_js.c
@@ -1168,7 +1168,7 @@ static void generate_class_end(struct generator * g) {
     w(g, "~Mreturn base.getCurrent();~N");
     w(g, "~-~M};~N");
     w(g, "~-};~N");
-    /* w(g, "window['~n'] = ~n;~N"); */
+    w(g, "if (typeof module === 'object' && module.exports) module.exports = ~n;~N");
 }
 
 static void generate_among_table(struct generator * g, struct among * x) {
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -1,0 +1,8 @@
+To generate a javascript stemmer for use with `node ./stemwords.js` in this
+directory, run the following:
+
+```bash
+./snowball -p "(require('./base-stemmer'))" ./algorithms/english.sbl -js -o javascript/english-stemmer
+```
+
+Replace "english" with the target language as needed.
--- a/javascript/base-stemmer.js
+++ b/javascript/base-stemmer.js
@@ -2,11 +2,11 @@
 BaseStemmer = function() {
     this.setCurrent = function(value) {
         this.current = value;
-	this.cursor = 0;
-	this.limit = this.current.length;
-	this.limit_backward = 0;
-	this.bra = this.cursor;
-	this.ket = this.limit;
+        this.cursor = 0;
+        this.limit = this.current.length;
+        this.limit_backward = 0;
+        this.bra = this.cursor;
+        this.ket = this.limit;
     };
 
     this.getCurrent = function() {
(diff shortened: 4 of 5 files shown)


ojwb:
That "Please resist reformatting existing code" comment is really about the C code in the compiler - currently there's a lot of duplication between `compiler/generator*.c` which will be easier to eliminate if people avoid reformatting individual files.

Reformatting non-C code is OK so long as it's not mixed with functional changes (so top marks on that).

However this PR fails CI (https://travis-ci.org/snowballstem/snowball/jobs/614730939#L643) so I think more updates are needed to complete these changes.

> although I do have other disagreements with the structure of the generated javascript code

That's probably partly due to my inexperience with modern javascript and partly due to the history.  Snowball's javascript backend was originally a JSX backend (not the JSX you're likely thinking of, but https://jsx.github.io/ which was "a faster, safer, easier JavaScript" that transpiled to javascript).  The JSX project is inactive, so I morphed it into producing javascript code directly, with the primary motivation being to have a demo on the website which could work without any server-side support.

forivall:
Ah! that explains things. I noticed a few comments in some stemmers floating around on npm that they were from the "generated using the snowball to jsx compiler", thanks for that background. I'll take a look at CI error (`make check_js` failing) too.

I'd also like to contribute a `BUILDING` document that covers what commands are in the makefile.

ojwb:
It would also be helpful if the generated JS worked as-is with both node and in the browser.  Currently it seems we have to adjust the output for the website build (possibly that's partly because it gets run through `closure-compiler` there, which significantly reduces the size of the javascript the browser has to fetch - without that it would be over a megabyte, with it's 247K).  Maybe node and browsers have diverged too much now though.

> I'd also like to contribute a `BUILDING` document that covers what commands are in the makefile.

I worry that adds maintenance work.  What's wrong with just directing people to use the makefile to build?  If someone really wants to know how to do it by hand then the commands can be found in the makefile (there's a little extra work to extract them, but they can be confident that those commands works because CI tests them for every push, whereas commands in a new `BUILDING` document would easily get out of step).

ojwb:
It would also be helpful if the generated JS worked as-is with both node and in the browser.  Currently it seems we have to adjust the output for the website build (possibly that's partly because it gets run through `closure-compiler` there, which significantly reduces the size of the javascript the browser has to fetch - without that it would be over a megabyte, with it's 247K).  Maybe node and browsers have diverged too much now though.

> I'd also like to contribute a `BUILDING` document that covers what commands are in the makefile.

I worry that adds extra maintenance work.  What's wrong with just directing people to use the makefile to build?  If someone just want the files built that's simpler and less error-prone than trying to follow a document, and if someone really wants to know how to do it by hand then the commands can be found in the makefile.  There is a little extra work required to extract them, but in exchange one can be confident that those commands work because CI tests them for every push, whereas commands in documentation would easily get out of step.

.gitignore
```diff
@@ -11,6 +11,8 @@
 /stemwords
 /dist
 /java/org/tartarus/snowball/ext/
+/javascript/*-stemmer.js
+!/javascript/base-stemmer.js
```
ojwb, line 15: The generated sources are actually put in `js_out/` (and `NODE_PATH` is set so both directories are searched) so these ignore rules shouldn't be needed.

ojwb:
I've peeled out the whitespace change but extended it to eliminating all the uses of tabs for indentation in javascript files and pushed that.  I've also clarified the anti-reformatting remark in the contributing document to say "C code".

The other changes here definitely need more work (since CI currently fails because of them) but I don't think I know enough about node to resolve that effectively.

I'd also be very happy to sort out the generated Javascript to work well out of the box both with node and in browser (or if that's not really possible, to provide a simple way to generate a version for each).

ojwb:
The changes here break CI so this can't be merged as is.

There's been no progress for about 3 years so I'm going to close.  If you (or anyone) is interested in improving the situation please say and I will reopen (or open a new PR if that makes more sense).  To reiterate what I said above, I'd be very happy to sort out the generated Javascript to work well out of the box both with node and in browser (or if that's not really possible, to provide a simple way to generate a version for each).

forivall:
Yeah, I lost my need to use snowball, so i'm unlikely to pick this back up.

PR #125, opened 2019-11-25
Added algorithm for Faroese language.

brandurjacobsen:
I used the Scandinavian algorithms as inspiration.

### DIFF
--- a/algorithms/faroese.sbl
+++ b/algorithms/faroese.sbl
@@ -1,0 +1,79 @@
+routines (
+  mark_regions
+  strip_suffix
+  strip_t
+)
+
+externals ( stem )
+
+integers ( p1 x )
+
+booleans ( ccend )
+
+groupings ( v )
+
+stringescapes {}
+
+
+/* Special characters. "'" indicates acute */
+
+stringdef a'    '{U+00E1}'
+stringdef ae    '{U+00E6}' // ash
+stringdef i'    '{U+00ED}'
+stringdef d/    '{U+00F0}' // eth
+stringdef o'    '{U+00F3}'
+stringdef o/    '{U+00F8}' // strikethrough "o"
+stringdef u'    '{U+00FA}'
+stringdef y'    '{U+00FD}'
+
+/* Vowels */
+
+define v 'a{a'}ei{i'}o{o'}u{u'}y{y'}{ae}{o/}'
+
+define mark_regions as (
+    $p1 = limit
+    unset ccend
+    test ( hop 3 setmark x )
+    try ( goto v gopast non-v setmark p1 )
+    try ( $p1 < x  $p1 = x )
+    backwards substring among
+      ('kt' 'gt' 'vt' 'mt' 'nt') (set ccend) ($p1 = x)
+)
+
+backwardmode (
+
+    define strip_suffix as (
+        setlimit tomark p1 for ([substring])
+        among(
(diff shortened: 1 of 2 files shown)


ojwb:
I told the CI to retry the automated tests after you fixed the name of the expected output, but the tests are still failing because of case differences in the expected output:

https://travis-ci.org/snowballstem/snowball/jobs/616560624#L605

The stemming test programs (`stemwords` for C, etc) lowercase ASCII in the input, but I guess you made the output file in a way that preserves the case?

You can run the tests for various languages locally if you have `snowball` and `snowball-data` checked out in sibling directories - `make check` for the C tests, `make check_java` for the Java tests, etc.

brandurjacobsen:
I've created new files for the tests, and verified that 'make check' runs as expected.
Also made some small changes to the algorithm.

ojwb:
> I've created new files for the tests, and verified that 'make check' runs as expected.

Thanks - looking better now.

> Also made some small changes to the algorithm.

It's a bit surprising to just going off a 3 character count from the start rather than looking at vowels and non-vowels, but if that works well for Faroese then simple has definite benefits.

The website really needs updating too (see `CONTRIBUTING.rst`).  If this algorithm is one you came up with just for Snowball and there's no paper to cite it would be useful to have some notes on the design decisions - for example, why just `hop 3` rather than looking at vowels and non-vowels?  And if there are suffixes that aren't removed because they would trigger in unwanted cases that's also very useful to have some notes on.  These sort of details are much harder for us to deduce later if we aren't able to contact you and somebody proposes a patch to modify the algorithm.

brandurjacobsen:
Sorry for the delay.
Since I haven't found any justification (in the course of my limited research) for setting the `limit` for suffix stripping according to a particular vowel/non-vowel pattern I decided to just set the limit at the third character. This has nothing to do with Faroese as such. Perhaps you can direct me to some resources that would dissuade me to do this? 

The algorithm is not based on any paper, only eyeballing.
I tried to get comments on the algorithm in a mailing-list where some members are linguists and Faroese native speakers, but to no avail. But I got some comments like "it looks good". 

I will make some comments on the algorithm on the website soon (hopefully).

ojwb:
This idea of counting based on vowel/non-vowel runs comes from Martin Porter's original English stemmer paper (https://tartarus.org/martin/PorterStemmer/def.txt), where he calls it "measure":

>>>
The algorithm is careful not to remove a suffix when the stem is too short,
the length of the stem being given by its measure, m. There is no linguistic
basis for this approach. It was merely observed that m could be used quite
effectively to help decide whether or not it was wise to take off a suffix.
>>>

For example, the following English words vary significantly in the number of characters, but all have the same measure, and so are handled in a similar way by the stemmer:

```
BAKING
BACKING
BOOKING
BAULKING
BLANKING
BLANCHING
FREIGHTING
SCROOTCHING
```

> The algorithm is not based on any paper, only eyeballing.
> I tried to get comments on the algorithm in a mailing-list where some members are linguists and Faroese native speakers, but to no avail. But I got some comments like "it looks good".

That's OK - it's useful to know when there is a paper.

The key thing is to capture any information that would be useful to have if in (say) 5 years time somebody suggests a change to the algorithm and we can't contact you.

ojwb:
@brandurjacobsen I'm not sure why this got closed - for some reason github isn't showing a reason.  I don't think I closed it but sorry if I accidentally did.  I only just noticed it while looking for a different closed PR.  Happy to work towards merging it if you're still interested.

PR #126, opened 2019-12-14
[python] Remove Lithuanian from classifiers blacklist

mitya57:
See https://github.com/pypa/warehouse/issues/7047#issuecomment-560366145.


ojwb:
Thanks.  Should we be seeing if we can get the others added too?

mitya57:
Ok, asked in the same bug.

PR #127, opened 2020-01-30
Pass -Wl,-dead_strip,-bind_at_load on macOS

rex4539:
From `man ld`:
```
-dead_strip
                 Remove functions and data that are unreachable by the entry point or exported symbols.
```
```
-bind_at_load
                 Sets a bit in the mach header of the resulting binary which tells dyld to bind all symbols when the binary is loaded,
                 rather than lazily.
```
Before:
```
% wc -c snowball 
 211964 snowball

% wc -c stemwords
 757096 stemwords
```
After:
```
% wc -c snowball 
 211852 snowball

% wc -c stemwords 
 756776 stemwords
```

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -153,6 +153,14 @@ JAVA_RUNTIME_CLASSES=$(JAVARUNTIME_SOURCES:.java=.class)
 CFLAGS=-O2 -W -Wall -Wmissing-prototypes -Wmissing-declarations
 CPPFLAGS=-Iinclude
 
+# pass -dead_strip and -bind_at_load on macOS to shave some binary size
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S),Darwin)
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks for the patch, but these seem really negligible savings - just 112 and 320 bytes (and your patch adds 254 bytes to the makefile's size!)

I'm not convinced those are worth complicating things for - it's easy enough already to specify any extra flags you want to use - e.g.:

```
gnumake LDFLAGS=-Wl,-dead_strip,-bind_at_load
```

Also once we sort out producing a shared library instead of a static one, these options presumably won't reduce the size at all (I'm assuming this is eliminating code which isn't called in a particular static linked case - if there's genuinely dead code we ought to eliminate it for everyone, not use linker options to achieve that only on a single platform).

rex4539:
Fair enough. Thanks for reviewing.

PR #128, opened 2020-02-08
[python] Remove the classifiers blacklist

mitya57:
Basque, Irish and Nepali were added as possible natural languages to PyPI. See pypa/pypi-support#138.


ojwb:
Thanks for sorting that out.

PR #129, opened 2020-02-09
fix GNUmakefile dependencies on Windows

jannick0:
On Windows platforms executables have the extension '.exe'.  Compilers
like gcc append the extension if not provided in the '-o' flag.  A Windows
platform can easily be recognized by checking if the environment variable
OS equals 'Windows_NT'.

This commit adds the extension of built executables to prerequisites of
GNUmakefile rules, such that make can figure out the dependencies.

NOTE: Extensions are not added to executable names when called by a shell
(e.g., cmd, sh) which are usually able to call the executable even when
the name is provided without extension.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,11 +81,15 @@ before_install:
   # CI tested easily.
   #
   # If that fails, just use the standard snowball-data repo's default branch.
+  #
+  # We currently skip this step on Microsoft Windows as checking out the
+  # data repo seems to stall indefinitely, so for now we just do the build
+  # there and don't run any tests.
   - GH_BRANCH=${TRAVIS_PULL_REQUEST_BRANCH:-$TRAVIS_BRANCH}
   - GH_REPO_SLUG=${TRAVIS_PULL_REQUEST_SLUG:-$TRAVIS_REPO_SLUG}
   - GH_REPO_URL=https://github.com/${GH_REPO_SLUG%%/*}/snowball-data.git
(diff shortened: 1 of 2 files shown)


ojwb:
This looks reasonable, though without CI coverage I worry how long it would actually stay working for as this isn't a platform I use myself.

Would you be up for adding a CI build?  Apparently travis now supports it: https://docs.travis-ci.com/user/reference/windows/

jannick0:
> Would you be up for adding a CI build? Apparently travis now supports it: https://docs.travis-ci.com/user/reference/windows/
I can try. This might pile up to a commit series while testing which I'll will ultimately squash. ... just as a heads-up. ;)

ojwb:
Hmm, it seems to hit a 10 minute timeout trying to clone the snowball-data repo, even with `--depth=1`.  I'm guessing git on this platform is much slower with the large files in this repo.

I wonder if downloading the branch we want as a ZIP file and unpacking it would be a better option.  It looks like we didn't have the `.git` on `GH_REPO_URL` we'd could fetch `$GH_REPO_URL/archive/$GH_BRANCH.zip` to get the branch we want.

 Thoughts?

jannick0:
Well, the issue rather is that travis' Windows support is beta.

ojwb:
I managed to get this to work, and even passing the C tests.  It looks like the git clone issue got solved in the meantime.

None of the other languages get tested currently, but happy to take patches to address that (notably there's special handling for csharp in the changes here which doesn't get tested).

PR #136, opened 2020-06-08
fix leak

ObiWahn:


ObiWahn:
Hey @ojwb I am not sure why the go-tests fail. They already fail for me locally using the official master. Can you confirm this or has it something to do with my PR? In case it works for you, I am interested in the the exact commands and environment variables.

compiler/driver.c
```diff
@@ -140,8 +140,19 @@ static int read_options(struct options * o, int argc, char * argv[]) {
...
+                char * new_name = NULL;
+                size_t len = 0;
+
                 check_lim(i, argc);
-                o->name = argv[i++];
+                name = argv[i++];
+
+                len = strlen(name);
+                new_name = malloc(len);
+                memcpy(new_name, name, len);
```
mitya57, line 153: These three lines can be replaced with `strdup` usage, can't they? Also `strdup` will make sure the new string ends with null byte, which your code doesn't seem to do.

ObiWahn: @mitya57 can `strdup` somehow be used with `-std=c90`?

ojwb: I'd probably avoid `strdup()` - it's specified by POSIX but not part of ISO C so it potentially limits portability for marginal convenience.

ojwb:
It's true that we leak memory in the sense that we don't ever free `o->name` and in some situations it is allocated with `malloc()`.

However the snowball compiler exits shortly afterwards at which point the OS will reclaim all memory from the process anyway so the only real consequence of a one-off allocation like this never being freed I can see is a false positive if you use a leak-finding tool on the code.

If people actually care about that, I guess I'm OK to take a clean patch to address it.

ojwb:
The go tests look OK to me: https://travis-ci.org/github/snowballstem/snowball/jobs/696243743

Not sure why the CI results aren't getting shown in the PR any more though.

ObiWahn:
@ojwb we need it for our asan build. Snowball comes as part of the source. It is build and used during the build-process. The memory sanitizer will change the exit code to non 0 when the leak is triggered. Which will make our build fail. I guess it would be somehow possible to fix the build of the iresearch lib. But that is almost like a 3rd party lib for the bigger parts of the company and it seemed easier to fix snowball. Is c90 really required? That is 30 years old.

ojwb:
> we need it for our asan build. Snowball comes as part of the source. It is build and used during the build-process. The memory sanitizer will change the exit code to non 0 when the leak is triggered. Which will make our build fail. I guess it would be somehow possible to fix the build of the iresearch lib. But that is almost like a 3rd party lib for the bigger parts of the company and it seemed easier to fix snowball.

I'd just set up a suppression for snowball - see: https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer#suppressions

Note you can just match on the binary name, so:

```
$ make CC=clang-10 CFLAGS='-fsanitize=address'
[...]
==1513933==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 18 byte(s) in 1 object(s) allocated from:
    #0 0x496a6d in malloc (/home/olly/git/snowball/snowball+0x496a6d)
    #1 0x4f0f39 in read_options (/home/olly/git/snowball/snowball+0x4f0f39)
    #2 0x4ed0c6 in main (/home/olly/git/snowball/snowball+0x4ed0c6)
    #3 0x7fe28f031e0a in __libc_start_main /build/glibc-WZtAaN/glibc-2.30/csu/../csu/libc-start.c:308:16

SUMMARY: AddressSanitizer: 18 byte(s) leaked in 1 allocation(s).
make: *** [GNUmakefile:216: src_c/stem_UTF_8_arabic.h] Error 1
$ echo leak:snowball > suppr.txt
$ LSAN_OPTIONS=suppressions=suppr.txt make CC=clang-10 CFLAGS='-fsanitize=address'
/snowball algorithms/basque.sbl -o src_c/stem_UTF_8_basque -eprefix basque_UTF_8_ -r ../runtime -u
-----------------------------------------------------
Suppressions used:
  count      bytes template
      1         18 snowball
-----------------------------------------------------

[...]
```

You can just add a line to the suppression file for any other build tools where leaks on exit aren't relevant, which is a lot less work than having to come up with patches and push them upstream.

> Is c90 really required? That is 30 years old.

It's not really the age of C90 that's relevant, but the age of what we'd use instead (so ~21 years if you think it should be C99; ~9 years for C11; ~2 years for C18).

We do currently aim for C90, though mostly because there's not so far been a compelling reason to require anything newer.

I don't have strong feelings about sticking to C90, but I think it's something that we ought to consult the user base about before changing given that snowball by its nature is likely embedded into build systems.

But I don't think any newer version of the C standard specifies `strdup()` anyway.

ObiWahn:
> You can just add a line to the suppression file for any other build tools where leaks on exit aren't relevant, which is a lot less work than having to come up with patches and push them upstream.

Good point! Thank You! Maybe you can still apply the patch, now that it as been created.

> But I don't think any newer version of the C standard specifies `strdup()` anyway.

I saw that it has been recently added to the ISO: 
https://en.cppreference.com/w/c/experimental/dynamic/strdup

ojwb:
> > But I don't think any newer version of the C standard specifies strdup() anyway.

> I saw that it has been recently added to the ISO:
> https://en.cppreference.com/w/c/experimental/dynamic/strdup

That's [C2x](https://en.wikipedia.org/wiki/C2x) which isn't yet a standard (looks like it's currently targetting release in 2021).  I think it would definitely be too aggressive to require a standard which is -1 years old!

compiler/driver.c
```diff
@@ -567,6 +579,7 @@ extern int main(int argc, char * argv[]) {
             lose_b(p->b); FREE(p); p = q;
         }
     }
+    free((void*)o->name);
```
ojwb, line 582: Sorry, this is just too nasty.

I think you're doing this to cast away `const`-ness so the pointer can be passed to `free()`, but really if we want to be able to call `free()` we shouldn't make the pointer `const` to start with.  That way if somebody tries to do `o->name = "some_constant_string";` it'll fail to compile, but as you have this it'll compile and only fail at run-time, and only if that code path is actually taken.

ojwb:
As I said before I'm OK with a clean patch for this, but reviewing the current state I'm afraid I'm not really OK with this one.

Incidentally, I've since discovered there's an lsan option to disable this exit-time check completely:

```
LSAN_OPTIONS=leak_check_at_exit=0 
```

Leaks during the run (due to there being no pointers left to a particular allocation) should still be reported.

ojwb:
Closing in favour of #166.

PR #140, opened 2020-11-24
Yiddish changes

ojwb:
@urieli This fixes the implementation bugs I [mentioned](https://github.com/snowballstem/snowball-website/pull/20#issuecomment-724255567).  It also eliminates the uses of `hop` with a negative argument.

### DIFF
--- a/algorithms/yiddish.sbl
+++ b/algorithms/yiddish.sbl
@@ -11,6 +11,8 @@ routines (
        prelude
        mark_regions
        R1
+       R1plus1
+       R1plus3
        standard_suffix
 )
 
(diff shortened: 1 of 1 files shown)


ojwb:
In snowballstem/snowball-data#16 @urieli wrote:
> In almost all cases, I prefer the stemming prior to the change. I'm going to have to take some time to review snowballstem/snowball#140 and figure out the correct handling for -endik vs -ndik

Presumably there are reasons why you tried to handle -ndik and -endik differently in the first place, but if you did conclude the current behaviour is the best option then we could just reduce the -endik handling to:

```
        '{Ayen}{Nun}{Dalet}{Yud}{Kuf}'
           ( R1 delete )
```

urieli:
Included your commits + my review in separate PR: #142

ojwb:
Thanks, closing this in favour of #142.

PR #142, opened 2021-01-13
Yiddish review

urieli:
This extends and replaces #140

### DIFF
--- a/algorithms/yiddish.sbl
+++ b/algorithms/yiddish.sbl
@@ -11,6 +11,8 @@ routines (
        prelude
        mark_regions
        R1
+       R1plus1
+       R1plus3
        standard_suffix
 )
 
@@ -125,8 +127,7 @@ define mark_regions as (
 
         // If verbal prefix followed by Tsu- or Ge-, replace it
         (
-
-          (('{Tsadek}{Vov}{Giml}{Nun}' or '{Tsadek}{Vov}{Kuf}{Tes}' or '{Tsadek}{Vov}{Kuf}{Nun}') not (hop 1) hop -4)
+          test (('{Tsadek}{Vov}{Giml}{Nun}' or '{Tsadek}{Vov}{Kuf}{Tes}' or '{Tsadek}{Vov}{Kuf}{Nun}') atlimit)
           or
           ( ['{Giml}{Ayen}'] <- 'GE')
           or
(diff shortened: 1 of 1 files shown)


algorithms/yiddish.sbl
```diff
@@ -129,6 +130,8 @@ define mark_regions as (
         (
           test (('{Tsadek}{Vov}{Giml}{Nun}' or '{Tsadek}{Vov}{Kuf}{Tes}' or '{Tsadek}{Vov}{Kuf}{Nun}') atlimit)
           or
+          test ('{Giml}{Ayen}{Beys}{Nun}')
+          or
           ( ['{Giml}{Ayen}'] <- 'GE')
```
ojwb, line 135: Should there also be an exception for `'{Giml}{Ayen}{Lamed}{Tes}'` here too?

(Maybe that just isn't a situation that occurs in valid words - I'm afraid I know very little about the Yiddish language...)

algorithms/yiddish.sbl
```diff
@@ -159,9 +162,7 @@ define mark_regions as (
 
 backwardmode (
   define R1 as $p1 <= cursor
-  // Like R1, but also allows the cursor to be outside R1 by the width of Tes.
-  define R1plus1 as $p1 <= cursor + sizeof '{Tes}'
-  // Like R1plus1, but by the width of Giml Yud Samekh
+  // Like R1, but by the width of Giml Yud Samekh
```
ojwb, line 165: I think this needs to incorporate the first part of the comment about R1plus, or else the meaning isn't clear without looking at the history.  I'll push a tweak for that.

ojwb:
I've merged the current state of this as it seems reasonable and merging means I can more easily work on getting the PR to add an Ada backend merged too, but it would be good to get an answer to my query about `'{Giml}{Ayen}{Lamed}{Tes}'` above.

algorithms/yiddish.sbl
urieli: No, but I should have commented it.

Here Yiddish is acting like a Germanic language (which it is mostly, at least from a morphological point of view). The regular past participle starts with the prefix "ge-", as in "ge-zungen" for the verb "zingen" (to sing).
In the vast majority of cases, it's valid to remove this "ge-". However, there are some common exception, including "gebn" (to give) and "gelt" (money). There are certainly some other exceptions that I could add with time.

However, after the verbal particles, you will never find "gelt" as it isn't a verb.
You could have "avek-gebn" (to give away), but never "avek-gelt".

Also, in prefixed verbs, the "ge-" past participle marker comes after the particle, giving, "avek-ge-gebn" for "gave away". So we only include the "gebn" exception in past participles, not the "gelt".

The form "avek-tsu-gebn" is another particularity (possibly it exists in German as well, but I don't speak any German!). The preposition "tsu" (to) is inserted inside a verb with a particle prefix, e.g. "avek-tsu-gebn" (to give away).

mitya57: The form "avek-tsu-gebn" is another particularity (possibly it exists in German as well, but I don't speak any German!)

You are right, it exists: [wegzugeben](https://de.wiktionary.org/wiki/wegzugeben).

ojwb: Thanks - I see #143 adds a comment about this.

PR #144, opened 2021-03-04
add Ukrainian lang

tggo:


ojwb:
Looks like you pushed this branch before the corresponding snowball-data branch (the instructions do say to push snowball-data first) and the CI faieds because the new testdata wasn't there yet.

I'll trigger it to rerun.

algorithms/ukrainian.sbl
```diff
@@ -0,0 +1,163 @@
...
+stringdef f    '{U+0444}'
+stringdef kh   '{U+0445}'
+stringdef ts   '{U+0446}'
+stringdef ch   '{U+0447}'
+stringdef sh   '{U+0448}'
+stringdef shch '{U+0449}'
+stringdef soft '{U+044C}'
+stringdef iu   '{U+044E}'
+stringdef ia   '{U+044F}'
+stringdef apostrophe '{U+027F}'
```
jimregan, line 38: should be '{U+0027}', U+027F is ɿ

ojwb: Well spotted - I've applied that.

Since the differences all seemed to involve apostrophes it could well explain the mismatching test results (I'm guessing that typo was introduced after generating the expected output file).  CI should be working again (aside from the csharp build which is currently failing to run properly on master).

ojwb:
With the apostrophe definition fix CI passes for everything except csharp, which is currently broken in master too.  We have other wide character Unicode programming languages passing so I don't expect this hides a problem but I'll try to check it locally and merge the branch if it is indeed OK.

stefanvodita:
What would it take to have this merged?

ojwb:
@stefanvodita Thanks for your interest.

I've fixed CI on master and pushed a fix for a minor nit here (the new `modules.txt` entry wasn't in sorted order) which has demonstrated that the C# failure wasn't anything to do with these changes.  This PR looks good now but needs the other two to be ready as well - I'll make some comments on those and @-mention you.

ojwb:
I'm going to close this in favour of #178 as that's been actively worked on.

Issue #146, opened 2021-04-05
Thread-safe version for C# / Moving the state out of the Stemmer instance


Turnerj:
Currently from the generated code for C#, the stemmer uses an internal state to keep track of position etc. Unfortunately this prevents re-use of the same stemmer instance across multiple threads. Creating a new instance of a stemmer per thread (or in a web application, per request) means the constructor for the stemmer (with all the `Among`-type arrays generated) occurs frequently, generating the same data, creating additional GC pressure.

In doing this, it should allow for the ability to also switch to using `Span<char>` to do a lot of the work where you have `StringBuilder`, avoiding further allocations too.

I'm a C# dev, not a C dev, so while I want to help wherever I can, I would be somewhat limited in my capacity to help in the generational side of things.

Turnerj:
Looking at the compiler more closely, the `Among`-type declarations _could_ be pushed to a static properties with a static constructor (thus are shared across instances of the same stemmer) which would relieve some pressure from creating new instances of a stemmer.

https://github.com/snowballstem/snowball/blob/8b49140e5cb6675c58bd908dbd4c7e12146a4f72/compiler/generator_csharp.c#L1194

https://github.com/snowballstem/snowball/blob/8b49140e5cb6675c58bd908dbd4c7e12146a4f72/compiler/generator_csharp.c#L1239

ojwb:
Please open a PR if you want to propose a change - then the CI will test it, and I won't have to ham-fistedly try to interpret what you're suggesting through the lens of my very limited grasp of C#!

Turnerj:
Hahaha - no problem, I'll give it a shot.

ojwb:
Cool.

FWIW, making the data structures such as the `Among` stuff static sounds a good idea to me.

I wonder if you could inline all the snowball routines to give a single C# function so then the cursor, slice ends, limits, etc can just be local variables of that (as indeed could the Snowball variables).  Recursion couldn't be handled, but I've never seen a Snowball stemmer which was recursive so in practice that isn't really a limitation.  It's a bit ugly to only support a subset of the language for a particular backend though...

Turnerj:
Yeah, it should be possible - I've seen something similar in someone else's version of the Porter algorithm in C# (theirs was pretty allocate-y as they kept regenerating strings throughout the calculating).

When doing so, it should be possible with maybe 2 or 3 sets of allocations:
- Initially converting a `string` to a `char[]`
[...]

PR #147, opened 2021-04-08
Basic optimizations for C# stemmer output

Turnerj:
This moves the among values to static, defined at their declaration. Additionally sets the grouping declarations to be constant rather than static.

Related to discussions in #146


ojwb:
As you noted in #146, the first commit needs more work to support among conditions, but the second by itself works and makes sense so I've cherry-picked that.

I then realised we could only use the static among approach for amongs without functions, so I've pushed a tweak to do that to your branch - this gets us most of the benefits of this since most amongs don't use functions.  It passed `make check_csharp` locally, but I'll wait for CI and merge if it looks good.

PR #148, opened 2021-05-14
Honor build flags passed as command line arguments

DimStar77:


GNUmakefile
```diff
@@ -162,8 +162,8 @@ C_OTHER_OBJECTS = $(C_OTHER_SOURCES:.c=.o)
 JAVA_CLASSES = $(JAVA_SOURCES:.java=.class)
 JAVA_RUNTIME_CLASSES=$(JAVARUNTIME_SOURCES:.java=.class)
 
-CFLAGS=-O2 -W -Wall -Wmissing-prototypes -Wmissing-declarations
-CPPFLAGS=-Iinclude
+override CFLAGS += -W -Wall -Wmissing-prototypes -Wmissing-declarations
+override CPPFLAGS += -Iinclude
```
ojwb, line 166: This means optimisation is now off by default, which isn't good.  As well as making the code slower and (most likely) larger, some compiler warnings only work if optimisation is on.

The option which is really needed here is `-Iinclude` since without that the headers won't be found.  The others are desirable as defaults, but all are reasonable to override - you might want a different optimisation setting; you might be using a compiler which doesn't understand `-Wmissing-prototypes`.

We could just leave `CFLAGS` as it is and only have the `override CPPFLAGS += -Iinclude` change, but perhaps it'd be better to move `-Iinclude` in a new variable (`INCLUDES`, or `REQUIRED_CPPFLAGS` or something), and leave these as-is.

Thoughts?

DimStar77: > This means optimisation is now off by default, which isn't good. As well as making the code slower and (most likely) larger, some compiler warnings only work if optimisation is on.

It's OFF if the user/packager has a CFLAGS= defined in his env, Otherwise ON; If somebody overrides it, they should know what they do (most distros have their own set, the LTO enabled, stack-protector enabled and all such things.

 > The option which is really needed here is `-Iinclude` since without that the headers won't be found. The others are desirable as defaults, but all are reasonable to override - you might want a different optimisation setting; you might be using a compiler which doesn't understand `-Wmissing-prototypes`.
> 
> We could just leave `CFLAGS` as it is and only have the `override CPPFLAGS += -Iinclude` change, but perhaps it'd be better to move `-Iinclude` in a new variable (`INCLUDES`, or `REQUIRED_CPPFLAGS` or something), and leave these as-is.
> 
> Thoughts?

Moving he really required one to a different variable, which is not overridden, sounds actually good (then packagers could override the CFLAGS/CPPGLAGS based not he distro settings, and not having to care for the -Iinclude (which, in openSUSE rpm packaging, I had to of course inject again into CPPFLAGS)

ojwb: > It's OFF if the user/packager has a CFLAGS= defined in the env, Otherwise ON;

But your patch removes the `-O2` from `CFLAGS`, and looking at e.g. the Java CI build, there's no `-O2` in the `gcc` invocations now:

https://travis-ci.org/github/snowballstem/snowball/jobs/771174270

> Moving he really required one to a different variable, which is not overridden, sounds actually good (then packagers could override the CFLAGS/CPPGLAGS based not he distro settings, and not having to care for the -Iinclude (which, in openSUSE rpm packaging, I had to of course inject again into CPPFLAGS)

Yes, that was pretty much my thinking.  I'll make that change instead then.

PR #150, opened 2021-07-30
Use .a suffix for the static library

paper42:
.o is meant to be used for object files


ojwb:
Weird, it's been like that since the first commit (e9dd099a0ab5a63ba7d23ccb0b41bff0c0805fb6) and I'd never noticed.

This looks good to me, but I'd like to see CI pass before merging and for some reason CI hasn't run.  Might be due to the travis-ci.org migration though I thought I'd migrated everything.  I'll investigate.

PR #154, opened 2021-09-03
swedish: Remove -et or -en when stem ends in et

### LINKED ISSUE #47: Swedish stemmer
I think i have found a bug in the swedish stemmer. When searching for "mötet" (the meeting) i should get result for "möte" and "möten". I think the problem is when stemming words ending with "et". (words ending with "andet" and "het" should work though. Those endings are in the suffix list.

When searching for the longest suffix in the first step i added this suffix "et" and that works. Don't know if that is the right way to fix this though.

ojwb:
Removing -et and -en in general is problematic, as many words end in -et
or -en where this isn't a suffix, but very few end in -etet or -eten
where the last two letters aren't a suffix (and those that do don't seem
to suffer if we make the stem not have the -et).

Fixes #47

### DIFF
--- a/algorithms/swedish.sbl
+++ b/algorithms/swedish.sbl
@@ -1,5 +1,6 @@
 routines (
            mark_regions
+           R1
            main_suffix
            consonant_pair
            other_suffix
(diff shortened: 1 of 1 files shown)


ojwb:
Closing in favour on #215 which seems a better approach.

PR #155, opened 2021-09-07
New feature: Typescript compiler

znakeeye:
Derived a `Typescript` compiler from the existing `Javascript` compiler.

- Typescript compiler is happy + code is well-formatted.
- Verified for a few languages.

I believe `snowball` will greatly benefit from this. Please consider a merge after scrutinizing all my tweaks. Particularly the build scripts... I did my best trying to figure out all parameters and stuff 😄

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -28,6 +28,11 @@ python_sample_dir = sample
 js_output_dir = js_out
 js_runtime_dir = javascript
 js_sample_dir = sample
+
+ts_output_dir = ts_out
+ts_runtime_dir = typescript
+ts_sample_dir = sample
+
 NODE ?= nodejs
 
 cargo ?= cargo
@@ -69,6 +74,7 @@ COMPILER_SOURCES = compiler/space.c \
 		   compiler/generator_csharp.c \
 		   compiler/generator_java.c \
 		   compiler/generator_js.c \
+           compiler/generator_ts.c \
 		   compiler/generator_pascal.c \
 		   compiler/generator_python.c \
 		   compiler/generator_rust.c \
@@ -99,6 +105,10 @@ CSHARP_STEMWORDS_SOURCES = csharp/Stemwords/Program.cs
 JS_RUNTIME_SOURCES = javascript/base-stemmer.js
 
 JS_SAMPLE_SOURCES = javascript/stemwords.js
+
+TS_RUNTIME_SOURCES = typescript/BaseStemmer.ts
+
+TS_SAMPLE_SOURCES = typescript/stemWords.ts
 
 PASCAL_RUNTIME_SOURCES = pascal/SnowballProgram.pas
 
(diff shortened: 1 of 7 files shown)


typescript/stemWords.ts
```diff
@@ -0,0 +1,125 @@
+import * as fs from 'fs';
+
+function usage() {
+    console.log("usage: stemwords.js [-l <language>] -i <input file> -o <output file> [-c <character encoding>] [-p[2]] [-h]\n");
```
ojwb, line 4: `.js` here...

typescript/stemWords.ts
```diff
@@ -0,0 +1,125 @@
...
+    lines.push("");
+    fs.writeFileSync(output, lines.join('\n'), encoding);
+}
+
+function create(name: string) {
+    let lc_name = name.toLowerCase();
+    if (!lc_name.match('\\W') && lc_name != 'base') {
+        let algo = lc_name.substr(0, 1).toUpperCase() + lc_name.substr(1);
+        try {
+            const stemmer = require(lc_name + '-stemmer.js');
```
ojwb, line 117: `.js` here too ... which presumably means we're testing the JS stemmers, not the typescript ones!

compiler/generator_ts.c
```diff
@@ -0,0 +1,1308 @@
+
```
ojwb, line 1: This file is very similar to `generator_js.c` - from a maintenance point of view I think it would make more sense to share a single generator implementation between javascript and typescript and just parameterise the small number of differences - that way future fixes only need applying in one place rather than two (if there are two places to fix, inevitably one gets missed sometimes).

Also, it seems many of the differences are due to trying to format the generated code to match typescript code formatting conventions.  I'm not sure how worthwhile a goal that is - having generated code with a basically readable formatting to aid debugging generator problems is useful, but coding conventions almost universally aim at making code more maintainable, and the generated code inherently doesn't need to be maintained directly.

So I'd be inclined to just tell typescript to shut up about code formatting (with more/broader `eslint-disable` magic comments), but if that's not possible we could format the javascript output in the same way.

There's also the approach the Go generator takes which is to run the snowball `.go` outputs through `gofmt` to mechanically reformat them to match Go coding conventions.  That has the advantage that there's no snowball-specific work to do if new formatting checks get added.

GNUmakefile
```diff
@@ -489,6 +507,24 @@ dist_libstemmer_js: $(JS_SOURCES)
...
 
+dist_libstemmer_ts: $(TS_SOURCES)
+	destname=tsstemmer-$(SNOWBALL_VERSION); \
+	dest=dist/$${destname}; \
+	rm -rf $${dest} && \
+	rm -f $${dest}$(tarball_ext) && \
+	mkdir -p $${dest} && \
+	mkdir -p $${dest}/$(ts_runtime_dir) && \
+	mkdir -p $${dest}/$(ts_sample_dir) && \
+	cp -a doc/libstemmer_ts_README $${dest}/README.rst && \
```
ojwb, line 518: There's no `doc/libstemmer_ts_README` so `make dist` fails (hence the CI failures)

ojwb:
This PR still needs work, and #183 suggests a different approach:

> My feeling is that there's not much point having a separate TS build, as TS-compatible types can be supplied by JSDoc without any loss of JS compatibility.

Thoughts?  Should we close this PR in favour of taking that approach instead?

ojwb:
Taking that "thumbs up" as a yes, and closing.

ojwb:
FYI, git master now generates ES6 Javascript with ESM by default (and as the only option), and we're hopefully close to supporting typescript via jsdoc comments - perhaps even there now.

I've opened a new ticket to pull together various things which have been suggested here and in other issues and PRs, which includes that: #244

PR #158, opened 2021-11-15
[python] Update metadata to indicate that Python ≥ 3.3 is required

mitya57:
- No `universal=1` anymore — universal wheel means it works with both Python 2 and Python 3.
- `python_requires = >=3.3` — this will cause pip to install older versions of snowball on older versions of Python.
- Classifiers don't affect anything, they are just shown on PyPI. I just ran `python=python3.10 make check_python` locally and it passed, so we may advertise all versions up to 3.10 as supported.


ojwb:
We have announced that we're no longer actively supporting Python 2, but at this point that's really just a policy change and Snowball 2.2.0 should still work with Python 2.

I believe the Python 3.3 requirement is just because we generated Unicode string constants with a `u` prefix (e.g. `u"foo"`), but Snowball has **always** done that for Python (it's present in a29c164b59e079333e2d7f48de7af1367ac28ded which added the Python generator) so pip installing an older Snowball for Python 3.2 is not going to help as the user will still get a non-working installation.

mitya57:
Ok, understood. How about just updating the classifiers then?

ojwb:
> How about just updating the classifiers then?

Yes, that definitely makes sense - I've pushed that part.

If there's a good way to specify 2.x or >= 3.3 that is technically correct, but the consequence of pip then installing an older version isn't great as it seems liable to confuse.  I guess we can't retroactively fix the missing requirement specification for older versions though.

Presently I've essentially said we'd take clean patches for continuing Python 2 support if anybody cares sufficiently, but at some point we can give up on even that, drop the `u` prefix on string literals, and specify the Python requirement as `>= 3` because it then should work with any Python 3.x.

Not sure about the appropriate timing for "at some point".  There wasn't any feedback aside from yours on the list email re Python 2, but many users don't read the list, but the announcement has now been in a release which should mean more users are now aware of it.

PR #44, opened 2016-08-02
Add Greek stemmer

master:
Based on the algorithm described in http://sais.se/mthprize/2007/ntais2007.pdf
and http://tampub.uta.fi/bitstream/handle/10024/80480/gradu03463.pdf

Despite of some shortcoming (such as not handling diacritics) it reaches a good accuracy on the Universal Dependencies corpus.

Test data is in another PR: https://github.com/snowballstem/snowball-data/pull/5

PR #160, opened 2021-12-18
Update of #44 used to package for alpine linux

### LINKED ISSUE #34: Snowball as a shared library

a16bitsysop:
* fixes #34


MehdiChinoune:
This couldn't be used on Windows!

a16bitsysop:
would the debian version work under windows?
https://salsa.debian.org/debian/snowball/-/blob/debian/latest/debian/patches/shared-library.diff

MehdiChinoune:
> would the debian version work under windows? https://salsa.debian.org/debian/snowball/-/blob/debian/latest/debian/patches/shared-library.diff

No, Windows has `.dll` extension for dynamic libraries, and also requires another file called import library with `.dll.a` to be generated.

PR #163, opened 2021-12-23
Add docs for non-thread-safe usage for python

DBCerigo:
@ojwb feel to use or extend or close as you see best.


ojwb:
Great, thanks.

> I suppose the same addition might also be required for https://github.com/snowballstem/pystemmer

Indeed - done.

I'll add something equivalent for the other languages too.

PR #165, opened 2022-02-01
Add special cases for the two (!) legal pluralizations of "bus"

pramsey:
Found by a PostgreSQL user trying to search for various transit words using full text search. Seems like a genuine corner case? I cannot thing of another word with the same pluralization pattern.

### DIFF
--- a/algorithms/english.sbl
+++ b/algorithms/english.sbl
@@ -176,6 +176,8 @@ define exception1 as (
         'dying'     (<-'die')
         'lying'     (<-'lie')
         'tying'     (<-'tie')
+        'busses'    (<-'bus')
+        'buses'     (<-'bus')
 
         /* special -LY cases */
 


ojwb:
@pramsey I just noticed this while looking for a different closed PR.  It seems github auto-closed it when you deleted your fork, and I can't even seem to reopen it now (not sure why github think this helpful behaviour!)

There is a cost to exceptions, especially ones that need to be checked for every word stemmed, so we don't generally worry about irregular cases if we're no worse off than we would be without any stemming.  However if there's unwanted conflation of the irregular forms with words which have different (or different enough) meanings then that's a different matter.  E.g. the existing cases in this group are:

```
        'skis'      (<-'ski')
        'skies'     (<-'sky')
```

and also an "invariant form" entry for `sky` lower down: without these exceptions, `skis` stems to `skis`, but `ski`, `sky` and `skies` all stem to `ski` so a search for `ski` matches `sky` and `skies` in documents, etc.

```
        'dying'     (<-'die')
        'lying'     (<-'lie')
        'tying'     (<-'tie')
```

These would otherwise stem to `dy`, `ly` and `ty` which would conflate with `dye` (to colour cloth), `lye` (caustic soda) and `tye` (which seems to have only obscure and archaic meanings but I guess was added as part of a pattern).  (These three special cases have since been generalised to words matching exactly the pattern `<non-vowel>ying`.)

Currently `buses` stems to `buse` and `busses` to `buss`, and the only potentially unwanted conflation seems to be with `buss` (archaic word for a kiss/to kiss).  If we're worrying about `buss` then `busses` is also the plural of the noun and third person singular of the verb, so it's inherently ambiguous.

If this is part of a wider pattern which we can come up with a sensible rule for then it might be worth an exception.  So far I've spotted these other words ending `-s` which add `-es` for the plural and can double the `s` or not:

* The plural of `bias` can be `biases` (stem `bias`) or `biasses` (stem `biass`, same as `biassed` and `biassing`)
* The plural of `gas` can be `gases` (stem `gase`) or `gasses` (stem `gass`, same as `gassed` and `gassing`)
* The plural of `yes` (as a noun) can be `yeses` (stem `yese`) or `yesses` (stem `yess`)

(We already have an exceptional invariant entry for `bias` to present us removing the `s`.)

PR #166, opened 2022-03-22
fix a memleak in snowball compiler

jsteemann:
When an output file is used and the name option is not, then the
compiler will dynamically allocate value for the `name` option and never
free it.
This is not a large problem, because at the end of the compile process
the OS will free all allocated memory anyway.
However, when using snowball as part of a larger toolchain and then
using compile options such as `-fsanitize=leak`, a memleak in snowball
can break the entire build.
This exactly what happened to us. We could work around this somehow, but
it seems better to fix the leak in the compiler properly.

The same issue was already reported previously by a former colleague of
mine in https://github.com/snowballstem/snowball/pull/136, but that PR
seemingly was disliked by the snowball maintainers. Trying it again with
this PR with a slightly modified solution and hope it can be merged now.

### DIFF
--- a/compiler/driver.c
+++ b/compiler/driver.c
@@ -146,8 +146,17 @@ static int read_options(struct options * o, int argc, char * argv[]) {
                 continue;
             }
             if (eq(s, "-n") || eq(s, "-name")) {
+                char * new_name;
+                size_t len;
+
                 check_lim(i, argc);
-                o->name = argv[i++];
+                /* Take a copy of the argument here, because
+                 * later we will free o->name */
+                len = strlen(argv[i]);
+                new_name = malloc(len + 1);
+                memcpy(new_name, argv[i++], len);
+                new_name[len] = '\0';
+                o->name = new_name;
                 continue;
             }
 #ifndef DISABLE_JS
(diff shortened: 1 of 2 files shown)


jsteemann:
2 out of 18 Travis CI sub-jobs failed during `git clone`. This is likely due to Github's outage yesterday.

jsteemann:
Is there a chance this PR can make it into a future version of snowball?

jsteemann:
Any chance this PR can make it into a future version of snowball?

ojwb:
I'll merge the patch, but I'd really strongly recommend using `LSAN_OPTIONS=leak_check_at_exit=0` when running snowball to avoid this sort of problem (as I suggested in the other ticket).

Doing that would address the issue for you once and for all; otherwise next time the code changes such that something is allocated and not released on exit you'll have to spend time submitting another patch and I'll have to spend time reviewing and applying it.

ojwb:
This patch introduce a bogus `-1 blocks unfreed` message at the end of every run, fixed in 4af181be519a0dbdf5d3657ac86ce03db5eb9011.

PR #167, opened 2022-03-28
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/snowball/commit/25df83387d7b449b530cdc1a38306cba71d9e714#commitcomment-69675396

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/snowball/commit/3ed0647bd9596b724133e8273188e38624fef328

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

### DIFF
--- a/algorithms/indonesian.sbl
+++ b/algorithms/indonesian.sbl
@@ -62,9 +62,9 @@ backwardmode (
     // prefix not in {ke, peng, per}
     define SUFFIX_KAN_OK as (
         // On page 29, the example "kompas Q.31" says "Both Nazief and Porter
-        // stemmer converted the word peledakan (blast, explotion) to ledak (to
-        // blast, to explode)".  However, the algorithm as described doesn't
-        // behave in this way - grammatically the prefix pe- occurs as a
+        // stemmer converted the word peledakan (blast, explotion [sic]) to
+        // ledak (to blast, to explode)".  However, the algorithm as described
+        // doesn't behave in this way - grammatically the prefix pe- occurs as a
         // variation of both the first-order derivational prefix peng- and the
         // second-order derivational prefix per-, but table 2.5 doesn't include
         // "pe", only table 2.6 does, so "peledakan" is handled (incorrectly)
--- a/algorithms/tamil.sbl
+++ b/algorithms/tamil.sbl
@@ -68,7 +68,7 @@ stringdef vs_au    '{U+0BCC}'
 /* Pulli */
 stringdef pulli    '{U+0BCD}'
 
-/* AU length markk */
+/* AU length mark */
 stringdef au_lmark '{U+0BD7}'
 
 
--- a/compiler/header.h
+++ b/compiler/header.h
@@ -238,7 +238,7 @@ enum name_types {
     t_string = 0, t_boolean = 1, t_integer = 2, t_routine = 3, t_external = 4,
     t_grouping = 5
 
-/*  If this list is extended, adjust wvn in generator.c  */
+/*  If this list is extended, adjust write_varname in generator.c  */
 };
 
 /*  In name_count[i] below, remember that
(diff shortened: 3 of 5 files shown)


NEWS
```diff
@@ -6,7 +6,7 @@ New Code Generators
 
 * Add Ada generator from Stephane Carrez (#135).
 
-Javascript
+JavaScript
```
jsoref, line 9: Brand

algorithms/indonesian.sbl
```diff
@@ -62,7 +62,7 @@ backwardmode (
     // prefix not in {ke, peng, per}
     define SUFFIX_KAN_OK as (
         // On page 29, the example "kompas Q.31" says "Both Nazief and Porter
-        // stemmer converted the word peledakan (blast, explotion) to ledak (to
+        // stemmer converted the word peledakan (blast, explosion) to ledak (to
```
jsoref, line 65: I'm pretty sure that this is meant to be an English word...

algorithms/tamil.sbl
```diff
@@ -68,7 +68,7 @@ stringdef vs_au    '{U+0BCC}'
 /* Pulli */
 stringdef pulli    '{U+0BCD}'
 
-/* AU length markk */
+/* AU length mark */
 stringdef au_lmark '{U+0BD7}'
```
jsoref, line 72: I'm not at all certain about this, but https://www.compart.com/en/unicode/U+0BD7 doesn't conflict with this change.

compiler/header.h
```diff
@@ -238,7 +238,7 @@ enum name_types {
     t_string = 0, t_boolean = 1, t_integer = 2, t_routine = 3, t_external = 4,
     t_grouping = 5
 
-/*  If this list is extended, adjust wvn in generator.c  */
+/*  If this list is extended, adjust write_varname in generator.c  */
```
jsoref, line 241: This was the only instance of `wvn`. I think this is the right item.

algorithms/indonesian.sbl
ojwb: Me too, but this is a verbatim quote from the paper describing the algorithm we're implementing and I think it's best overall to quote exactly what is written in the paper since we're quoting it here to note an inconsistency between two parts of that paper and how we've resolved that in our implementation.

We could add a "(sic)" or similar.

compiler/header.h
ojwb: Yes - it used to be called `wvn` in some generators but that got unified a while back.

NEWS
ojwb: Not sure I really agree we should be kowtowing to Oracle here.

algorithms/indonesian.sbl
jsoref: Fair. Sold.

NEWS
jsoref: Java and JavaScript aren't related beyond a branding coup selected by marketing folks at Netscape (which was later bought by AOL) who aligned LiveScript with the "up and coming" Java by Sun Microsystems (later bought by Oracle) making it JavaScript.

ojwb: Yes, I'm well aware of the history of this.

My dislike of the trademarked spelling here is (a) that it unhelpfully emphasises a connection to Java when really the two languages have very little in common (more than once I've had to try to resolve Java vs JS confusion by less technical management types) and (b) that it's now an Oracle-owned trademark which I don't really have any inclination to support.

jsoref: Frankly, you're inconsistent. The first item in here is `JavaScript`: https://github.com/snowballstem/snowball/blob/5b33b67cba5893571eabc060af16dd6440817d06/CONTRIBUTING.rst#key-problems-to-solve

If you wanted to make a political stand, you could use ECMAScript instead.

That said, this isn't meant to be a political PR, so I'll drop it.

Issue #191, opened 2024-01-14
snowballstemmer Python package does not declare supported Python versions


andy-maier:
The Python package for snowballstemmer defines the supported Python versions only as Trove classifiers (here: https://github.com/snowballstem/snowball/blob/master/python/setup.py#L48). That is not used by pip to find out whether the package can be installed on a particular Python version.

pip (and other installers) use the Python versions declared in the `python_requires` argument to `setup()`. If Python 2.x is still supposed to show up there, the syntax to match the currently specified Trove classifiers (except for Python 2.6) would be:
```
python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*'
```

I suggest to remove Python 2.6 at this point because it was EOL about 10 years ago, and people that still need Python 2.x for some reason all use 2.7 today. 2.7 is also EOL, but since it is the last 2.x version several projects still support it. See here for what the PSF supports: https://devguide.python.org/versions/

andy-maier:
I created PR #192 to address this issue.

ojwb:
Closing in favour of #192.

PR #192, opened 2024-01-14
Added python_requires for Python package; Removed Python 2.6

andy-maier:
For details, see the commit message.
This addresses issue #191.

I did not know where to update the change history, as the NEWS file seems to have only the changes for released versions, but not for those in development. I think something like this should be added to the NEWS file once released:

```
Python
-------

* Specified supported Python versions via `python_requires` so that they can be
  recognized by Python installers.

* Removed Python 2.6 from the versions documented in the Trove classifiers.
```

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -45,11 +45,12 @@ for lang in langs:
     if lang_titlecase not in ('Armenian', 'Yiddish'):
         classifiers.append('Natural Language :: ' + lang_titlecase)
 
+python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*'
+
 classifiers.extend([
     'Operating System :: OS Independent',
     'Programming Language :: Python',
     'Programming Language :: Python :: 2',
-    'Programming Language :: Python :: 2.6',
     'Programming Language :: Python :: 2.7',
     'Programming Language :: Python :: 3',
     'Programming Language :: Python :: 3.4',


ojwb:
> Looks like 3.3 is missing from the trove classifiers

I've now fixed that.

ojwb:
@mitya57 What are your thoughts?

You raised adding `python_requires` before in #158 with the rationale:

> this will cause pip to install older versions of snowball on older versions of Python.

As noted in the old ticket, I believe the only restriction is due to generating string literals with a `u` prefix e.g. `u"foo"` which Python 3 didn't support until Python 3.3.0, and all versions of Snowball with Python support actually support the same set of Python versions.  We seem to have concluded before not to bother but it does seem cleaner to have this explicitly specified.

It seems the accurate specification would be this (since actually any Python 2 should work):

```
python_requires='!=3.0.*, !=3.1.*, !=3.2.*'
```

As things are now if someone did try to install snowball using pip on e.g. Python 3.2 they would get the latest version installed but it will fail to work.  AIUI with this change they would instead get a version of snowball from before we made this change which would still fail to work.  This seems a bit of an odd outcome but realistically I don't think it's a case we need to worry greatly about at this point (3.2.x is approaching 8 years past EOL).

mitya57:
I don't think anyone cares about Python 2.6 in 2024, so it should be absolutely fine to merge this.

We can also go one step further and declare dependency on `>= 3.4`. People who still use Python 2.7 can probably stick to snowball version that they currently have installed.

Or we can go two steps further and do what I do in my projects: support only those Python versions that can be tested automatically, i.e. in GitHub Actions or another CI system.

ojwb:
As I said above already, I don't see a reason to deliberately prevent installation on older versions just because they're old.  Better to document what we understand the requirements to actually be rather than try to enforce something stricter.

Sure there's no hard guarantee that new versions of Snowball will generate code that works with Python 2 (similarly for Python 3 versions we can no longer easily test in CI) but Snowball's generated code is pretty straightforward Python that isn't likely to stop working with older versions without a deliberate decision to change the generation to make that happen - e.g. we could change to generate literal strings as `"foo"` instead of `u"foo"` and/or slightly simplify the code generated for integer division (which none of the shipped algorithms uses anyway), but neither change has any tangible benefit.

I'm going to infer from your reply that you're happy with the slightly odd "installs an older non-working version" side-effect, since all your suggested options have that too.

I also think the change as proposed isn't quite right as it just seems to set a variable whereas the documentation says this should be a named parameter in the `setup(`...`)` call.

I'll push a change to specify `python_requires='!=3.0.*, !=3.1.*, !=3.2.*'` as a named parameter.

Issue #187, opened 2023-12-07
Apostrophes


ojwb:
We seem to have a lack of consistency in how we expect apostrophes to be handled by code using Snowball stemmers, which means currently tokenisation before stemming needs to encode knowledge of the stemmer to be used.  This is explicitly noted in the docs, but that doesn't make it any less unhelpful:

> What is a word? For indexing purposes, a word in a European language is a sequence of letters bounded by non-letters. But in English, an internal apostrophe does not split a word, although it is not classed as a letter. The treatment of these word boundary characters affects the stemmer. For example, the Kraaij Pohlmann stemmer for Dutch (Kraaij, 1994, 1995) removes hyphen and treats apostrophe as part of the alphabet (so `'s`, `'tje` and `'je` are three of their endings). The Dutch stemmer presented here assumes hyphen and apostrophe have already been removed from the word to be stemmed.

* Rather contradicting the text quoted above, the English stemmer expects apostrophes to be treated as a letter: "the English stemmer treats apostrophe as a letter" (https://snowballstem.org/texts/apostrophe.html)

* As above, the Kraaij Pohlmann stemmer also expects apostrophes to be treated as a word character

* But the "Dutch" stemmer doesn't

* The French stemmer doesn't either and so doesn't expect `l'` and `d'` prefixes to be present on input

If it's feasible, I think it'd be more helpful for all the stemmers to handle apostrophe being treated as a word character.  If there's a reason why we can't, then we should provide some sort of metadata (e.g. an "apostrophe_is_word_character" flag that can be queried on each stemmer) so that code using the stemmers can automatically configure their tokenisation stage.

ojwb:
A full list of French elisions seems to be: `l'`, `j'`, `c'`, `m'`, `t'`, `s'`, `n'`, `d'`, `qu'`

ojwb:
I have adjusted the French stemmer to remove elisions.

Still to do:

[...]

PR #194, opened 2024-02-21
create persian.sbl

saeiddrv:


saeiddrv:
add Persian stemming algorithm!

python/modules.txt
```diff
@@ -0,0 +1,62 @@
...
+greek           UTF_8                   greek,el,gre,ell
+hindi           UTF_8                   hindi,hi,hin
+hungarian       UTF_8,ISO_8859_2        hungarian,hu,hun
+indonesian      UTF_8,ISO_8859_1        indonesian,id,ind
+irish           UTF_8,ISO_8859_1        irish,ga,gle
+italian         UTF_8,ISO_8859_1        italian,it,ita
+lithuanian      UTF_8                   lithuanian,lt,lit
+nepali          UTF_8                   nepali,ne,nep
+norwegian       UTF_8,ISO_8859_1        norwegian,no,nor
+persian         UTF_8                   persian,fa,fas,pers
```
ojwb, line 32: This line looks right, but you should add it to the existing `libstemmer/modules.txt` rather than creating a new version of the file in a different subdirectory.

algorithms/persian.sbl
```diff
@@ -0,0 +1,117 @@
...
+externals ( stem )
+
+
+groupings ()
+
+
+define Normalize as (
+    do repeat (
+        [substring] among (
+            '{U+06A9}' '{U+0643}' ( <- '{kaf}' ) // Normalize Kaf
```
ojwb, line 59: From above:

```
stringdef kaf     '{U+06A9}'
```

So as part of this you seem to be replacing kaf with itself, which seems pointless.

saeiddrv:
I have fixed the issues.

algorithms/persian.sbl
```diff
@@ -115,34 +112,34 @@ define Suffix_Adjective as (
...
+        '{be}{re}{khe}{ye}' ()
+        '{he}{mim}{ye}{nun}' ()
+        '{he}{mim}{alef}{nun}' ()
+        '{he}{mim}{he}' ()
+        '{mim}{te}{re}' ()
+        '{te}{ye}{mim}' ()
+        '{sin}{lam}{alef}{mim}' ()
+        '{alef}{sin}{lam}{alef}{mim}' ()
+        '{ye}{ain}{nun}{ye}' ()
+        '{aa}{lam}{be}{vav}{mim}' ()
```
ojwb, line 158: There's no need for all these explicit empty actions - no action is the default.

algorithms/persian.sbl
```diff
@@ -55,11 +54,11 @@ externals ( stem )
...
     do repeat (
         [substring] among (
-            '{U+0643}' ( <- '{kaf}' ) // Normalize Kaf
+            '{U+0643}' ( <- '{kaf}' )             // Normalize Kaf
             '{U+0626}' '{U+064A}' ( <- '{ye}'  )  // Normalize Ye
-            '{U+0629}' ( <- '{he}'  )  // Normalize Heh
+            '{U+0629}' ( <- '{he}'  )             // Normalize Heh
             '{U+0623}' '{U+0625}' ( <- '{alef}' ) // Normalize Alef
-            '{U+0624}' ( <- '{vav}' )  // Normalize Waw and Waw with Hamza above
+            '{U+0624}' ( <- '{vav}' )             // Normalize Waw and Waw with Hamza above
```
ojwb, line 67: As I noted before, it'd be better to use stringdef names for everything.

Then it's clearer to the reader what the code is doing.  Consider the line:

```
            '{U+0623}' '{U+0625}' ( <- '{alef}' ) // Normalize Alef
```

Currently I need to e.g. look up U+0623 to find it's "ARABIC LETTER ALEF WITH HAMZA ABOVE", then look up U+0625 to find it's "ARABIC LETTER ALEF WITH HAMZA BELOW" before I can fully understand this single line.  If it instead read something like this then it would be clear without additional research:

```
            '{alef_with_hamza_above}' '{alef_with_hamza_below}' ( <- '{alef}' ) // Normalize Alef
```

algorithms/persian.sbl
```diff
@@ -115,34 +112,34 @@ define Suffix_Adjective as (
...
+        '{sin}{lam}{alef}{mim}' ()
+        '{alef}{sin}{lam}{alef}{mim}' ()
+        '{ye}{ain}{nun}{ye}' ()
+        '{aa}{lam}{be}{vav}{mim}' ()
     )
 )
 
 define stem as (
-    Exception
+    ( Exception )
```
ojwb, line 163: This means the same with or without the parentheses (just noting as the commit that changed this was "fix exception section"...)

libstemmer/modules.txt
```diff
@@ -29,6 +29,7 @@ italian         UTF_8,ISO_8859_1        italian,it,ita
 lithuanian      UTF_8                   lithuanian,lt,lit
 nepali          UTF_8                   nepali,ne,nep
 norwegian       UTF_8,ISO_8859_1        norwegian,no,nor
+persian         UTF_8                   persian,fa,fas,pers
```
ojwb, line 32: I think that should be `per` not `pers` at the end (it should be a three letter code and https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes lists `per` as the ISO 639-2/B code for Persian).

ojwb:
> We also need test data and a description for a new algorithm - there's a detailed description of the requirements in `CONTRIBUTING.rst`.

@saeiddrv You've since opened https://github.com/snowballstem/snowball-data/pull/25 to add test data - thanks for doing that.

We really do still need a description of the new algorithm though, which should be a pull request against the https://github.com/snowballstem/snowball-website repo.

We've found that algorithms without good documentation are much harder to maintain - for example, if somebody reports that a particular stemmer doesn't remove a prefix or suffix it's a usually lot of work to try to determine if that's an oversight or something that was considered but rejected for a good reason (most languages seem to have a prefix or suffix that can't be safely removed without causes worse problems in words that start or end with the same letters but where they aren't the intended prefix/suffix), but if we have a document that says "suffixes A, B and C are not removed because of X and Y" then we can just point the bug reporter to that.

If the implementation is from an algorithm that's described in a published paper we may be able to link to that to provide most of the information, provided the paper does contain a sufficiently detailed description of the algorithm and discussion of the design decisions that lead to it.

I can help with proof-reading and the like, but I didn't develop the algorithm and I don't know the Persian language so it isn't a document I could write the actual content of.

https://snowballstem.org/algorithms/indonesian/stemmer.html is a reasonable example showing the sort of thing that's useful.  There are macros to help render the list of references, the table of example stems, and the snowball code - see `algorithms/indonesian/stemmer.tt` in the repo.

algorithms/persian.sbl
```diff
@@ -0,0 +1,181 @@
...
+    do repeat (
+        [substring] among (
+            '{ar_kaf}' ( <- '{kaf}' )
+            '{ar_ye_with_hamza_above}' '{ar_ye}' ( <- '{ye}'  )
+            '{ar_he_marbuta}' ( <- '{he}'  )
+            '{ar_alef_with_hamza_above}' '{ar_alef_with_hamza_below}' ( <- '{alef}' )
+            '{ar_vav_with_hamza_above}' ( <- '{vav}' )
+
+
+            '{gaf}{alef}{nun}' ( <- '{he}'  )     // gaan -> he
```
ojwb, line 79: I suspect you want to add `'' (next)` here, assuming it's meant to replace any instance of one of these characters in the word.

As currently written it will stop transliterating characters when it gets to a character which there isn't a transliteration for.

saeiddrv:
@ojwb Thanks for your review, guidance, and patience! I will fix those issues and work on making it more accurate in my next pushes.

algorithms/persian.sbl
```diff
@@ -65,18 +65,17 @@ externals ( stem )
...
-            '{ar_vav_with_hamza_above}' ( <- '{vav}' )
+            '{ar_kaf}' ( <- '{kaf}' next )
+            '{ar_ye_with_hamza_above}' '{ar_ye}' ( <- '{ye}' next )
+            '{ar_he_marbuta}' ( <- '{he}' next )
+            '{ar_alef_with_hamza_above}' '{ar_alef_with_hamza_below}' ( <- '{alef}' next )
+            '{ar_vav_with_hamza_above}' ( <- '{vav}' next )
 
 
-            '{gaf}{alef}{nun}' ( <- '{he}'  )     // gaan -> he
+            '{gaf}{alef}{nun}' ( <- '{he}' next )     // gaan -> he
```
ojwb: This doesn't seem to address the point I raised.

On the assumption that this routine is meant to loop over the whole string making these changes to any occurrences it finds then, I suggested adding a new `among` case:

```
            '' (next)
```

The longest matching string in the `among` is chosen and an empty string is always a match, so this acts kind of like a `default` case in a C switch.  Otherwise `repeat` will only loop until the `among` doesn't match, so it will only change things in the initial span of the word which is made up of strings in the `among`.

Instead you've added `next` after every action - the `among` matching already advances the cursor over the matched string (and `<- 'x'`/`delete` in the action will adjust the cursor to still be on the character after the matched string), so adding `next` like you have means that this routine won't perform two adjacent replacements.  If that's not possible for any Persian words then that's OK (and doing `next` after a replacement is effectively an optimisation), but you still need to advance in the case where the `among` does not match.

ojwb:
Hmm, it seems you've made updates to fix my comments, but github still shows my comment last so I'd missed this until I noticed the "Outdated" marker on my comment.

I'm currently focusing on getting a new release out (it's very nearly 3.5 years since the last one which is too long and there's a huge pile of unreleased useful changes), and my plan is to then dealing with the pending new algorithms (including this one) and making another release with those.

algorithms/persian.sbl
```diff
@@ -0,0 +1,287 @@
...
+stringdef ar_num3 '{U+0663}'
+stringdef ar_num4 '{U+0664}'
+stringdef ar_num5 '{U+0665}'
+stringdef ar_num6 '{U+0666}'
+stringdef ar_num7 '{U+0667}'
+stringdef ar_num8 '{U+0668}'
+stringdef ar_num9 '{U+0669}'
+
+// Other Characters
+stringdef half_space  '{U+200C}'
```
ojwb, line 86: This seems wrong as `UnicodeData.txt` says `U+200C` is:

```
200C;ZERO WIDTH NON-JOINER;Cf;0;BN;;;;;N;;;;;
```

Not sure if the `half_space` name is wrong or the hex code is wrong though.

Generally I'd expect the tokenisation stage which calls the stemmer to split on any space characters, so we shouldn't see anything that's `Zs` in `UnicodeData.txt`.  I don't see anything named `HALF SPACE` but e.g. I wouldn't expect the stemmer to see:

```
2009;THIN SPACE;Zs;0;WS;<compat> 0020;;;;N;;;;;
```

So if the hex code is wrong and should have been `U+2009` or similar, I'd say this is not something we need to check for.

dscorbett: U+200C ZWNJ is called “<span lang="fa">نیم فاصله</span>” (‘half space’) in Persian. It’s a word-internal morpheme separator.

ojwb: I see (that seems a misleading name as it's zero width not half width and isn't a space).

I can't see much evidence of U+200C also being known as _half space_ in English (there are some web search results but they all seem to be referring to the literal English translation of the Persian name).  So if we're going to refer to it as _half space_ here, I think we need to explain why in a comment.

algorithms/persian.sbl
```diff
@@ -0,0 +1,291 @@
...
+
+define Remove_Suffix_Adjectives as (
+    backwards (
+        [substring] among (
+            '{te}{re}' (delete)           // tar (comparative)
+            '{te}{re}{ye}{nun}' (delete)  // tarin (superlative)
+            '{mim}{nun}{dal}' (delete)    // mand
+            '{nun}{alef}{kaf}' (delete)   // naak
+            '{gaf}{ye}' (delete)          // gii
+            '{alef}{nun}{heh}' (delete)   // aaneh
```
ojwb, line 234: If an input word is exactly one of these suffixes, then the output stem is an empty string which seems better avoided even if none of these are actually valid words.

Checking `output.txt` from the snowball-data PR, there appear to be 155 entries in `voc.txt` which give an empty stem, at least some of which seem to be real words - e.g. _بهانه‌هایی_ which google translate says means "excuses".  We definitely don't want to stem real words to an empty stem.

If we should require leaving at least one character before the removed suffix, that can be easily (and efficiently) achieved by adding a check that the cursor isn't "`atlimit`" after the `substring` above like so:

```
[substring] not atlimit among (
```

If we should require leaving multiple characters then `$(len > X)` checks as used elsewhere are OK, if a bit inelegant (and less efficient for UTF-8 where the string length needs computing each time).

ojwb: https://en.wikipedia.org/wiki/Persian_alphabet says there are "32 letters [in] the modern Persian alphabet" - that means there are only 32 possible different single letter stems, so producing a single letter stem for a longer input (particular for real words) is probably also a design error because it's very likely to result in unwanted conflation.

algorithms/persian.sbl
```diff
@@ -0,0 +1,411 @@
...
+    )
+  )
+
+  // --- Noun plurals.
+  //     -YAN and -GAN are productive; -AN is guarded by 'protect_lex_an' probe.
+  define Noun_Plural as (
+    [substring] among (
+      '{ye}{alef}{nun}'     ($(len > 5) delete set changed)                 // -YAN
+      '{gaf}{alef}{nun}'    ($(len > 5) delete set changed)                 // -GAN
+      '{heh}{alef}{ye}{ye}' ($(len > 6) delete set changed)                 // -HAYI
```
ojwb: It looks to me like this rule can never match since any word which could match here ends `{ye}{ye}` and has `$(len > 4)` so will have matched the rule in `Noun_Other` on line 269 below.

If I comment out line 250 the output of the stemmer isn't affected.

algorithms/persian.sbl
```diff
@@ -0,0 +1,392 @@
...
+    setlimit tomark p1 for (
+      [substring] among (
+        // Noun possessive suffixes (long + short clitics)
+        '{alef}{mim}'       (delete) // -AM
+        '{alef}{te}'        (delete) // -AT (can also be a noun plural)
+        '{alef}{shin}'      (delete) // -ASH
+// FIXME: These were dominated by -AN below before refactoring.
+//        '{mim}{alef}{nun}'  (delete) // -MAN
+//        '{te}{alef}{nun}'   (delete) // -TAN
+//        '{shin}{alef}{nun}' (delete) // -SHAN
```
ojwb: Before I refactored we had:

```
    N_Hash or Noun_Other or Noun_Plural or Noun_Possessive
```

`Noun_Plural` had a rule for `'{alef}{nun}'` which would also match any word with one of the 3 commented out endings above.   (And note that the original condition on `'{alef}{nun}'` was `$(len > 4)` and the original conditions on the rules above were all `$(len > 5)`.)

I suspect we want to uncomment these, but that affects the stemming of quite a lot of words so needs reviewing.

algorithms/persian.sbl
```diff
@@ -0,0 +1,385 @@
...
+      '{alef}{nun}{dal}' ($(len > 5) delete)
+      '{alef}{sin}{te}'  ($(len > 5) delete)
+      '{alef}{sin}'      ($(len > 4) delete)
+      '{alef}{ye}'       ($(len > 4) delete)
+      '{ye}{dal}'        ($(len > 4) delete)
+      '{ye}{mim}'        ($(len > 4) delete)
+    )
+  )
+
+  // --- Verb: person/number endings
```
ojwb: I can see "person" endings handled below, but where are the "number" endings?

ojwb:
The Ada failure looks like a bug in Snowball (e.g. our Ada runtime code and/or the Ada code we generate).  So nothing wrong with the code in this PR, but unfortunately a blocker for merging.  It's a pre-condition failing, so potentially we could relax or drop that pre-condition, but the pre-condition seems valid so really should be allowed to remain - I think there's a deeper problem.

I've managed to reduce to this testcase (in case you're curious, and so I can't lose this reproducer):

```
stringescapes { }

externals ( stem )

define stem as (
  setlimit tomark size - 4 for [
  backwards (
    ['AB'] (delete)
  )
)
```

with input:

```
ABCDAEAB
AHFIAJ
ABI
```

This reduced testcase looks like valid Snowball to me, but also seems to fail for C (unlike the real `persian.sbl`).

It seems to actually fail on the second entry but the other two need to be there (odd that the entry after matters).  The need for multiple inputs to trigger suggests something isn't getting reset/set when we start a new word, but if so I've not spotted where so far.

I'll debug further and sort this out.

ojwb:
> I'll debug further and sort this out.

Turns out the code generated to handle `setlimit tomark X for Y` in a forwards context was buggy for all languages (the limit was incorrectly restored after doing Y).  I've fixed that, but the `make check_ada_persian` still fails so it seems this bug wasn't the cause (or not the full cause).  Good to have found and fixed it though.

(`setlimit tomark X for Y` is used fairly extensively in existing stemmers, but always in a backwards context.)

I'll keep investigating.

ojwb:
Those Ada codegen bugs are now fixed.

I've pushed a rebase of the branch onto the latest master (with a couple of additional clean-up commits) to my fork to test these fixes actually fix the failure for Persian

https://github.com/ojwb/snowball/commits/add-persian-stemmer/

and CI now passes for all target languages:

https://github.com/ojwb/snowball/actions/runs/24297535548

saeiddrv:
Thank you so much for all your help and the great effort you've put into this PR — I really appreciate it!

algorithms/persian.sbl
```diff
@@ -0,0 +1,393 @@
...
+          ( atlimit )
+    )
+  )
+
+  // --- Noun irregular rewrites (small hash table).
+  //     Only include entries that actually change the buffer.
+  define N_Hash as (
+    [substring] among (
+      '{alef}{khe}{be}{alef}{re}'      (<- '{khe}{be}{re}')
+      '{khe}{vav}{alef}{sin}'          (<- '{khe}{alef}{sin}{heh}')
```
ojwb: I tested with a (currently terribly hacky) coverage feature I've been working on and it seems that the `'{khe}{vav}{alef}{sin}'` case here is not exercised by anything in the testdata.  A simple demonstration is that the tests still pass with this change:
```
--- a/algorithms/persian.sbl
+++ b/algorithms/persian.sbl
@@ -244,7 +244,7 @@ backwardmode (
   define Irregular_Noun as (
     [substring] among (
       '{alef}{khe}{be}{alef}{re}'      (<- '{khe}{be}{re}')
-      '{khe}{vav}{alef}{sin}'          (<- '{khe}{alef}{sin}{heh}')
+      '{khe}{vav}{alef}{sin}'          (<- 'THISISNEVERUSED!')
       '{alef}{sin}{alef}{te}{ye}{dal}' (<- '{alef}{sin}{te}{alef}{dal}')
     )
   )
```

`Stem_Verb` below deletes suffix `{alef}{sin}` but that's tested *after* this so I don't think that is causing us not to match here.  Is there a typo, or is this just an uncommon case?

If it's correct it'd be good to add at least one real word to `voc.txt` which exercises it.

ojwb:
I think we're mostly there, but one other thing:

Currently the protection against overstemming short words just counts characters, but most of the stemmers instead use a "measure" which approximates counting syllables (see https://snowballstem.org/texts/r1r2.html for some details, and some English examples) - this generally found seems to work better than a simple character count.

This is usually done by counting alternating spans of vowels and consonants, but it seems written Persian typically omits short vowels which probably makes this more complex as the counting may need to discriminate a consonant cluster from consonants with an omitted short vowel in between.  (Also based on my limited reading it seems some letters can be a vowel or consonant depending on context - perhaps somewhat like English _y_).

The length checks in the stemmer code are now expressed in terms of `p1` so if a suitable syllable count is possible we can just set `p1` based on it and pretty much everything will then use it.

saeiddrv:
Thanks for the detailed suggestion — I looked into implementing a syllable-based R1 measure using the visible long vowel letters (آ، ا، و، ی) with a standard `gopast vowel / gopast non-vowel` pattern and falling back to `hop 3` when no visible vowel transition is found.

After testing it against common Persian words, I found the improvement is marginal at best. The core issue is that written Persian routinely omits short vowels (a, e, o), so a large portion of common stems contain no visible vowel letters at all — words like دست (dast, "hand"), مرد (mard, "man"), سخت (sakht, "hard"), بچه (bachche, "child"). For all of these, the visible-vowel R1 finds no transition and falls back to the character count anyway.

On top of that, و and ی are ambiguous — they can be consonants (/v/ and /y/) depending on position. Treating them unconditionally as vowels would set R1 too early for words like ورزش (varzesh, "sport") or یکی (yeki, "one"), giving *less* protection than the current `hop 3`.

So the syllable-counting approach that works well for European languages doesn't really translate well to Persian orthography due to the hidden short vowels. I think `hop 3` is a reasonable default for now.

algorithms/persian.sbl
saeiddrv: You're right — no word in the vocabulary ends with خواس, and no pipeline path produces this ending through suffix stripping either. Removed in 1213698.

algorithms/persian.sbl
```diff
@@ -0,0 +1,392 @@
...
+//  - Delete ZWJ (U+200D) and ASCII spaces so the token is solid.
+//  - ZWNJ (U+200C) is preserved here; used as prefix-stem separator in Phase 2,
+//    then deleted by Delete_ZWNJ after prefix stripping is complete.
+// ============================================================================
+define Normalize_Characters as (
+  repeat (
+    [substring] among (
+      '{ar_kaf}'                                                 (<- '{kaf}')
+      '{ar_ye}' '{ar_ye_with_hamza_above}'                       (<- '{ye}')
+      '{ar_heh}' '{ar_he_marbuta}'                               (<- '{heh}')
```
ojwb: This is mostly cosmetic, but having the code and description match is very helpful for people to understand the algorithm, and especially for future maintenance.

This rule in the description is:

> Arabic te marbuta (U+0629) / heh with hamza (U+06C1) → Persian heh (U+0647),

One difference is that the order is switched and it would be better if they matched - is one way round more natural?

But also the naming of `U+0629` doesn't match (`ar_he_marbuta` vs "Arabic te marbuta").  ``UnicodeData.txt` lists `U+0629` as:

```
0629;ARABIC LETTER TEH MARBUTA;Lo;0;AL;;;;;N;ARABIC LETTER TAA MARBUTAH;;;;
```

so maybe `ar_he_marbuta` is a typo?

Even if both names are valid, it would be more helpful to use matching names in code and description.

algorithms/persian.sbl
```diff
@@ -0,0 +1,392 @@
...
+        '{alef}{te}'        (delete) // -AT (Arabic sound plural; can also be possessive "your")
+        '{alef}{nun}'       (delete) // -AN (only if not lexical)
+        '{ye}{nun}'         (delete) // -IN (also adjective derivation ending)
+
+        // Other derivational noun endings (conservative set).
+        '{gaf}{alef}{heh}'  (delete) // -GAH
+        '{be}{alef}{nun}'   (delete) // -BAN
+        '{gaf}{ye}'         (delete) // -GI (abstract noun)
+        '{ye}{te}'          (delete) // -YAT
+        '{ye}{ye}'          (delete) // -YY (double Y; also adj. relative -Y)
```
ojwb, line 295: The description uses _-yi_ instead of `-YY` - we should pick a consistent name to use.

ojwb:
> Thanks for the detailed suggestion — I looked into implementing a syllable-based R1 measure using the visible long vowel letters (آ، ا، و، ی) with a standard `gopast vowel / gopast non-vowel` pattern and falling back to `hop 3` when no visible vowel transition is found. [...]

Thanks for looking into this.  I've summarised the relevant parts of your comment into a new section in the snowball-website PR.

> On top of that, و and ی are ambiguous — they can be consonants (/v/ and /y/) depending on position. Treating them unconditionally as vowels would set R1 too early for words like ورزش (varzesh, "sport") or یکی (yeki, "one"), giving _less_ protection than the current `hop 3`.

FWIW, the letter _y_ in English can also be a consonant of vowel - Snowball's rule there is that it's not a vowel at the start of a word or after a vowel (so e.g. it's not a vowel in _yoke_ or _buy_, but is in _bye_, _cry_ and _tyre_).  Treating as a non-vowel at the start of the word seems to work for those two examples, but perhaps it's not as easy as _y_ in English, and this doesn't help address omitted short vowels which seems a hard problem.

algorithms/persian.sbl
saeiddrv: You're right, renamed `ar_he_marbuta` to `ar_teh_marbuta` to match the Unicode name and reordered to match the description in 52defb3.

algorithms/persian.sbl
saeiddrv: Good catch. Updated the website description to use `-yy` to match the code in 9c48c56.

ojwb:
@saeiddrv Thanks for all your work on this.

One other question - I'm in the process of trying to sort out a more consistent approach to handling apostrophes and hyphens across the stemmers (#187).  Does Persian use apostrophes or hyphens?

saeiddrv:
@ojwb  Thank you so much for your guidance and support throughout this, it truly wouldn't have been possible without your help.

No, Persian doesn't use apostrophes or hyphens, so it shouldn't be affected by #187.

PR #197, opened 2024-04-28
Add Pāli Stemmer

khemarato:
This is currently a draft PR for starting a discussion.

The test cases are defined in https://github.com/snowballstem/snowball-data/pull/26

The current, super naive, implementation of just removing common suffixes achieves an admirable accuracy of 97.7% on the test set.  See [this gist](https://gist.github.com/khemarato/e7ce796e04a26c3ad93aa5522ec117bc) for the failing cases.

Any feedback at all is appreciated.

### DIFF
--- a/algorithms/pali.sbl
+++ b/algorithms/pali.sbl
@@ -1,0 +1,20 @@
+externals ( stem )
+
+stringescapes {}
+
+stringdef a-   '{U+0101}'  // ā
+stringdef i-   '{U+012B}'  // ī
+stringdef u-   '{U+016B}'  // ū
+stringdef m1   '{U+1E41}'  // ṁ
+stringdef m2   '{U+1E43}'  // ṃ
+
+define stem as (
+  backwards (
+    [substring] delete among (
+        'm{a-}na' 'anta' 'onta' 'unta' 'enta' 'issa'
+        'ati' '{a-}ti' 'ant' 'esi' '{a-}si' 'eti'
+        'a{m1}' 'a{m2}' 'u{m1}' 'u{m2}' 'i{m1}' 'i{m2}'
+        'i' '{i-}' 'a' '{a-}' 'o' 'u' '{u-}'
+    )
+  )
+)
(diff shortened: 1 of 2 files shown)


algorithms/pali.sbl
```diff
@@ -0,0 +1,20 @@
...
+
+stringdef a-   '{U+0101}'  // ā
+stringdef i-   '{U+012B}'  // ī
+stringdef u-   '{U+016B}'  // ū
+stringdef m1   '{U+1E41}'  // ṁ
+stringdef m2   '{U+1E43}'  // ṃ
+
+define stem as (
+  backwards (
+    [substring] delete among (
```
ojwb, line 13: While this code is correct, I'd suggest removing the `delete` here and instead adding `(delete)` after line 17 below.  Then it's clearer to a reader less familiar with the Snowball language that the deletion only happens when one of these suffixes matches.  Framing it as an explicit action also more easily allows for different actions for different suffixes if that is useful.

algorithms/pali.sbl
```diff
@@ -0,0 +1,20 @@
...
+        'm{a-}na' 'anta' 'onta' 'unta' 'enta' 'issa'
+        'ati' '{a-}ti' 'ant' 'esi' '{a-}si' 'eti'
+        'a{m1}' 'a{m2}' 'u{m1}' 'u{m2}' 'i{m1}' 'i{m2}'
+        'i' '{i-}' 'a' '{a-}' 'o' 'u' '{u-}'
+    )
```
ojwb, line 18: This will produce an empty stem if the input is exactly one of these suffixes, for example:

```
$ echo anta|./stemwords -l pali

$
```

Both `anta` and `esi` are in your test vocabulary and suffer from this, but it would be undesirable even if it only affected strings which weren't real words in the language as such strings can occur as proper nouns, words from other languages, typos, OCR errors, etc.

It can also produce very short stems (even a single character) - there are 86 cases of single character stems in your test vocabulary.  I don't really know anything about Pali so maybe some or even all of these are actually appropriate, but typically it is a sign of overstemming since there are a very limited number of single character stems.

Most of the existing stemmers define R1 and R2 regions based on vowels and non-vowels, and limit suffix removal to one of these regions which has proven to be a good approach in general - see https://snowballstem.org/texts/r1r2.html for more details.

ojwb:
I'm going to close this - I'm certainly open to adding a Pāli algorithm, but the current state here needs much more work and progress seems to have stalled.

Also Pāli is rather unusual case where the language is essentially fixed and maybe algorithmic stemming isn't the best approach.  There's more discussion in https://github.com/snowballstem/snowball-data/pull/26, especially https://github.com/snowballstem/snowball-data/pull/26#discussion_r1586889702 lays down pros and cons for a snowball Pāli algorithm.

PR #206, opened 2025-02-11
api: fixed potential NULL dereference

ProjectMutilation:
If we transfer control to the SN_create_env() function by using the error label when there is a memory allocation error of z->p or z->S, we can then dereference the NULL pointer z->S in the function SN_close_env(). Added the pointer check for avoiding potential problem.

Found by Postgres Pro with Svace static analyzer
Fixes: e9dd099 ("This module will contain only the code and build system, and documentation for building and running the stemming library. All sample data will be in a separate module, and the website will be in its own module too.")

### DIFF
--- a/runtime/api.c
+++ b/runtime/api.c
@@ -36,7 +36,7 @@ error:
 extern void SN_close_env(struct SN_env * z, int S_size)
 {
     if (z == NULL) return;
-    if (S_size)
+    if (z->S && S_size)
     {
         int i;
         for (i = 0; i < S_size; i++)


ojwb:
Thanks for spotting this.

I think your analysis of the history is slightly faulty though - looks like this issue was actually introduced by 5dd0851c5469ccab91b5d8a526101e68142c5d29 so Snowball 2.0.0.

e9dd099a0ab5a63ba7d23ccb0b41bff0c0805fb6 is the first commit in the repo and where the check you suggest changing was originally first committed to version control, but the check was OK until 5dd0851c5469ccab91b5d8a526101e68142c5d29 because `z->S_size` was only assigned once `z->S` and all its elements were successfully allocated.  Before that change it would be zero from `calloc()` at this point so we wouldn't try to free `z->S` (though this did mean we could leak memory if we failed to allocate before that change).

Just taking a deeper look...

ojwb:
I pushed a slightly simpler fix - we only need to check `if (z->S)` since if `S_size` is zero we'll never even try to allocate `z->S`.  Therefore `z->S` being non-NULL implies `S_size` is non-zero.  Also with `if (z->S && S_size)` it initially looks like there might be a memory leak when `S_size` is zero - one needs to review the allocation code to see that can't happen.

PR #211, opened 2025-03-04
python: Set minimum required Python version to 3.7

mitya57:
Python 2.7 reached its end of life on 2020-01-01.

Python 3.6 reached its end of life [two years later](https://devguide.python.org/versions/), on 2021-12-23, but that is still more than 3 years ago.

And bumping Python to 3.7 will allow me to upgrade setuptools to 61.2.0, which is the first version to properly support metadata in [`pyproject.toml`](https://packaging.python.org/en/latest/specifications/pyproject-toml/).

Also, replace Python 3.8 with 3.7 in the tests, so that we test both the lower and higher bounds of supported version range.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,8 +60,8 @@ jobs:
         - name: Pascal
           FPC: fpc
           apt_packages: 'fpc'
-        - name: Python 3.8
-          PYTHON_VERSION: 3.8
+        - name: Python 3.7
+          PYTHON_VERSION: 3.7
           os: 'ubuntu-22.04'
           # The pure Python versions run slowly - when we used travis for CI
           # we used to need to thin the testdata for languages such as Arabic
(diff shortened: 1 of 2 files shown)


mitya57:
The CI failure seems unrelated:

```
"C:\Program Files\Git\bin\sh.exe" -xc "curl -L 'https://github.com/mozilla/sccache/releases/download/v0.7.6/sccache-v0.7.6-x86_64-pc-windows-msvc.tar.gz' | tar xzf - -O --wildcards '*/sccache.exe' > 'C:\Users\runneradmin\.cargo\bin\sccache.exe'"
...
gzip: stdin: not in gzip format
```

Can you retry it please?

ojwb:
What are the benefits of being able to assume a newer setuptools version?

This seems a significantly more aggressive change than dropping Python2 support.  We'd already said we planned to stop actively supporting Python 2 in 2021 with an opportunity for people to respond if they thought that was unreasonable; we've not done anything like that for the Python 3 versions supported so it's hard to know what people might still be using.  Upstream EOL for Python 3.6 was essentially 2 years later than for 2.7, and upstream EOL dates are often much more aggressive than the versions available on LTS distros (not specific to Python even).

I'm currently trying to get another Snowball release out as it's been much longer than is ideal - I think we should postpone merging this until after that release at least.

PR #217, opened 2025-03-11
Make sure all routines get mangled names in Python

dscorbett:
The Python generator uses private mangled names for Snowball routines by prepending `"__"`. Names are only mangled if they being with `"__"` but don’t end with `"__"`. This makes a stemmer crash if an `among` string has a routine whose name ends in `"__"`.

For example, if you put the following in algorithms/esperanto.sbl:
```snowball
routines ( x__ )
externals ( stem )
define x__ as true
define stem as ( [substring] among('x' x__) delete )
```
then run this:
```zsh
make check_python_stemwords
echo xy | python python_check/stemwords.py -c UTF_8 -l esperanto -i /dev/stdin -o /dev/stdout
```
the output ends:
```
AttributeError: 'EsperantoStemmer' object has no attribute '_EsperantoStemmer__r_x__'
```

This PR fixes that bug by ensuring that a Snowball routine’s Python method is always mangled.

### DIFF
--- a/compiler/generator_python.c
+++ b/compiler/generator_python.c
@@ -37,7 +37,9 @@ static void write_varname(struct generator * g, struct name * p) {
             break;
         case t_routine:
             write_string(g, "__");
-            /* FALLTHRU */
+            write_s(g, p->s);
+            write_string(g, "_r");
+            return;
         default: {
             int ch = "SBIrxg"[p->type];
             write_char(g, ch);


ojwb:
OK, and moving the `_r` marker to the end can't actually result in a naming clash because we prefix all routines with `__` and prefix everything else (except `externals`) with `<letter>_`.

However using `__` in a Snowball identifier will result in the generated C code being ill-formed C++, so maybe we should just outlaw Snowball identifiers containing a double underscore.

Another option would be to generate new names for problematic (or perhaps all) non-`externals` which would solve any such naming issues for except for `externals`, and we already document that `externals` need to be named with consideration of naming restrictions in the target language.  Generating short names might actually be faster for the Python function-in-among case since we call the method via a string name which presumably ends up getting resolved on each call, though maybe Python caches the result of that lookup keyed on a hash stored in each object or something like that.  I'll see if I can usefully profile that.

dscorbett:
The fastest solution would be to not refer to such routines by strings at all. Rather, refer to them directly. That is, in `generate_among_table`, don’t add the quotation marks and don’t explicitly mangle the name. This would require defining all the `among` lists after the routines, and it would change how the method needs to be called in basestemmer.py. I chose this PR’s solution because it was the most minimal, but I can submit a PR for that alternative too.

ojwb:
Thanks, that does sound a better approach than using a string, though I seem to be unable to measure any difference in speed between the current code, the current code but with a shorter name, and this approach (I used the Python `timeit` module and a 40 character Snowball routine name, but I get almost exactly the same spread of times with `-r 100` between repeat runs as I do between different approaches).

I've pushed a change to do that - I think that means we don't need the change here.

I'm wondering if we should still make `__` in a Snowball identifier invalid.  It doesn't seem like it is something that's useful to allow, but to correctly support it we would need to do some sort of mangling or renaming when generating C++ code (probably also for C in case people are compiling it as C++).  It may also require special handling in other current and/or future target languages as it seems quite common for double underscores in identifiers to have a special meaning in some cases.

We could just add a documentation note saying that such names are better avoided, but then people might not see that note and a compiler error is hard to miss and not much harder to implement.

I'm guessing you weren't actually trying to name a Snowball routine `x__` but rather noticed this from looking at the generated code and knowing Python's name mangling rules?

dscorbett:
I don’t think Snowball should ban `__` in identifiers. Different languages have different identifier rules and it would be too restrictive for Snowball’s identifiers to follow all of them. Each language’s generator should be responsible for whatever mangling is necessary, as is already done for Ada, Go, and Pascal. I like your idea about generating new names for all non-externals. That would also fix a problem with the mangling implementation for Ada, which currently can map different identifiers together.

Right, I was not actually trying to use the identifier `x__`.

I expect (without having checked) that the Python code is slow primarily because it uses exceptions for control flow. That seems necessary in the general case, but one common pattern is this:
```python
        try:
            if not self.go_out_grouping(FrenchStemmer.g_v, 97, 251):
                raise lab4()
            self.cursor += 1
            if not self.go_in_grouping(FrenchStemmer.g_v, 97, 251):
                raise lab4()
            self.cursor += 1
            self.I_p1 = self.cursor
            if not self.go_out_grouping(FrenchStemmer.g_v, 97, 251):
                raise lab4()
            self.cursor += 1
            if not self.go_in_grouping(FrenchStemmer.g_v, 97, 251):
                raise lab4()
            self.cursor += 1
            self.I_p2 = self.cursor
        except lab4: pass
```
If all the exceptions within a `try` block are raising the same exception, which is mentioned in the associated `except` block, and there are no nested `try` blocks, it is possible to rewrite it to:
```python
        while True:
            if not self.go_out_grouping(FrenchStemmer.g_v, 97, 251):
                break
            self.cursor += 1
            if not self.go_in_grouping(FrenchStemmer.g_v, 97, 251):
                break
            self.cursor += 1
            self.I_p1 = self.cursor
            if not self.go_out_grouping(FrenchStemmer.g_v, 97, 251):
                break
            self.cursor += 1
            if not self.go_in_grouping(FrenchStemmer.g_v, 97, 251):
                break
            self.cursor += 1
            self.I_p2 = self.cursor
            break
```
or:

```python
        if self.go_out_grouping(FrenchStemmer.g_v, 97, 251):
            self.cursor += 1
            if self.go_in_grouping(FrenchStemmer.g_v, 97, 251):
                self.cursor += 1
                self.I_p1 = self.cursor
                if self.go_out_grouping(FrenchStemmer.g_v, 97, 251):
                    self.cursor += 1
                    if self.go_in_grouping(FrenchStemmer.g_v, 97, 251):
                        self.cursor += 1
                        self.I_p2 = self.cursor
```
which might help.

ojwb:
Hmm, I'd not properly looked at that code for mangling identifiers for Ada before - it's not very well thought through.

Looking at the Ada docs, it seems Ada identifiers can't contain two consecutive underscores (which the mangling doesn't attempt to address) or end in an underscore (which it addresses by appending `E`, causing a collision for Snowball identifiers `x_` and `x_E` of the same type, for example).

It also uppercases the first character of the Snowball identifier so `x` and `X` with the same type will also collide.  I don't see a reason why it does this though.  Looking at the history of the code, it was originally a copy of the Pascal generator and this upper-casing seems to have evolved from upper-casing the first letter of `externals` in Pascal (which it seems isn't really needed as Pascal identifiers are case insensitive).

I have some reservations about generating all internal identifiers because they become more cryptic and make it harder to follow the generated code, which makes it harder to debug.  I've enhanced the `-comments` option to add a comment before each routine, etc which at least means turning on comments makes it clear which Snowball routine is which in the target code.

Maybe we should switch to generated identifiers and see how annoying it is in practice.  We could later add an option to use identifiers based on Snowball names to aid debugging, or change back to using such identifiers for the common situation where generating using the snowball name is not problematic in the target language.

Re the generated python code:

The heavy use of exceptions in the generated Python code is certainly not ideal.  Using `break` won't work everywhere because we generate Python `while` loops for some other Snowball constructs.  We'd have to inspect the code inside to see if one of these constructs is used (optionally checking if there's actually anything inside that which might need to fail to outside).

I have actually previously looked to see if we could generate code more like your last example but I couldn't see a way to do it easily.  I suspect we could create a network of basic blocks and the control flow between them and then have some algorithm to find a good way to flatten that into Python code.

Mostly I've been hoping some keen Python hacker would appears as this is mostly a Python problem - the other languages all have goto or a way to break from a labelled block; they're also compiled (only usually so in the case of Javascript) and so we can just let the target language compiler optimise out redundant control flow.

ojwb:
Incidentally I did a quick experiment to try generating the `while` with `break` alternative.  It's fiddly because `while` is used for other purposes so we need to check for that, but I've got it generating code with fewer uses of `try` which seems to work and be a bit faster.

The hard part really is deciding when we can safely use it, and once we have that it's mainly just a tweak on the code that currently generates a `try` block to check if it should use `while`/`break` instead.  I'm not confident the logic of this hard part is currently correct though so I'm thinking to look at merging it after the release I'm trying to focus on getting us to.

The `if` cascade approach would be better still really, but that seems to require more major restructuring.

PR #151, opened 2021-08-31
Merge Czech stemmer

ojwb:
This has been on the web site since 2012, but never actually got
included in the code distribution.

Points to resolve:

- [x] R1 and RV
  - [x] syllabic constants in definitions (r, l, maybe m, maybe n). https://github.com/snowballstem/snowball/pull/151#issuecomment-2303057423
  - [x] when to apply them
 - [x] `č` suffix in snowball vs `če` in Java (Snowball seems to have copied `-č` typo in Java comment)
 - [x] `čtí`/`ští` in Java vs `čté`/`šté` in Snowball (again seems to be due to Java comment typo)
 - [x] `len- 2` instead of `len- 3` for Java `ště`/`šti`/`ští` check. Seems fairly clear improvement.
 - [x] Snowball version removes extra character before calling `palatalise`.
 - [x] Snowball version doesn't remove final character by default if `palatalise` doesn't otherwise match.
 - [x] if `do_case` doesn't make a replacement then `do_possessive` won't get called, but in the java code, `removePossessives` is always called. https://github.com/snowballstem/snowball/pull/151#issuecomment-1791896886
 - [x] Java version leaves the first character of a removed suffix behind when calling `palatalise` **except** for `-es`/`-ém`/`-ím`
 - [x] `setlimit tomark p1 for ([substring])` vs `[substring] R1`
 - [x] Look into addressing unhelpful splits from palatalise change
 - [x] `-ětem` isn't listed by https://en.wikipedia.org/wiki/Czech_declension but seems to be valid from e.g. https://en.wiktionary.org/wiki/hrab%C4%9B https://en.wiktionary.org/wiki/markrab%C4%9B and https://en.wiktionary.org/wiki/ml%C3%A1d%C4%9B
 - [x] `-os`, `-es`, `-iho`, `-imu` aren't listed by https://en.wikipedia.org/wiki/Czech_declension
 - [x] `-ich` seems to only be a suffix for two pronouns
 - [x] Should we also remove `-ima`? Probably not.
 - [x] Should we also remove `-ímu` (with a diacritic on the `i`)? Yes.
[...]

PR #220, opened 2025-03-14
Improved Polish stemmer from #159

mitya57:
I took the stemmer which was originally written by @tomek-ai in #159 and rewrote it using substring checks.

The original commit is preserved (but rebased), my modifications are in separate commits for review purposes (please feel free to squash when merging). My first commit does not cause behavior changes, the subsequent commits do.

There is some room for improvement, e.g. the stemmer does not recognize all possible superlative forms (`naj-`), but it handles simpler cases quite well.

The corresponding snowball-data PR is snowballstem/snowball-data#30.

### DIFF
--- a/algorithms/polish.sbl
+++ b/algorithms/polish.sbl
@@ -78,11 +78,16 @@ backwardmode (
 
   define remove_diminutive as (
     setlimit tomark p1 for ([substring]) among (
-      'eczek'  // e.g. człowieczek (diminutive of człowiek)
-      'iczek'  // e.g. czajniczek (diminutive of czajnik)
-      'iszek'  // e.g. braciszek (diminutive of brat)
-      'aszek'  // e.g. młodzieniaszek (diminutive of młodzieniec)
-      'uszek'  // e.g. kwiatuszek (diminutive of kwiat)
+      'eczek' 'eczka' 'eczkowi' 'eczkiem' 'eczku'      // e.g. człowieczek (diminutive of człowiek)
+      'eczki' 'eczk{o'}w' 'eczkom' 'eczkami' 'eczkach' // plural
+      'iczek' 'iczka' 'iczkowi' 'iczkiem' 'iczku'      // e.g. czajniczek (diminutive of czajnik)
+      'iczki' 'iczk{o'}w' 'iczkom' 'iczkami' 'iczkach' // plural
+      'iszek' 'iszka' 'iszkowi' 'iszkiem' 'iszku'      // e.g. braciszek (diminutive of brat)
+      'iszki' 'iszk{o'}w' 'iszkom' 'iszkami' 'iszkach' // plural
+      'aszek' 'aszka' 'aszkowi' 'aszkiem' 'aszku'      // e.g. młodzieniaszek (diminutive of młodzieniec)
+      'aszki' 'aszk{o'}w' 'aszkom' 'aszkami' 'aszkach' // plural
+      'uszek' 'uszka' 'uszkowi' 'uszkiem' 'uszku'      // e.g. kwiatuszek (diminutive of kwiat)
+      'uszki' 'uszk{o'}w' 'uszkom' 'uszkami' 'uszkach' // plural
       'ek'     // e.g. ogródek (diminutive of ogród)
       'ak'     // e.g. gęsiak (diminutive of gęś)
         (delete)


algorithms/polish.sbl
```diff
@@ -0,0 +1,178 @@
...
+	remove_general_ends
+)
+
+
+backwardmode (
+	define remove_nouns as (
+		(
+			$(len > 7)
+			[substring] among (
+				'zacja' 'zacji'
```
ojwb, line 33: The corresponding original code was

```
('zacja' or 'zacja' or 'zacji')
```

which duplicates `'zacja'`.

I'm not familiar with the Polish language, but I notice the two other suffixes ending `cja` both have a corresponding suffix ending `cj{a'}` so I wonder if maybe one `'zacja'` here should have been `'zacj{a'}'`?

ojwb:
Thanks, this is definitely an improvement.  Especially that it's now mostly working `backwards` rather than trying to remove suffixes but while working forwards.

> If you want, I can add explanatory comments what each ending means, maybe with examples.

That would be very useful.  We really need some documentation of the algorithm or else it's much harder to maintain going forwards because we have to guess at the intent; having meanings and examples for each ending would cover a significant part of that.

algorithms/polish.sbl
mitya57: Yes, you are right, `-zacją` is the accusative case of `-zacja`, probably it was meant here. Interestingly, adding it does not affect the tests, probably because all such words are also covered by the next case for `-acją`.

algorithms/polish.sbl
```diff
@@ -0,0 +1,179 @@
+stringescapes {}
+
+stringdef a' '{U+0105}' 
+stringdef e' '{U+0119}' 
+stringdef l' '{U+0142}' 
+stringdef s' '{U+015b}' 
+stringdef c' '{U+0107}' 
+stringdef o' '{U+00f3}'
```
ojwb, line 8: These are (from `UnicodeData.txt`):

```
00F3;LATIN SMALL LETTER O WITH ACUTE;Ll;0;L;006F 0301;;;;N;LATIN SMALL LETTER O ACUTE;;00D3;;00D3
0105;LATIN SMALL LETTER A WITH OGONEK;Ll;0;L;0061 0328;;;;N;LATIN SMALL LETTER A OGONEK;;0104;;0104
0107;LATIN SMALL LETTER C WITH ACUTE;Ll;0;L;0063 0301;;;;N;LATIN SMALL LETTER C ACUTE;;0106;;0106
0119;LATIN SMALL LETTER E WITH OGONEK;Ll;0;L;0065 0328;;;;N;LATIN SMALL LETTER E OGONEK;;0118;;0118
0142;LATIN SMALL LETTER L WITH STROKE;Ll;0;L;;;;;N;LATIN SMALL LETTER L SLASH;;0141;;0141
015B;LATIN SMALL LETTER S WITH ACUTE;Ll;0;L;0073 0301;;;;N;LATIN SMALL LETTER S ACUTE;;015A;;015A
```

So `'` is being used for 3 different diacritics here, which is confusing.

Based on usage in other stemmers (and what's [recommended on the website](https://snowballstem.org/codesets/guide.html)) I'd suggest changing to `{ak}`, `{ek}` and `{l/}` (`'` for acute matches recommended/existing usage).

algorithms/polish.sbl
ojwb: Good spot, and it actually seems that first set of endings here is entirely redundant so might as well be removed - if `$(len > 7)` then `$(len > 6)` so we'll definitely do the next set and that does equivalent changes to everything in the first set.

If linguistically `-zacja`, etc are different endings we can just note in comments that `acja` handles both `-acja` and `-zacja`.

algorithms/polish.sbl
```diff
@@ -0,0 +1,179 @@
...
+			backwards (
+				[substring] among (
+					'sze' 'szy' 'szych'
+					// Add more forms like 'szego', 'szemu', etc?
+						(delete)
+				)
+			)
+		)
+	)
+	['naj'] delete
```
ojwb, line 163: (Not such much a review comment, just me thinking out loud...)

This seems a bit clumsy, but I don't really see a better way.

If snowball had a way to delete everything *EXCEPT* the current slice it could just be something like:

```
  'naj'
  [ backwards among (
    'sze' 'szy' 'szych'
    // Add more forms like 'szego', 'szemu', etc?
  ) ] delete_all_but_the_slice
```

Not sure if that operation would be more widely useful though.

mitya57: Yes, I spent some time figuring out how to implement it. The only similar thing that I found in other algorithms is `Lose_prefix` function in Dutch, that removes `ge-` prefix, although it seems to remove it from all words, no matter if they have `-en` or similar suffix.

In Polish we can remove `naj-` in most of the cases, but not all, because there are some words with `na-` prefix and root starting with `j`, e.g. `najechać` which is `na-` + `jechać`.

ojwb: I think for now we go with your solution, but probably tweak the final `['naj']` with `[tomark sizeof 'naj']` as a little optimisation - that will compile to (e.g. for Python):

```
        self.bra = self.cursor
        if self.cursor > 3:
            return False
        self.cursor = 3
        self.ket = self.cursor
```

rather than having to call `eq_s` and actually check the substring matches:

```
        self.bra = self.cursor  
        if not self.eq_s(u"naj"):
            return False
        self.ket = self.cursor
```

If the compiler tracked possible current values (or ranges of values) for the cursor and the current string then it should be able to emit this given either:

```
        self.bra = self.cursor
        self.cursor = 3
        self.ket = self.cursor
```

That's probably going to be easier to pull off for my revised version though as tracking things we know about the value of the current string through code that modifies parts of it seems a harder problem.

Longer term we can extend the Snowball language with `delete_all_but_the_slice` (but with a better name) or another feature which allows a cleaner solution to this, but I'd rather not rush to do that in case we end up with something we later regret.

Also, could you repush the snowball-data branch from your own repo with the same name as this branch i.e. `polish-stemmer-2`?  Then the CI will know to use the new polish vocabulary.

mitya57:
Updated to add comments and examples (I hope I can use unicode in comments? The Lithuanian stemmer also has them). And addressed all your other comments.

Unfortunately the Ada tests fail. I have no experience with that language, but I can take a look tomorrow.

Also should I replace tabs with spaces in indentation? The original file used tabs, but I see that most existing stemmers (all except Nepali) use spaces.

mitya57:
Regarding Ada tests: it fails on `delete` command on `[tomark sizeof 'naj'] delete` line.

This patch makes the tests pass, but I'm not sure if it's the proper fix:
```diff
--- a/ada/src/stemmer.ads
+++ b/ada/src/stemmer.ads
@@ -181,11 +181,11 @@ private
                       S          : in String;
                       Adjustment : out Integer) with
      Global => null,
-     Pre => C_Bra >= Context.Lb and C_Ket >= C_Bra and C_Ket <= Context.L;
+     Pre => C_Ket >= C_Bra and C_Ket <= Context.L;
 
    procedure Slice_Del (Context : in out Context_Type'Class) with
      Global => null,
-     Pre => Context.Bra >= Context.Lb and Context.Ket >= Context.Bra
+     Pre => Context.Ket >= Context.Bra
      and Context.Ket <= Context.L;
 
    procedure Slice_From (Context : in out Context_Type'Class;
```

ojwb:
Those Ada preconditions should not include anything to do with `Context.L` or `Context.Lb` - the manual says the limits apply to cursor movement, so it should be allowed to modify a slice outside of the limits.

Dropping the invalid preconditions is easy, but the Ada code also seems to use `Context.L` in place of tracking the current length of the string.  That's probably OK for the final result (`setlimit C1 for C2` changes the limit for C2, but then it gets reverted, so the limits on exit should be the start and end of the final string) but it means we don't know where the actual end of the string is so a `delete`, `insert`, etc doesn't know how much of the tail of the string to copy.  Just removing the pre-condition will make the code buggy if there's a forwards limit (i.e. `Context.L` < the current true length of the string).  So I think we need to separately track the string length.  I'll sort out a fix.  Anyway I agree this doesn't look like a problem with `polish.sbl`.

Unicode is fine in comments (sources should be UTF-8 with Unix line endings).

The intention is to only use spaces for indentation except where there's a syntax reason (e.g. `GNUmakefile`) or a strong convention (e.g. Go standard seems to be tabs: `gofmt` reindents code with tabs).  I've just pushed 7a0e9ed78caac97fbc002fe7a589dd7077a19de9 to clean up the deviations from this.

algorithms/polish.sbl
```diff
@@ -0,0 +1,225 @@
+stringescapes {}
+
+stringdef ak '{U+0105}' // ę e + ogonek
+stringdef ek '{U+0119}' // ą a + ogonek
```
ojwb, line 4: The comments seem swapped here.

algorithms/polish.sbl
```diff
@@ -0,0 +1,225 @@
...
+  define remove_verbs_ends as (
+    (
+      $(len > 5)
+      [substring] among (
+        'bym'        // conditional 1st person singular, e.g. czytał(a)bym
+                     // TODO: maybe also add 'by{s'}' for 2nd person?
+        'esz'        // present 2nd person singular, e.g. piszesz
+        'asz'        // present 2nd person singular, e.g. czytasz
+        'cie'        // present 2nd person plural, e.g. czytacie
+        'e{s'}{c'}'  // infinitive, e.g. nieść
```
ojwb, line 124: The example fails the `$(len > 5)` requirement above.

algorithms/polish.sbl
```diff
@@ -0,0 +1,225 @@
...
+    ) or (
+      $(len > 5)
+      [substring] among (
+        'ach'  // plural locative
+        'ami'  // plural instrumental
+        'nia'  // genitive of words ending with -nie, e.g. czytanie - czytania
+        'niu'  // dative/locative of words ending with -nie, e.g. czytanie - czytaniu
+        'cia'  // genitive of words ending with -cie, e.g. ukrycie - ukrycia
+        'ciu'  // dative/locative of words ending with -cie, e.g. ukrycie - ukryciu
+        'ce'   // dative/locative of words ending with -ka, e.g. rzeka - rzece
```
ojwb, line 55: The example fails the `$(len > 5)` requirement above.

algorithms/polish.sbl
```diff
@@ -0,0 +1,225 @@
...
+      [substring] among (
+        'eczek'  // e.g. człowieczek (diminutive of człowiek)
+        'iczek'  // e.g. czajniczek (diminutive of czajnik)
+        'iszek'  // e.g. braciszek (diminutive of brat)
+        'aszek'  // e.g. młodzieniaszek (diminutive of młodzieniec)
+        'uszek'  // e.g. ‎kwiatuszek (diminutive of ‎kwiat)
+          (delete)
+        'enek'   // e.g. ogienek (diminutive of ogień)
+          (<- 'en')
+        'ejek'   // e.g. olejek (diminutive of olej)
```
ojwb, line 79: The example fails the `$(len > 6)` requirement above.

ojwb: However `olejek` is stemmed to `olej` by the `ek` rule in the group below, so this rules seems to be unnecessary (same for the `enek` and `erek` rules before/after it).

algorithms/polish.sbl
```diff
@@ -0,0 +1,225 @@
...
+    (
+      $(len > 6)
+      [substring] among (
+        // Borrowed words that match English -ation, e.g. fortyfikacja (fortification).
+        'acja'     // nominative
+        'acji'     // genitive/dative/locative
+        'acj{ak}'  // accusative
+        // Plural locative ending of (rare) words ending with -ę, e.g. zwierzę - zwierzętach.
+        // TODO: It can also be roots ending with -t and -ach ending, e.g. pirat - piratach.
+        'tach'
```
ojwb, line 37: Based on the comment, the obvious thing to try seemed to be:

```
diff --git a/algorithms/polish.sbl b/algorithms/polish.sbl
index eb8dca53..e992e12d 100644
--- a/algorithms/polish.sbl
+++ b/algorithms/polish.sbl
@@ -35,6 +35,7 @@ backwardmode (
         // Plural locative ending of (rare) words ending with -ę, e.g. zwierzę - zwierzętach.
         // TODO: It can also be roots ending with -t and -ach ending, e.g. pirat - piratach.
         'tach'
+          ('{ek}' delete)
         // Verbal nouns, e.g. czytanie, narodzenie.
         'anie' 'enie'  // nominative
         'aniu' 'eniu'  // dative/locative
```

The `stemmer-compare` script says:

<ul>
<li>A total of 178 words changed stem
<li>161 words changed stem but aren't interesting
<li>11 merges of groups of stems:
<div>
<div title="aforyst">{ aforystach } + { aforystów }</div>
<div title="czworolist">{ czworolistach } + { czworolisty }</div>
<div title="ergoprojekt">{ ergoprojektach } + { ergoprojekt }</div>
<div title="herbat">{ herbatach } + { herbatowata }</div>
<div title="homeopat">{ homeopatach } + { homeopatiom }</div>
<div title="kabalist">{ kabalistach } + { kabalistą }</div>
<div title="kazuist">{ kazuistach } + { kazuistom }</div>
<div title="petynet">{ petynetach } + { petynetom }</div>
<div title="półmat">{ półmatach } + { półmatowy półmatów }</div>
<div title="teoremat">{ teorematach } + { teorematem }</div>
<div title="żarot">{ żarotach } + { żarotami }</div>
</div>
<li>5 splits of groups of stems:
<div>
<div>{ <span title="ban">banałami banowania</span> | <span title="bant">bantach</span> }</div><div>{ <span title="domin">dominiów</span> | <span title="dominac">dominacją</span> }</div><div>{ <span title="przeor">przeorowa</span> | <span title="przeorat">przeoratach</span> }</div><div>{ <span title="rekr">rekruta</span> | <span title="rekrut">rekrutach</span> }</div><div>{ <span title="ubikwis">ubikwista</span> | <span title="ubikwist">ubikwistach</span> }</div></div>
<li>3 words moving between stem groups:

[Image showing "krztach" was grouped with "krzemek" but is now grouped with "krztą"]

Based on just pasting stuff into google translate (which can be randomly bad from past experience) it seems the merges are all good but the splits are a mixed bag (e.g. `{ rekruta | rekrutach }` is apparently "recruit"/"recruits" so was better conflated).  The "moving between" case seems to be 3 unrelated words so not really better or worse.

(Note you can mouse-over words in the splits and merges to see what they stem to.)

ojwb:
> Dropping the invalid preconditions is easy, but the Ada code also seems to use `Context.L` in place of tracking the current length of the string. That's probably OK for the final result (`setlimit C1 for C2` changes the limit for C2, but then it gets reverted, so the limits on exit should be the start and end of the final string) but it means we don't know where the actual end of the string is so a `delete`, `insert`, etc doesn't know how much of the tail of the string to copy. Just removing the pre-condition will make the code buggy if there's a forwards limit (i.e. `Context.L` < the current true length of the string). So I think we need to separately track the string length. I'll sort out a fix. Anyway I agree this doesn't look like a problem with `polish.sbl`.

Dear me, quite a lot of wrong stuff in Ada runtime support (and the generator too).  I've pushed fixes for the stuff that's relevant here so hopefully the next build will be all green.

algorithms/polish.sbl
mitya57: Until now I tried to preserve 100% compatibility with snowballstem/snowball-data#21. Now it looks like I will need to deviate from that and make my own PR for snowball-data.

The merges look mostly correct (*półmatowy* is an adjective, but derived from *półmat* so it should be fine).
First three splits are good (although ideally *banałami* should be split from *banowania* too). Last two splits are wrong: these are two forms of *rekrut* and two forms of *ubikwist*.

I will probably just do *ętach → ę* without touching unrelated words with *-tach*, that looks safer to me.

mitya57: I tried to remove the rules for *-tach* and *-ta* completely and looked at the diff reported by the tests. In that diff, I found only two words that belong to the rare [*-ę*](https://en.wiktionary.org/wiki/-%C4%99#Polish) declension (which means young people/animals): *królewięta* (literally king’s children) and *wnuczęta* (grandchildren).

Most other affected words which had *ę* before *ta* are feminine passive participles, e.g. *wzdęta* (from the verb *wzdąć*), *wysunięta* (from the verb *wysunąć*), *dopadnięta* (from the verb *dopaść*) and so on.

So, by keeping the *-tach* and *-ta* rules around we had more false positives than true positives, and I propose to drop these rules completely.

There is some chance that I did not understand @tomeklysiak’s intention here, and the purpose of *-tach* and *-ta* rules was something different. But still, the fact that these rules are in `remove_nouns` function, and many affected words are participles, signals that the rule did not work as intended.

algorithms/polish.sbl
```diff
@@ -0,0 +1,219 @@
...
+
+  define remove_dimunitive as (
+    (
+      $(len > 6)
+      [substring] among (
+        'eczek'  // e.g. człowieczek (diminutive of człowiek)
+        'iczek'  // e.g. czajniczek (diminutive of czajnik)
+        'iszek'  // e.g. braciszek (diminutive of brat)
+        'aszek'  // e.g. młodzieniaszek (diminutive of młodzieniec)
+        'uszek'  // e.g. ‎kwiatuszek (diminutive of ‎kwiat)
```
ojwb, line 75: Locally this line has two U+200E characters ("LEFT-TO-RIGHT MARK") which seems unnecessary since all the text is in LTR scripts, and better avoided.

Unhelpfully they don't seem to show up in the github diff viewer...

algorithms/polish.sbl
```diff
@@ -0,0 +1,219 @@
...
+  define remove_adverbs_ends as (
+    (
+      $(len > 4)
+      [substring] among (
+        'nie'  // e.g. dokładnie
+          (<- 'n')
+        'wie'  // e.g. możliwie
+          (<- 'w')
+        //'rze'  // e.g. dobrze
+        //  (<- 'r')
```
ojwb, line 152: Apparently this algorithm is a reimplementation in Snowball of https://github.com/Tutanchamon/pl_stemmer/blob/master/pl_stemmer.py

I went through and compared and it does seem to be equivalent (one bug in the original aside which I'll comment on separately), except that the equivalent rule to this is **not** commented out there.  If I uncomment it here, the effect on stemming the sample vocabulary is:

<ul>
<li>A total of 93 words changed stem
<li>86 words changed stem but aren't interesting
<li>7 merges of groups of stems:
<div>
<div title="danser">{ danserów } + { danserze }</div>
<div title="filmistor">{ filmistor } + { filmistorze }</div>
<div title="gabr">{ gabrowe } + { gabrze }</div>
<div title="integrimetr">{ integrimetr } + { integrimetrze }</div>
<div title="lur">{ lur } + { lurze }</div>
<div title="sakr">{ sakry } + { sakrze }</div>
<div title="ser">{ seria } + { serze }</div>
</div>
</ul>

I can't find translations for some of these, but at least the first and last seem helpful.

Some options are:

* leave it out
* uncomment it
* uncomment it but adjust the prefix length requirement to exclude cases we don't want it to fire for
* uncomment it but add some requirement of what comes before the `rze` to exclude cases we don't want it to fire for

algorithms/polish.sbl
```diff
@@ -0,0 +1,219 @@
...
+        'e{c'}'  // infinitive, e.g. chcieć
+        'i{c'}'  // infinitive, e.g. wozić
+        'em'     // past 1st person singular masculine, e.g. czytałem
+        'am'     // past 1st person singular feminine, e.g. czytałam
+        'a{l/}'  // past 3rd person singular masculine, e.g. czytał
+        'i{l/}'  // past 3rd person singular masculine, e.g. woził
+        '{ak}c'  // contemporary adverbial participle (transgressive), e.g. czytając
+          (delete)
+        'aj'     // imperative, e.g. czytaj
+          (<- 'a')
```
ojwb, line 138: In the Python original, the "aj" rule is buggy and will never fire because `word[-3:]` will be 3 characters long:

```
    if len(word) > 3 and word[-3:] in {"aj"}:
        return word[:-1]
```

This seems an obvious typo, but any evaluation of the Python version may not have included this rule (depending when the typo was introduced - it seems likely it was there from the start as the rules above test 3 character substrings).

So I did a quick compare to see the effect of commenting this out here - NB this is going FROM having the rule TO not having it:

<ul>
<li>A total of 119 words changed stem
<li>93 words changed stem but aren't interesting
<li>5 merges of groups of stems:
<div>
<div title="podglądaj">{ podglądaj } + { podglądające }</div>
<div title="przyzwyczaj">{ przyzwyczajeniu } + { przyzwyczajamy }</div>
<div title="spieraj">{ spieraj } + { spierającej }</div>
<div title="wrębiaj">{ wrębiaj } + { wrębiające }</div>
<div title="wzrastaj">{ wzrastaj } + { wzrastającego wzrastającej }</div>
</div>
<li>18 splits of groups of stems:
<div>
<div>{ <span title="nawypraw">nawyprawiamy</span> | <span title="nawyprawiaj">nawyprawiaj</span> }</div><div>{ <span title="obrast">obrastałem</span> | <span title="obrastaj">obrastaj</span> }</div><div>{ <span title="odpożycz">odpożyczam</span> | <span title="odpożyczaj">odpożyczaj</span> }</div><div>{ <span title="odtlen">odtlenił</span> | <span title="odtleniaj">odtleniaj</span> }</div><div>{ <span title="pomniejsz">pomniejszałem</span> | <span title="pomniejszaj">pomniejszaj</span> }</div><div>{ <span title="przeciek">przeciekom</span> | <span title="przeciekaj">przeciekaj</span> }</div><div>{ <span title="rozkurz">rozkurzał rozkurzcie</span> | <span title="rozkurzaj">rozkurzaj</span> }</div><div>{ <span title="rozżer">rozżeram</span> | <span title="rozżeraj">rozżeraj</span> }</div><div>{ <span title="szam">szamcie</span> | <span title="szamaj">szamaj</span> }</div><div>{ <span title="unaturaln">unaturalniacie unaturalniał</span> | <span title="unaturalniaj">unaturalniaj</span> }</div><div>{ <span title="upiększ">upiększeniu</span> | <span title="upiększaj">upiększaj</span> }</div><div>{ <span title="wychark">wycharkać</span> | <span title="wycharkaj">wycharkaj</span> }</div><div>{ <span title="wysnuw">wysnuwania</span> | <span title="wysnuwaj">wysnuwaj</span> }</div><div>{ <span title="zaczern">zaczerniamy</span> | <span title="zaczerniaj">zaczerniaj</span> }</div><div>{ <span title="zamarz">zamarzłbym</span> | <span title="zamarzaj">zamarzaj</span> }</div><div>{ <span title="zmierz">zmierzaniu</span> | <span title="zmierzaj">zmierzaj</span> }</div><div>{ <span title="zostaw">zostawiam zostawiałem zostawicie</span> | <span title="zostawaj">zostawaj</span> }</div><div>{ <span title="zruch">zruchałem</span> | <span title="zruchaj">zruchaj</span> }</div></div>
<li>9 words moving between stem groups:
<br>

compare

algorithms/polish.sbl
ojwb: Just to note now we know the origin of the algorithm, https://github.com/Tutanchamon/pl_stemmer/blob/master/pl_stemmer.py#L200 has `zacją` so this fixes a deviation from that:

```
    if len(word) > 7 and word[-5:] in {"zacja", u"zacją", "zacji"}:
```

algorithms/polish.sbl
mitya57: Oops, fixed.

algorithms/polish.sbl
mitya57: Looking at the last picture, neither the old variant nor the new one is perfect. On the left, we group participle separately from imperative and infinitive. On the right, we group infinitive separately from the participle and imperative. I think we should add more possible endings for participles to merge all forms together.

Can you please run this through your script and see if it makes it better?
```diff
@@ -159,6 +159,17 @@ backwardmode (
         'a{l/}'  // past 3rd person singular masculine, e.g. czytał
         'i{l/}'  // past 3rd person singular masculine, e.g. woził
         '{ak}c'  // contemporary adverbial participle (transgressive), e.g. czytając
+        // different forms of active adjectival participle
+        '{ak}cy'
+        '{ak}cego'
+        '{ak}cemu'
+        '{ak}cym'
+        '{ak}ca'
+        '{ak}cej'
+        '{ak}c{ak}'
+        '{ak}ce'
+        '{ak}cych'
+        '{ak}cymi'
           (delete)
         'aj'     // imperative, e.g. czytaj
           (<- 'a')
```

Alternatively, we can move the handling of *ąc* from the function for verbs to function for adjectives. This way we remove the possible adjectival endings (*-y*, *-ego*, *-emu* and so on), and if there was *ąc* before the ending, remove it too. This is what Russian stemmer does (Polish *ąc*/*jąc* correspond to Russian *ущ*/*ющ*, and it's handled on [line 116](https://github.com/snowballstem/snowball/blob/master/algorithms/russian.sbl#L116)).

algorithms/polish.sbl
ojwb: It's certainly OK to produce a linguistically incorrect stem, but indeed if we do we ideally still want to map other forms of the word to the same stem (and actually more importantly that we don't map forms of unrelated words to that stem).

algorithms/polish.sbl
ojwb: Ah right, I'd not spotted that this rule was uncommented but effectively not enabled due to a bug in the original Snowball version.  I agree we should just uncomment this.

> These 7 merges are mostly good. The last one merges unrelated words (_seria_ is a Latin word that means series, _serze_ is locative of _ser_, which means cheese). However reducing _serze_ to _ser_ is still correct, and if we wanted to avoid the merge we would better change stemming of _seria_ to _seri_.

Yes, if that's worth addressing then adjusting how _seria_ stems seems a better approach.  Are there other Polish words (probably also from Latin) which follow a similar pattern?

algorithms/polish.sbl
ojwb: > Can you please run this through your script and see if it makes it better?

I'll do that shortly, but you can do it locally pretty easily:

* Install graphviz (for `dot`)
* Check out the snowball-data repo into a sibling directory to the snowball one (you probably already have this set up so you can run the snowball repo testsuite)
* in the snowball repo run `make check_utf8_polish` - if this fails it will leave the wrong stemmed output in `tmp.txt`
* in the snowball-data repo run `scripts/stemmer-compare polish` - with a single argument this uses `voc.txt` and `output.txt` from that subdirectory of snowball-data and `../snowball/tmp.txt` as the new version of output.txt
* `firefox compare.html` (other browsers probably work but might need a full URL; inline SVG is used by default but you can change `$image_format` in the script to get e.g. PNG if the inline SVG doesn't work for you)

ojwb:
Snowball is currently all licensed as 3-clause BSD, and I really don't want to start to allow a proliferation of licences.  MIT/X is a broadly equivalent licence, but it still makes it harder for users to understand the licensing situation.

The original PR author says in https://github.com/snowballstem/snowball-website/pull/29#discussion_r2002201169 :

> The implementation in Snowball was created independently, using the same principles but rewritten from scratch and adapted to Snowball’s structure.

Copyright applies to the implementation of an algorithm, so a rewrite from scratch can use a different choice of licence.

(We should of course acknowledge the origin of the algorithm, but we don't need to use the same licence to do so.)

algorithms/polish.sbl
ojwb: The changes are all merges which is promising, though some could be unwanted merges.

<ul>
<li>A total of 1656 words changed stem
<li>1386 words changed stem but aren't interesting
<li>225 merges of groups of stems:
<div>
<div title="amnestiuj">{ amnestiuję } + { amnestiującym }</div>
<div title="angielszcz">{ angielszczące } + { angielszczącymi }</div>
<div title="aprowiduj">{ aprowidujące } + { aprowidującą }</div>
<div title="biadol">{ biadol } + { biadolący }</div>
<div title="blikuj">{ blikując } + { blikująca } + { blikującymi }</div>
<div title="brązow">{ brązowiących brązowić } + { brązowiącymi }</div>
<div title="charcz">{ charczącemu } + { charczącymi }</div>
<div title="chlej">{ chlejąc } + { chlejącą }</div>
<div title="chrup">{ chrup } + { chrupiący }</div>
<div title="chrustaj">{ chrustającego } + { chrustająca }</div>
<div title="chłon">{ chłoniecie } + { chłonąca }</div>
<div title="czadz">{ czadzenie } + { czadzącymi }</div>
<div title="czworakuj">{ czworakujesz } + { czworakującym }</div>
<div title="czyn">{ czynienia } + { czyniącym }</div>
<div title="częstuj">{ częstuję } + { częstujący }</div>
<div title="daszkuj">{ daszkujcie } + { daszkującemu }</div>
<div title="debarkuj">{ debarkujących } + { debarkującym }</div>
<div title="dekompletuj">{ dekompletujesz } + { dekompletującą }</div>
<div title="depiluj">{ depilującemu } + { depilującym }</div>
<div title="domagaj">{ domagającą } + { domagającymi }</div>
<div title="doszlifowuj">{ doszlifowujemy } + { doszlifowującą }</div>
<div title="dozuj">{ dozujcie dozują } + { dozującym }</div>
<div title="dośrubowuj">{ dośrubowują } + { dośrubowującą }</div>
<div title="drapaczuj">{ drapaczujcie } + { drapaczującą }</div>
<div title="dribluj">{ driblujesz } + { driblującą }</div>
<div title="dygocz">{ dygoczę } + { dygoczący }</div>
<div title="dyssymuluj">{ dyssymuluj } + { dyssymulującym }</div>
<div title="dystansuj">{ dystansując } + { dystansującymi }</div>
<div title="dziab">{ dziab } + { dziabiącą }</div>
<div title="ekranuj">{ ekranujących } + { ekranującym }</div>
<div title="fiszkuj">{ fiszkuj } + { fiszkującym }</div>
<div title="gorzej">{ gorzeją } + { gorzejącym }</div>
<div title="grypsuj">{ grypsującej } + { grypsująca }</div>
<div title="hajluj">{ hajlując } + { hajlującymi }</div>
<div title="hukaj">{ hukając } + { hukającą }</div>
<div title="jonizuj">{ jonizuję } + { jonizującą }</div>
<div title="kiereszuj">{ kiereszującego } + { kiereszującymi }</div>
<div title="kiełbasz">{ kiełbaszących } + { kiełbasząca }</div>
<div title="kiksuj">{ kiksujemy } + { kiksującemu }</div>
<div title="kitwasz">{ kitwaszących } + { kitwaszącemu }</div>
<div title="koliduj">{ kolidującej } + { kolidująca }</div>
<div title="konweniuj">{ konweniującej } + { konweniujący }</div>
<div title="konwertuj">{ konwertujących } + { konwertującymi }</div>
<div title="kooptuj">{ kooptuję } + { kooptującymi }</div>
<div title="kopsaj">{ kopsające } + { kopsającą }</div>
<div title="kreduj">{ kredującego } + { kredujący }</div>
<div title="kłóc">{ kłócące } + { kłócącą }</div>
<div title="lansuj">{ lansująca } + { lansującym }</div>
<div title="latynizuj">{ latynizujące } + { latynizującą }</div>
<div title="logarytmuj">{ logarytmującego } + { logarytmujący }</div>
<div title="lśni">{ lśniące } + { lśniąca }</div>
<div title="manipuluj">{ manipulujemy } + { manipulującym }</div>
<div title="matuj">{ matujące } + { matującemu }</div>
<div title="miewaj">{ miewającej } + { miewającemu }</div>
<div title="modrz">{ modrzej } + { modrzący }</div>
<div title="męcz">{ męczenie męczące } + { męczącemu }</div>
<div title="nacieraj">{ nacierając } + { nacierającymi }</div>
<div title="nacieszaj">{ nacieszającego } + { nacieszającymi }</div>
<div title="nadarzaj">{ nadarzają } + { nadarzającymi }</div>
<div title="nagl">{ nagleniu } + { naglącą }</div>
<div title="nakreślaj">{ nakreślająca } + { nakreślającymi }</div>
<div title="nakręcaj">{ nakręcających } + { nakręcającymi }</div>
<div title="namieraj">{ namierające } + { namierającymi }</div>
<div title="niebajcz">{ niebajczące } + { niebajczącemu }</div>
<div title="niebastuj">{ niebastującą } + { niebastującym }</div>
<div title="niebieszcz">{ niebieszczycie } + { niebieszczącymi }</div>
<div title="niecenzuruj">{ niecenzurujące } + { niecenzurującą }</div>
<div title="niechybotaj">{ niechybotającej } + { niechybotająca }</div>
<div title="nieczyszcz">{ nieczyszczące } + { nieczyszczący }</div>
<div title="niedołaż">{ niedołażenie } + { niedołażącemu }</div>
<div title="niedruzgocz">{ niedruzgoczącej } + { niedruzgocząca }</div>
<div title="niedziej">{ niedziejące } + { niedziejącym }</div>
<div title="niedziob">{ niedziobanie } + { niedziobiący }</div>
<div title="niefederuj">{ niefederujący } + { niefederującemu }</div>
<div title="niegdacz">{ niegdaczącego niegdaczących } + { niegdaczącymi }</div>
<div title="niegliwiej">{ niegliwiejącą } + { niegliwiejącym }</div>
<div title="niegoogluj">{ niegooglujące } + { niegooglującym }</div>
<div title="niegromadz">{ niegromadzącego } + { niegromadzącymi }</div>
<div title="niegruberuj">{ niegruberującej } + { niegruberującą }</div>
<div title="niegrupuj">{ niegrupującą } + { niegrupującymi }</div>
<div title="niegrępluj">{ niegręplujący } + { niegręplującymi }</div>
<div title="niejełczej">{ niejełczejącemu } + { niejełczejącym }</div>
<div title="niekolęduj">{ niekolędującej } + { niekolędujący }</div>
<div title="niekołacz">{ niekołaczące } + { niekołaczący }</div>
<div title="niekreśl">{ niekreśleniu } + { niekreśląca }</div>
<div title="nielakuj">{ nielakującą } + { nielakującymi }</div>
<div title="niemiej">{ niemieją } + { niemiejącą }</div>
<div title="nienadwoż">{ nienadwożące } + { nienadwożącym }</div>
<div title="nienażeraj">{ nienażerających } + { nienażerającymi }</div>
<div title="nieoblatuj">{ nieoblatującego } + { nieoblatująca }</div>
<div title="nieociekaj">{ nieociekających } + { nieociekającą } + { nieociekającymi }</div>
<div title="nieodkopuj">{ nieodkopującego } + { nieodkopująca }</div>
<div title="nieodtłukuj">{ nieodtłukujący } + { nieodtłukującym }</div>
<div title="nieomotuj">{ nieomotującemu } + { nieomotującym }</div>
<div title="nieopłacaj">{ nieopłacającą } + { nieopłacającym }</div>
<div title="nieoszpecaj">{ nieoszpecające } + { nieoszpecającą }</div>
<div title="niepogdakuj">{ niepogdakujące } + { niepogdakująca }</div>
<div title="niepolegaj">{ niepolegająca } + { niepolegającymi }</div>
<div title="niereaguj">{ niereagującego } + { niereagująca }</div>
<div title="nierzeźb">{ nierzeźbiącej } + { nierzeźbiącą }</div>
<div title="niesiadaj">{ niesiadającego } + { niesiadający }</div>
<div title="niesmarcz">{ niesmarczącego } + { niesmarczącemu }</div>
<div title="niestolarz">{ niestolarzące } + { niestolarzący }</div>
<div title="nieszczep">{ nieszczepiących } + { nieszczepiącym }</div>
<div title="nietaksuj">{ nietaksującej } + { nietaksującemu }</div>
<div title="nietygluj">{ nietyglujących } + { nietyglującemu }</div>
<div title="niewierc">{ niewiercący } + { niewiercącym }</div>
<div title="niewlok">{ niewlokącego } + { niewlokącemu }</div>
<div title="niewrzucaj">{ niewrzucającą } + { niewrzucającemu }</div>
<div title="niewydaj">{ niewydające } + { niewydającemu }</div>
<div title="niewydechuj">{ niewydechujące } + { niewydechująca }</div>
<div title="niewyglądaj">{ niewyglądający } + { niewyglądającym }</div>
<div title="niewłamuj">{ niewłamujący } + { niewłamującymi }</div>
<div title="nieznaj">{ nieznające } + { nieznająca }</div>
<div title="nieślimacz">{ nieślimaczą } + { nieślimaczącemu }</div>
<div title="obciosuj">{ obciosujemy } + { obciosująca }</div>
<div title="obdzwaniaj">{ obdzwaniajcie } + { obdzwaniający }</div>
<div title="obojętniej">{ obojętniejcie } + { obojętniejąca }</div>
<div title="odchrzaniaj">{ odchrzaniającej } + { odchrzaniającym }</div>
<div title="odczerpuj">{ odczerpującemu } + { odczerpującymi }</div>
<div title="odejmuj">{ odejmującej } + { odejmującą }</div>
<div title="odgartuj">{ odgartują } + { odgartującymi }</div>
<div title="odmrażaj">{ odmrażając odmrażającej } + { odmrażającemu }</div>
<div title="odmładzaj">{ odmładzającą } + { odmładzającemu }</div>
<div title="odpieprzaj">{ odpieprzającego } + { odpieprzająca }</div>
<div title="odpylaj">{ odpylajcie } + { odpylającą }</div>
<div title="odszczurzaj">{ odszczurzającą } + { odszczurzającym }</div>
<div title="odtykaj">{ odtykający } + { odtykającemu }</div>
<div title="odwzorcowuj">{ odwzorcowujcie } + { odwzorcowujący }</div>
<div title="opiniuj">{ opiniujący } + { opiniującymi }</div>
<div title="opluwaj">{ opluwając } + { opluwająca }</div>
<div title="osrebrzaj">{ osrebrzający } + { osrebrzającemu }</div>
<div title="otruwaj">{ otruwające } + { otruwający }</div>
<div title="perl">{ perliczek } + { perlącą }</div>
<div title="piecuchuj">{ piecuchujemy piecuchującego } + { piecuchującemu }</div>
<div title="piek">{ piekąc piekącego } + { piekącemu }</div>
<div title="pilotuj">{ pilotuj } + { pilotujący }</div>
<div title="plew">{ plewi } + { plewiącą }</div>
<div title="pocieszaj">{ pocieszające } + { pocieszającą }</div>
<div title="podróżuj">{ podróżujemy podróżując } + { podróżującemu }</div>
<div title="postuj">{ postuj } + { postująca postującą }</div>
<div title="poszczękuj">{ poszczękujące } + { poszczękującą }</div>
<div title="potworniej">{ potworniejąc potworniejącego } + { potworniejącą }</div>
<div title="powzdychuj">{ powzdychujące } + { powzdychująca }</div>
<div title="powłócz">{ powłóczcie } + { powłóczącym }</div>
<div title="prefabrykuj">{ prefabrykujemy } + { prefabrykującą }</div>
<div title="prowokuj">{ prowokujemy prowokują } + { prowokująca }</div>
<div title="przechorowuj">{ przechorowujesz } + { przechorowującą }</div>
<div title="przenosz">{ przenoszą } + { przenoszącemu }</div>
<div title="przepijaj">{ przepijając } + { przepijająca }</div>
<div title="przesiewaj">{ przesiewającego przesiewających } + { przesiewający }</div>
<div title="przesmarowuj">{ przesmarowujcie } + { przesmarowującą }</div>
<div title="prześwięcaj">{ prześwięcając } + { prześwięcający }</div>
<div title="przybielaj">{ przybielając przybielające } + { przybielającym }</div>
<div title="przymykaj">{ przymykające } + { przymykającym }</div>
<div title="przypakowuj">{ przypakowuj } + { przypakowującym }</div>
<div title="przywabiaj">{ przywabiają } + { przywabiającymi }</div>
<div title="pykaj">{ pykającemu } + { pykającym }</div>
<div title="rekognoskuj">{ rekognoskujesz } + { rekognoskująca }</div>
<div title="reprywatyzuj">{ reprywatyzuj } + { reprywatyzująca }</div>
<div title="rezygnuj">{ rezygnują } + { rezygnująca }</div>
<div title="rozbijaj">{ rozbijają } + { rozbijającą }</div>
<div title="rozgotowuj">{ rozgotowujesz } + { rozgotowującym }</div>
<div title="rozjadaj">{ rozjadające } + { rozjadającymi }</div>
<div title="rozkaszliwaj">{ rozkaszliwajcie } + { rozkaszliwająca }</div>
<div title="rozlatuj">{ rozlatują } + { rozlatującemu }</div>
<div title="rozmiękczaj">{ rozmiękczającej } + { rozmiękczającą }</div>
<div title="rozplotkowuj">{ rozplotkowujące } + { rozplotkowujący }</div>
<div title="rozprawiaj">{ rozprawiającą } + { rozprawiającymi }</div>
<div title="rozpławiaj">{ rozpławiających } + { rozpławiającą }</div>
<div title="rozsiewaj">{ rozsiewających } + { rozsiewająca }</div>
<div title="rykoszetuj">{ rykoszetujesz } + { rykoszetująca }</div>
<div title="samonapędzaj">{ samonapędzają } + { samonapędzający }</div>
<div title="schrzaniaj">{ schrzaniajcie } + { schrzaniającym }</div>
<div title="siorbaj">{ siorbającej } + { siorbający }</div>
<div title="skrzepiaj">{ skrzepiający } + { skrzepiającemu }</div>
<div title="solidaryzuj">{ solidaryzuj } + { solidaryzującym }</div>
<div title="sołtysuj">{ sołtysujemy } + { sołtysujący }</div>
<div title="spekuluj">{ spekulującemu } + { spekulującym }</div>
<div title="spotwarzaj">{ spotwarzającą } + { spotwarzającym }</div>
<div title="sroż">{ srożą } + { srożącą }</div>
<div title="startuj">{ startujemy startując } + { startujący }</div>
<div title="studz">{ studzi } + { studzącą }</div>
<div title="stulaj">{ stulających } + { stulającym }</div>
<div title="sumuj">{ sumując } + { sumującą }</div>
<div title="szafarz">{ szafarzyłem } + { szafarzącym }</div>
<div title="szepcz">{ szepczę } + { szepczący }</div>
<div title="szurfuj">{ szurfuj } + { szurfującym }</div>
<div title="szwindluj">{ szwindlująca } + { szwindlującym }</div>
<div title="szypruj">{ szyprującemu } + { szyprującym }</div>
<div title="teleportuj">{ teleportuj } + { teleportujący }</div>
<div title="traumatyzuj">{ traumatyzującej } + { traumatyzującą }</div>
<div title="trzeźw">{ trzeźwieć trzeźwił } + { trzeźwiącą }</div>
<div title="tumaniej">{ tumaniejcie } + { tumaniejącą }</div>
<div title="uplątuj">{ uplątują } + { uplątującymi }</div>
<div title="upuszczaj">{ upuszczającemu } + { upuszczającymi }</div>
<div title="uładzaj">{ uładzające } + { uładzającemu }</div>
<div title="uświęcaj">{ uświęcając } + { uświęcającymi }</div>
<div title="warchol">{ warchol } + { warcholąca }</div>
<div title="wgrzebuj">{ wgrzebującemu } + { wgrzebującym }</div>
<div title="wiedz">{ wiedz wiedziemy } + { wiedzącymi }</div>
<div title="wierz">{ wierzej } + { wierzącemu }</div>
<div title="winyluj">{ winylujące winylującej winylujących } + { winylującymi }</div>
<div title="wiórkuj">{ wiórkującego } + { wiórkująca }</div>
<div title="wtachuj">{ wtachującego } + { wtachującymi }</div>
<div title="wygalaj">{ wygalające } + { wygalającą }</div>
<div title="wyglądaj">{ wyglądającą } + { wyglądającym }</div>
<div title="wygrzebuj">{ wygrzebujcie } + { wygrzebująca }</div>
<div title="wyiskrzaj">{ wyiskrzającą } + { wyiskrzającemu }</div>
<div title="wynurzaj">{ wynurzającej } + { wynurzającymi }</div>
<div title="wyokrętowuj">{ wyokrętowujesz } + { wyokrętowującym }</div>
<div title="wysmarkuj">{ wysmarkujących } + { wysmarkującemu }</div>
<div title="wózkuj">{ wózkują } + { wózkujący }</div>
<div title="węgl">{ węglowy węglących } + { węglący }</div>
<div title="zabijaj">{ zabijają zabijając } + { zabijający }</div>
<div title="zadręczaj">{ zadręczajcie } + { zadręczający }</div>
<div title="zaklinowuj">{ zaklinowujcie } + { zaklinowującymi }</div>
<div title="zażeraj">{ zażerającego } + { zażerającą }</div>
<div title="zbudzaj">{ zbudzającą } + { zbudzającemu }</div>
<div title="zgonuj">{ zgonującemu } + { zgonującym }</div>
<div title="zmazuj">{ zmazując } + { zmazująca }</div>
<div title="znacz">{ znaczącą } + { znaczącym }</div>
<div title="znikaj">{ znikająca } + { znikającemu }</div>
<div title="zrzędz">{ zrzędzeniu } + { zrzędzącą }</div>
<div title="zwieszaj">{ zwieszajcie } + { zwieszającą } + { zwieszającemu }</div>
</div>
</ul>

algorithms/polish.sbl
mitya57: > Are there other Polish words (probably also from Latin) which follow a similar pattern?

Yes, [many](https://en.wiktionary.org/wiki/Category:Polish_terms_suffixed_with_-ia) Latin borrowings have this ending. However, the algorithm currently removes all vowel endings as the last step, so it is not easy to change it.

Given that the number of false positives should be quite low, I propose to leave it as is.

mitya57:
I pushed a new version that hopefully addresses all your comments and all my own TODO comments.

I had a temptation to rewrite the stemmer to avoid repeating adjective-like endings in three places (for adjectives, superlatives and participles) and even started doing that in mitya57/snowball@0e54859befaffddb591b5baca5961958b3e1c216, but those changes were too disruptive, so in the end I took a more conservative approach which was extending the current algorithm.

mitya57:
And a question. Currently we have logic like:
```
(
  $(len > 6)
  // remove 4 characters
) or (
  $(len > 5)
  // remove 3 characters
)
```

But what this code really tries to check is that after removing the endings, the resulting stem is not too short (e.g. there are at least 2 characters left). What is the most elegant way to achieve this?

Other Slavic stemmers try to operate in terms of syllables rather than letters, maybe we should try to do the same here?

ojwb:
> But what this code really tries to check is that after removing the endings, the resulting stem is not too short (e.g. there are at least 2 characters left). What is the most elegant way to achieve this?

Most of the other stemmers set up regions in advance and then test those.  Usually these are based on syllables, but a length based version would look like:

```
$p1 = limit
do (hop 2 setmark p1)
```

where `p1` is an integer variable:

```
integers (p1)
```

and (usually) declare a helper routine to test for the region:

```
define R1 as $p1 <= cursor
```

Then there's a choice for how you apply it - either:

```
[substring] R1 among (
  //...
)
```

or:

```
setlimit tomark p1 for ([substring]) among (
  //...
)
```

The difference is that the first finds the longest matching suffix from those in the `among` and then checks if it is in R1, while the second finds the longest matching suffix in R1 from those in the `among`.

E.g. the English stemmer's `Step_2` contains suffixes `fulli` and `li` (note that `-y` has been changed to `-i` before this setp) so that gives stems like so with the current `[substring] R1 among`:

```
fully                         fulli
```

`fulli` is the longest suffix, but it's not entirely in R1 so we don't remove it.

If that used `setlimit tomark p1 for ([substring])` instead we'd get:

```
fully                         ful
```

Due to the limit, the string the among actually tests against is `lli` so suffix `li` is the longest.

I don't really have any useful insights as to which to use when - I just end up trying both for each situation and seeing which works better.

> Other Slavic stemmers try to operate in terms of syllables rather than letters, maybe we should try to do the same here?

Most of the stemmers work in terms of syllables.  It does seem a better approach, though the difference it makes will depends on the language - if most words alternate vowel and consonant then it'll not make as much difference.

Most languages use the same R1 definition as `porter.sbl`, or the slight variant in `german.sbl` which imposes a minimum of 3 characters before R1.  There are also various other minor and major variations though: `english.sbl` has an exception list; `czech.sbl` (not merged yet) takes syllabic consonants into account; the Kraaij-Pohlmann dutch stemmer treats the `ij` digraph as a vowel rather than a vowel followed by a consonant; Hungarian has totally different R1 definition; Indonesian instead just counts the number of vowels; in Serbian for words that starts with two or more vowels, R1 starts after those vowels but there's also some special handling of words containing `r`; Yiddish is "Either 3 consonants or the first non-vowel after a vowel".

One of the first two is probably a good starting point unless there's something that Polish has in common with one of the other languages that's relevant here.  The `stemmer-compare` script allows comparing the results.

I notice though that the difference between the length threshold and the suffix length is not the same for all suffixes - it looks like it can be 1, 2 or 3.  That may essentially be a bug, or actually make not difference in practice.  The 1 vs 3 could be equivalent to R1 vs R2 (R2 is another syllable after R1).

(Incidentally, `len` in Snowball is O(n) when working in UTF-8.  Here `n` is the length of a word so not large, but it's a bit silly to repeatedly call `len` like the code currently does.  This was something I thought best to leave for now as it's really just an optimisation; also the Snowball compiler could probably be smarter and reuse the calculated length when it can see the string hasn't changed...)

mitya57:
Updated once again. This time I also tried to solve problem that the original stemmer removed some d-suffix but did not support all possible i-suffices for it. In particular, I added possible declension forms for diminutives and for -owy adjectives.

Also, I added support for more verbal forms and made the list of verbal forms better structured (e.g. all past forms are grouped together).

The corresponding data changes are in snowballstem/snowball-data#30, each one in a separate commit to make it easier to review.

algorithms/polish.sbl
```diff
@@ -0,0 +1,240 @@
...
+  )
+)
+
+backwardmode (
+  define remove_nouns as (
+    setlimit tomark p1 for ([substring]) among (
+      'ach'  // plural locative
+      'ami'  // plural instrumental
+      'nia'  // genitive of words ending with -nie, e.g. czytanie - czytania
+      'niu'  // dative/locative of words ending with -nie, e.g. czytanie - czytaniu
```
ojwb, line 56: _czytaniu_ would actually be handled by `'aniu'` on line 62.

algorithms/polish.sbl
```diff
@@ -0,0 +1,240 @@
...
+      'cia'  // genitive of words ending with -cie, e.g. ukrycie - ukrycia
+      'ciu'  // dative/locative of words ending with -cie, e.g. ukrycie - ukryciu
+      'ce'   // dative/locative of words ending with -ka, e.g. muzyka - muzyce
```
ojwb, line 59: I notice _muzyka_ stems to _muzyk_ while _muzyce_ stems to _muz_ - not necessary a problem, but if it's unexpected or more widespread this might be worth looking at.

algorithms/polish.sbl
```diff
@@ -0,0 +1,240 @@
...
+      '{ak}c'      // contemporary adverbial participle (transgressive), e.g. czytając
+      // different forms of active adjectival participle, e.g. czytający (one who is reading)
+      '{ak}cy'     // nominative singular (masculine)
+      '{ak}cego'   // genitive singular (masculine/neuter)
+      '{ak}cemu'   // dative singular (masculine)
+      '{ak}cym'    // instrumental/locative singular (masculine)
+      '{ak}ca'     // nominative singular (feminine)
+      '{ak}cej'    // genitive/dative/locative singular (feminine)
+      '{ak}c{ak}'  // accusative singular (feminine)
+      '{ak}ce'     // nominative/accusative singular (neuter)
```
ojwb, line 158: _czytające_ will actually get _-ce_ removed by `remove_nouns` then `-ą` removed by `remove_general_ends`, rather than by this rule.  So this may be redundant, but perhaps _-ące_ can be exposed by removal of an ending in a previous step.

algorithms/polish.sbl
```diff
@@ -0,0 +1,240 @@
...
+      'ali{s'}my'     'ili{s'}my'     // past 1st person plural virile, e.g. czytaliśmy, chodziliśmy
+      'a{l/}y{s'}my'  'i{l/}y{s'}my'  // past 1st person plural nonvirile, e.g. czytałyśmy, chodziłyśmy
+      'ali{s'}cie'    'ili{s'}cie'    // past 2nd person plural virile, e.g. czytaliście, chodziliście
+      'a{l/}y{s'}cie' 'i{l/}y{s'}cie' // past 2nd person plural nonvirile, e.g. czytałyście, chodziłyście
+      'ali'           'ili'           // past 3rd person plural virile, e.g. czytali, chodzili
+      'a{l/}y'        'i{l/}y'        // past 3rd person plural nonvirile, e.g. czytały, chodziły
+      '{ak}c'      // contemporary adverbial participle (transgressive), e.g. czytając
+      // different forms of active adjectival participle, e.g. czytający (one who is reading)
+      '{ak}cy'     // nominative singular (masculine)
+      '{ak}cego'   // genitive singular (masculine/neuter)
```
ojwb, line 152: _czytającego_ will actually get _-ego_ removed by `remove_adjective_ends` then `-ąc` removed by the rule in this `among` (line 149).

algorithms/polish.sbl
```diff
@@ -0,0 +1,240 @@
...
+      '{ak}cemu'   // dative singular (masculine)
```
ojwb, line 153: Very similar to the _-ego_ case just above.

algorithms/polish.sbl
```diff
@@ -0,0 +1,240 @@
...
+      '{ak}cym'    // instrumental/locative singular (masculine)
+      '{ak}ca'     // nominative singular (feminine)
+      '{ak}cej'    // genitive/dative/locative singular (feminine)
```
ojwb, line 156: Also similar to the _-ego_ case.

algorithms/polish.sbl
```diff
@@ -0,0 +1,240 @@
...
+      '{ak}c{ak}'  // accusative singular (feminine)
+      '{ak}ce'     // nominative/accusative singular (neuter)
+      '{ak}cych'   // genitive plural
```
ojwb, line 159: Similar to the _-ego_ case.

algorithms/polish.sbl
```diff
@@ -0,0 +1,240 @@
...
+      '{ak}cymi'   // instrumental plural
+        (delete)
+      'aj'     // imperative, e.g. czytaj
+         (<- 'a')
```
ojwb, line 163: _czytając_ (example given further above) stems to _czytaj_, but _czytaj_ gets the _-j_ removed here and then _-a_ removed by `remove_general_ends` to give stem _czyt_.

They both seem to be forms of _czytać_ and that also stems to _czyt_ which kind of suggests maybe we ought to also be removing the _aj_ from _czytając_?

Maybe that causes other problems though.

algorithms/polish.sbl
```diff
@@ -0,0 +1,240 @@
...
+      'rze'  // e.g. dobrze
+        (<- 'r')
+    )
+  )
+
+  define remove_plural_forms as (
+    setlimit tomark p1 for ([substring]) among (
+      '{o'}w'  // genitive plural, e.g. włos - włosów
+      'om'     // dative plural, e.g. włos - włosom
+      'ami'    // instrumental plural, e.g. włos - włosami
```
ojwb, line 182: _-ami_ actually gets removed from _włosami_ by `remove_nouns` (noted as "plural instrumental" there).  Maybe this is needed if _-ami_ can be exposed by removing another suffix but we also need to remove _-ami_ earlier to expose suffixes, but if so that kind of suggests the order of rules might not be quite right.

mitya57:
Thank you for the comments! Please don't wait for me if you want to get a new release out, I will fix them eventually.

As I will need to rework it again anyway, I want to ask some questions myself first:

1. Currently we call all `remove_*` functions with `do` syntax, which means they are all called subsequently. However this does not always make sense: e.g. if we removed noun endings, it does not make much sense to remove verb or adjective endings, since a word cannot be noun and verb/adjective at the same time. The only scenarios where this may make sense are:
    - when a word is an adjective, e.g. *czytającego*, we may first remove an adjective ending *-ego* from it, but after that also remove adjective-specific preceding part (*-ąc-* or *-ając-*).
    - in the end, maybe it makes sense to cleanup all ending vowels, to make sure that the algorithm consistently produces stems ending with a consonant.

    Maybe I should restructure the `stem` function to use `or` instead of `do`?

2. As I mentioned above, the original algorithm that I took as a base did not handle all inflectional suffixes well, but at the same time tried to work with derivational suffixes (*-an-* and *-acj-* in nouns, diminutive suffixes, *-ow-* in adjectives). Maybe we should focus on one job (inflectional suffixes) and do it well? This will reduce the number of possible combinations and will make the code cleaner.

3. In my first question I mentioned removing ending vowels. But do we really want to do that as the unconditional last step? If a vowel is at end of the word then sure, it should be removed. But for example if we take the verb *czyt**a**ć*, then all forms of it contain the *a* that I highlighted. Maybe we should stem it to *czyta* in this case?

ojwb:
Sorry, I got rather swallowed up trying to make progress towards a new release.

> Thank you for the comments! Please don't wait for me if you want to get a new release out, I will fix them eventually.

Thanks - I am currently leaning towards getting a new release out (it's very nearly 3.5 years since the last one which is too long and there's a huge pile of unreleased useful changes), and then dealing with the pending new algorithms and making another release with those.  As well as this there's Czech (#151) and Persian (#194) waiting for review and Ukrainian (#144) waiting for the submitter.

> As I will need to rework it again anyway, I want to ask some questions myself first:
> 
>     1. Currently we call all `remove_*` functions with `do` syntax, which means they are all called subsequently. However this does not always make sense: e.g. if we removed noun endings, it does not make much sense to remove verb or adjective endings, since a word cannot be noun and verb/adjective at the same time. The only scenarios where this may make sense are:
>        
>        * when a word is an adjective, e.g. _czytającego_, we may first remove an adjective ending _-ego_ from it, but after that also remove participle-specific preceding part (_-ąc-_ or _-jąc-_).
>        * in the end, maybe it makes sense to cleanup all ending vowels, to make sure that the algorithm consistently produces stems ending with a consonant.
>        
>        Maybe I should restructure the `stem` function to use `or` instead of `do`?

If there's a clean structure to the suffixes in the language then reflecting that in the code makes sense, and eliminating checks for suffixes that can never be present seems a good idea - more efficient and eliminates the risk of removing something that looks like a suffix when it isn't.

However you may find there are cases the suffix is removed but it's just how a word ends, and removing other suffixes incorrectly actually helps to conflate forms of a word together, albeit on a stem which isn't linguistically correct (which is OK for our purposes provided it doesn't collide with stems of unrelated words).  This is quite common in English (which is really a mess of a language with a lot of loan words and the like from many other languages), but I've seen it in several others too.  I saw a good example of this just the other day in an old mailing list post, but I can't now find it or think of a good example myself.

>     2. As I mentioned above, the original algorithm that I took as a base did not handle all inflectional suffixes well, but at the same time tried to work with derivational suffixes (_-an-_ and _-acj-_ in nouns, diminutive suffixes, _-ow-_ in adjectives). Maybe we should focus on one job (inflectional suffixes) and do it well? This will reduce the number of possible combinations and will make the code cleaner.

The key goal is to improve retrieval (without harming precision too much), so inflectional suffixes are probably the most critical.

Diminutive suffices can change meaning enough that conflating or not are both reasonable.  The linguistic structure may steer one to a conclusion which isn't fully supported by the meanings of the words - E.g. in English, a young dog is a "puppy" and not conflating seems the obvious choice, but in German a young "Hund" is a "Hündchen" - the structure of the words perhaps suggests conflating.  The meanings are probably similarly different in the two languages (I'm not fluent enough in German to say for sure) and the difference is really that in English you'd need a special case, while in German you could conceivably have a _-chen_ rule that also handled "Kätzchen", etc.

_-acj-_ is like English "ation", right?  That seems more useful than diminutives at least.

Not sure what _-an-_ and _-ow-_ mean.

>     3. In my first question I mentioned removing ending vowels. But do we really want to do that as the unconditional last step? If a vowel is at end of the word then sure, it should be removed. But for example if we take the verb _czyt**a**ć_, then all forms of it contain the _a_ that I highlighted. Maybe we should stem it to _czyta_ in this case?

Returning _czyt_ as the stem is fine for our purposes, provided it doesn't result in conflating unrelated words.  Thinking about English words ending "e", e.g. "pasting` comes from "paste" which has an entirely different meaning to "past".  The stem for all of these was "past", but I've added an exception for the next release to address this.  The English stemmer goes to some trouble to try to remove or re-add _-e_ but was still getting this case wrong.

mitya57:
Thank you for the detailed response!
I also had little time in the last few weeks, but I hope I will find some time in May to make a new version of this PR.

PR #221, opened 2025-03-17
Add ES6 Support with Classes and Closures for JavaScript

titusng0110:
# Description:
This pull request introduces ES6 support for JavaScript in the Snowball project by utilizing modern JavaScript features like classes and closures. These changes are designed to minimize compatibility issues while supporting ECMAScript Module (ESM) importing.

# Key Changes:

- Build System: Updated the GNUmakefile to include tasks for building ES6 modules.
- Code Generation: Modified driver.c to handle a new JavaScript type parameter (es6), enabling conditional compilation for ES6 syntax.
- JavaScript Syntax: Enhanced generator_js.c to use let and const where applicable, and introduced ES6 classes for base and stemmer implementations.
- Closures and Classes: Utilized closures and classes in base-stemmer-es6.mjs and stemwords-es6.mjs to encapsulate functionality and maintain minimal changes in compatibility with existing systems.
- ESM Importing: Enabled ESM importing by structuring the JavaScript files to export default classes, ensuring seamless integration with modern JavaScript environments.

# Motivation:
The shift to ES6 syntax, supported by classes and closures, modernizes the codebase, providing better maintainability and compatibility with contemporary JavaScript environments. This transition facilitates easier integration into applications that utilize ESM.

# Testing:

- Confirmed that the build process works correctly with the updated GNUmakefile.
- Conducted tests with sample input files to ensure the ES6 modules produce consistent and correct output.
- Verified that the use of closures and classes does not disrupt existing functionality.
- Added CI for ES6. Passed CI


mitya57 (changes requested):

compiler/driver.c
```diff
@@ -168,15 +168,17 @@ static int read_options(struct options * o, int argc, char * argv[]) {
...
             if (startswith(s, "-js=")) {
                 o->make_lang = LANG_JAVASCRIPT;
                 if (eq(s + 4, "global")) {
-                    o->js_esm = false;
+                    o->js_version = 0;
                 } else if (eq(s + 4, "esm")) {
-                    o->js_esm = true;
+                    o->js_version = 1;
+                } else if (eq(s + 4, "es6")) {
+                    o->js_version = 2;
```
ojwb, line 181: Please can we use an enum instead of these magic numbers?

compiler/generator_js.c
```diff
@@ -88,7 +88,10 @@ static void write_savecursor(struct generator * g, struct node * p,
     g->B[0] = str_data(savevar);
     g->S[1] = "";
     if (p->mode != m_forward) g->S[1] = "base.limit - ";
-    writef(g, "~Mvar /** number */ ~B0 = ~S1base.cursor;~N", p);
+    if(g->options->js_version == 2)
+        writef(g, "~Mlet /** number */ ~B0 = ~S1base.cursor;~N", p);
+    else
+        writef(g, "~Mvar /** number */ ~B0 = ~S1base.cursor;~N", p);
```
ojwb, line 94: We could avoid a lot of this repetitive conditionalisation if we add a new format sequence which expands to either `let` or `var`.  I'd suggest `~d` as that's mnemonic (for declaration) and not used anywhere already.

compiler/generator_js.c
```diff
@@ -1263,7 +1291,13 @@ static void generate(struct generator * g, struct node * p) {
...
 
 static void generate_class_begin(struct generator * g) {
-    if (g->options->js_esm) {
+    if (g->options->js_version == 2) {
+        w(g, "import ~P from './base-stemmer-es6.mjs'~N"
+             "~N"
+             "const ~n = (() => {~+~N"
+             "~Mconst base = new ~P();~N");
+    }
+    else if (g->options->js_version == 1) {
```
ojwb, line 1300: `} else if` is the style used by the existing code.

compiler/generator_js.c
```diff
@@ -1283,14 +1317,26 @@ static void generate_class_begin(struct generator * g) {
 
 static void generate_class_end(struct generator * g) {
     w(g, "~N");
+    if(g->options->js_version == 2)
```
ojwb, line 1320: `if (` is the style used by the existing code.

GNUmakefile
```diff
@@ -33,7 +33,7 @@ python_sample_dir = sample
 js_output_dir = js_out
 js_runtime_dir = javascript
 js_sample_dir = sample
-NODE ?= nodejs
+NODE ?= node
```
ojwb, line 36: Maybe a worthwhile change, but seems unconnected with added ES6 support.

javascript/base-stemmer-es6.mjs
```diff
@@ -0,0 +1,264 @@
+class BaseStemmer {
```
ojwb, line 1: There's a lot of code duplicated here.  Is there not some way to have a single version which works for both?

If not, please minimise differences - e.g. currently the diff contains some arbitrary reformatting which will make future maintenance unnecessarily harder.  This sort of thing:

```
         if (this.limit - this.cursor < s.length) return false;
-        if (this.current.slice(this.cursor, this.cursor + s.length) != s) {
+        if (this.current.slice(this.cursor, this.cursor + s.length) != s)
+        {
             return false;
         }
```

Also all the comments seem to have been stripped which will similarly make it needlessly harder to apply changes from one version to the other.

javascript/base-stemmer-es6.mjs
```diff
@@ -0,0 +1,264 @@
...
+        }
+        return result;
+    }
+
+    assign_to() {
+        return this.current.slice(0, this.limit);
+    }
+}
+
+export default BaseStemmer;
```
ojwb, line 264: Please ensure there's a newline at the end of new text files.

javascript/stemwords-es6.mjs
```diff
@@ -0,0 +1,129 @@
+import fs from 'fs';
```
ojwb, line 1: It would be more helpful for ongoing maintenance if this was written such that the parts which are effectively the same as the existing version were actually the same.  (I don't mind if we adjust the existing code if that makes that easier, so long as it doesn't stop it working in situations where it currently works.)

Even ignoring whitespace, the current diff looks more like diffing two unrelated files.

ojwb:
Sorry, looks like CI is failing due to my recent change to stop `evening` stemming to `even` in English - it's using the latest snowball-data but it appears without the corresponding code change.  I thought GHA for a PR ran on the merge of that branch and the branch the PR was targetted at but it seems not.

I think you probably need to rebase your branch onto current upstream master and force-push (can explain if you don't know what that means).

compiler/generator_js.c
ojwb: Sketching that out:

`writef()` would gain an extra `case` something like:

```
            case 'd':
                if (g->options->js_version == JS_ES6)
                    write_string(g, "let");
                else
                    write_string(g, "var");
                continue;
```

and usage would look like:

```
    writef(g, "~M~d /** number */ ~B0 = ~S1base.cursor;~N", p);
```

ojwb: I have implemented this on master and merged to the branch (using `~D` for `let`/`var`; also `~C` for `const`/`/** @const */ var`).

compiler/generator_js.c
ojwb: Pushed a fixed.

compiler/generator_js.c
ojwb: Fixed.

ojwb:
> I think you probably need to rebase your branch onto current upstream master and force-push

To try to make progress I have merged master to the branch and pushed and CI has at least now run and passed, though there's still work to do here.  (I went for a merge rather than rebasing as it seemed wrong to rewrite the history for the master branch of someone else's fork.)

javascript/base-stemmer-es6.mjs
ojwb: An illustration of the problems of duplicating this code, especially with additional unnecessary modifications, is that merging from the master branch broke `-js=es6` because it lacks new runtime methods to support the `goto`/`gopast` grouping optimisation which is now implemented for Javascript.  For now I've just hacked this optimisation off in this PR branch (done by b23a3b094cb2632779d120d790bcb7007922203c) but this needs resolving.

I think the ideal for on-going maintenance would be to generate `javascript/base-stemmer-es6.mjs` from `javascript/base-stemmer.js` (or vice versa).  I don't know if there's Javascript tooling to support something liek this, but if not we could probably do it with `sed` or a `perl` script.

Failing that, if the two versions have only differences which are actually required, then at least loading them into an editor in `diff` mode would mean copying over changes was less painful and less error-prone.

compiler/driver.c
ojwb: Addressed.

GNUmakefile
ojwb: I've changed this in master (and renamed the variable to `JSRUN` since it ought to be possible to get this to work with e.g. deno).

compiler/generator_js.c
```diff
@@ -1278,39 +1278,71 @@ static void generate(struct generator * g, struct node * p) {
...
+        case JS_ESM:
+            w(g, "~M/**@return{string}*/~N");
+            w(g, "~Mthis['stemWord'] = function(/**string*/word) {~+~N");
+            w(g, "~Mbase.setCurrent(word);~N");
+            w(g, "~Mthis.stem();~N");
+            w(g, "~Mreturn base.getCurrent();~N");
+            w(g, "~-~M};~N");
+            w(g, "~-};~N");
+            w(g, "~N"
+                 "export default ~n~N");
```
ojwb, line 1331: The merged version doesn't use `export default`.  The upshot of this seems to be that dynamically loading the module and creating an object now requires generating both the module name and class name from the language name, whereas with a default export you would only need to generate the module name.

Maybe we should provide a standard way to create a stemmer object from a language name (which could allow creating from an ISO-639 code too, like you can with C libstemmer).  It still seems a bit clumsy if that has to generate both a module name and a class name though.

javascript/stemwords-es6.mjs
```diff
@@ -0,0 +1,129 @@
...
+
+The input file consists of a list of words to be stemmed, one per
+line. Words should be in lower case.
+
+If -c is given, the argument is the character encoding of the input
+and output files.  If it is omitted, the UTF-8 encoding is used.
+
+The output file consists of the stemmed words, one per line.
+
+-h displays this help`);
```
ojwb, line 15: I've peeled off this change to use a single `console.log()` call as it makes the source more readable.

The rest seems to be rewriting it to be async?

Not entirely convinced that's helpful - at least to me it makes it harder to follow so worse in it's partial role as an example of how to use the API, but maybe if you write a lot of JS async code it's actually easier to read in that form.

It doesn't seem to be measurably faster as async either (it can probably parallelise IO with stemming a bit, but there is presumably extra overhead from the parallelism too).

javascript/base-stemmer-es6.mjs
```diff
@@ -0,0 +1,264 @@
+class BaseStemmer {
+    constructor() {
```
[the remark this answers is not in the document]
ojwb: Now done by #239.

AA-Turner:
#240 has been merged, ought this remain open?

A

compiler/generator_js.c
ojwb: 5c37d044783231e5574c871edf861b73b56efc8f changed to using `export default` and makes creating a stemmer object easier as you only need to build the filename of the stemmer subclass, and not also its class name.

Creating from a text name and/or ISO-639 code is covered by #244.

Issue #229, opened 2025-05-08
KeyError: "Stemming algorithm 'porter' not found" on version 3.0.0 of the Python snowballstemmer package


laymonage:
The following code doesn't work after upgrading to 3.0.0:

```
import snowballstemmer
snowballstemmer.stemmer('porter')
```

with the following error:

```
  File "/path/to/venv/lib/python3.13/site-packages/snowballstemmer/__init__.py", line 27, in stemmer
    raise KeyError("Stemming algorithm '%s' not found" % lang)
KeyError: "Stemming algorithm 'porter' not found"
```

It worked fine with v2.2.0. Tested on Python 3.13 and Python 3.11.

I don't use snowballstemmer directly, but it's used by Sphinx's search functionality: https://github.com/sphinx-doc/sphinx/blob/c4929d026c8d22ba229b39cfc2250a9eb1476282/sphinx/search/en.py#L215

Not sure if this is a compatibility issue between Sphinx's usage of the library or a genuine bug in the v3.0.0 release.

Thanks!

Thorsten-Sick:
pydocstyle which uses snowballstemmer has similar issues. Those can also be reproduced by:

```
import snowballstemmer
stemmer = snowballstemmer.stemmer('english')
```

which raises the KeyError "Stemming algorithm 'english' not found"

AA-Turner:
cc @ojwb -- we've had a report of this in Sphinx, please let us know if we need to change our use of the API. I don't see anything obvious in [NEWS](https://github.com/snowballstem/snowball/blob/master/NEWS) about this, though.

A

amotl:
Hi. Our builds also have been tripped by this.
- https://app.readthedocs.org/projects/sarcam/builds/28103686/

ameraner:
We also see the same 'english' issue in https://github.com/ssec/sift/pull/448 https://github.com/ssec/sift/actions/runs/14903088149/job/41859433281?pr=448

andy-maier:
In addition to the missing "porter" algorithm, I'd like to point out that the issue is surfaced with a Python traceback:
```
Traceback
=========

    Traceback (most recent call last):
      File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/sphinx/cmd/build.py", line 432, in build_main
        app.build(args.force_all, args.filenames)
[...]

PR #230, opened 2025-05-08
chore: protect empty languages dict

henryiii:
See https://github.com/snowballstem/snowball/issues/229, 3.0.0 was released with an empty dict. This should help that show up.

### DIFF
--- a/python/create_init.py
+++ b/python/create_init.py
@@ -21,7 +21,8 @@ for pyscript in os.listdir(python_out_folder):
 imports.sort()
 languages.sort()
 
-assert languages
+if len(languages) == 0:
+    raise AssertionError('languages list is empty!')
 
 src = '''__all__ = ('language', 'stemmer')
 
--- a/python/create_init.py
+++ b/python/create_init.py
@@ -21,6 +21,8 @@ for pyscript in os.listdir(python_out_folder):
 imports.sort()
 languages.sort()
 
+assert languages
+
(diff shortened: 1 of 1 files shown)


python/create_init.py
```diff
@@ -21,6 +21,8 @@
 imports.sort()
 languages.sort()
 
+assert languages
```
AA-Turner, line 24: Minor change to be explicit on the length check, and use an explicit assertion (to avoid `-O` impact).
```suggestion
if len(languages) == 0:
    raise AssertionError('languages list is empty!')
```

AA-Turner (approved):

python/create_init.py
henryiii: I think `if not languages` reads well too; it's clearly saying that languages has to be something to pass. I considered avoiding -O, but this is a utility script so I thought that was unlikely to be present. 

I'll change to match but personally prefer `if not languages`.

AA-Turner: Happy with whatever you think best.

mitya57 (approved):

python/create_init.py
```diff
@@ -21,6 +21,9 @@
 imports.sort()
 languages.sort()
 
+if len(languages) == 0:
```
henryiii, line 24:
```suggestion
if not languages:
```

Here's the alternative if someone wants it.

PR #233, opened 2025-05-09
python: Reduce overhead of forwarding to PyStemmer

ojwb:
Only import the pure Python modules if we're going to use them.

Eliminate the cext_available flag variable and instead just assign the PyStemmer functions to their snowballstemmer names, or define the appropriate versions for the pure Python case.

### DIFF
--- a/python/create_init.py
+++ b/python/create_init.py
@@ -16,8 +16,8 @@ for pyscript in os.listdir(python_out_folder):
     if (match):
         langname = match.group(1)
         titlecase = re.sub(r"_", "", langname.title())
-        languages.append("    '%(lang)s': %(title)sStemmer," % {'lang': langname, 'title': titlecase})
-        imports.append('from .%(lang)s_stemmer import %(title)sStemmer' % {'lang': langname, 'title': titlecase})
+        languages.append("        '%(lang)s': %(title)sStemmer," % {'lang': langname, 'title': titlecase})
+        imports.append('    from .%(lang)s_stemmer import %(title)sStemmer' % {'lang': langname, 'title': titlecase})
 imports.sort()
 languages.sort()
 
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks for the review - I was going to ask your thoughts once I'd checked CI was green but you beat me to it!

python/create_init.py
```diff
@@ -26,31 +26,26 @@
...
-    if lang.lower() in _languages:
-        return _languages[lang.lower()]()
-    else:
-        raise KeyError("Stemming algorithm '%%s' not found" %% lang)
+    def stemmer(lang):
+        lang = lang.lower()
+        if lang in _languages:
+            return _languages[lang]()
+        else:
+            raise KeyError("Stemming algorithm '%%s' not found" %% lang)
```
AA-Turner, line 48: An alternate approach here could be using `importlib.import_module()` to only load the language requested. `languages` becomes `dict[str, str]` mapping the language name to the stemmer class, and then here you do

```python
try:
    mod = importlib.import_module(f'snowballstemmer.{lang}_stemmer')
    stemmer_class = getattr(mod, _languages[lang])
    return stemmer_class()
except (AttributeError, ImportError, KeyError):
    raise KeyError(f'Stemming algorithm {lang!r} not found') from None
```

ojwb: We could even get rid of `_languages` entirely if each stemmer module provided a common way to create a stemmer of its class - e.g. if I append `factory = EnglishStemmer` to `english_stemmer.py` and similarly for each subclass then I can do:

```
>>> import importlib
>>> mod = importlib.import_module('snowballstemmer.english_stemmer')
>>> stemmer = mod.factory()
>>> stemmer
<snowballstemmer.english_stemmer.EnglishStemmer object at 0x7f1f34b92270>
>>> stemmer.stemWord('stemming')
'stem'
```

There's probably a more pythonic way to do that but hopefully you get the idea.

Issue #183, opened 2023-08-29
JS/TS ESM version (modern JS module system vs global scope polution)


lionel-rowe:
The current JS build script outputs files with the following shape (example for English):

```js
StemmerEnglish = function() {
    // ...
}
```

In non-strict-mode (e.g. root-level non-modularized JavaScript), this automatically adds `StemmerEnglish` to the global scope, which is less-than-ideal but generally doesn't cause problems. However, in strict-mode (e.g. a JavaScript module), this throws `Uncaught ReferenceError: StemmerEnglish is not defined`. This makes it impossible to run on Deno, [which always uses strict mode](https://github.com/denoland/deno/issues/5845).

In addition, the JSDoc types added aren't picked up by TypeScript.

Is there interest in adding an ESM build, either as a replacement for or in addition to the current JS one? I've managed to hack one together in my fork ([README](https://github.com/clearlylocal/snowball/blob/esm/esm/README.md), [compilation logic](https://github.com/clearlylocal/snowball/blob/esm/esm/scripts/compile.ts), [dynamic imports by locale code](https://github.com/clearlylocal/snowball/blob/esm/esm/core/getStemmerByLocale.mjs), [tests](https://github.com/clearlylocal/snowball/blob/esm/esm/tests/stemmer.test.ts)), and it works nicely in Deno, newer versions of Node, and in-browser HTTP imports, with TypeScript-available types out of the box. I could probably work it into a PR with some effort (I don't know much C), but I'd need a steer on a couple of things:
* Should there be separate builds for "global-scope JS" vs "ESM" vs "TypeScript" (see also https://github.com/snowballstem/snowball/pull/155)? My feeling is that there's not much point having a separate TS build, as TS-compatible types can be supplied by JSDoc without any loss of JS compatibility. But it might be worthwhile keeping the old global-scope build for backward-compatibility, e.g. to continue supporting older versions of Nodejs.
* If it's worthwhile having more than 1 JS/TS build, how should this be specified in CLI flags? E.g. supplying `-esm` as an additional flag along with `-js`, or `-esm` being a completely separate flag from `-js`?
[...]

PR #234, opened 2025-05-18
Add JavaScript Module support

AA-Turner:
cc @ojwb 

This is based on #221 with significant revisions, so I've rewritten the history but kept co-authored-by credit.

This PR:

* Removes use of the `.mjs` file extension, which [is not reccomended](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#aside_%E2%80%94_.mjs_versus_.js). We use `.esm.js` instead, to avoid overwriting the 'global' variant.
* Rewrites `BaseStemmer` as a [JavaScript class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_classes). The `export { BaseStemmer }` line is conditionally appended via the makefile. [It is possible](https://stackoverflow.com/a/72314371) to put this in the file itself, but too fragile for my taste.
* Removes Node.js-specific content (`module.exports`)
* Adds an 'ESM' / JS Modules variant of the JS generator to generate and export `{LANG}Stemmer` as an anonymous class via an immediately-executed function.
  * My understanding and testing is that this should work cross-file, but if not we could add `globalThis.{LANG}Stemmer = {LANG}Stemmer` to explicitly export to the global scope.
* Adds an `--esm` mode to `stemwords.js` for testing.
* Uses `const` and `let` within the bodies of 'global' variant stemmers.

I think it would be possible to rewrite the entire JS generator to use (proper) JS classes, in the style of `base-stemmer.js` in this PR. Doing this properly would enable use of `#private` fields and methods, which would remove the complexity of the current IIFEs whilst achieving the same thing. In this scenario, the difference between 'ESM' and 'Global' mode would just be the `export` line. This is more work, though.

Let me know what you think,
Adam

xref:
* #183 
* #221

Feel free to push to this branch / use as you see fit.

### DIFF
--- a/compiler/generator_js.c
+++ b/compiler/generator_js.c
@@ -218,19 +218,11 @@ static void writef(struct generator * g, const char * input, struct node * p) {
             case 'n': write_string(g, g->options->name); continue;
             case 'P': write_string(g, g->options->parent_class_name); continue;
             case 'C': { // Constant.
-                if (g->options->js_esm) {
-                    w(g, "const");
-                } else {
-                    w(g, "/** @const */ var");
-                }
+                w(g, "const");
                 continue;
             }
             case 'D': { // Declare variable.
-                if (g->options->js_esm) {
-                    w(g, "let");
-                } else {
-                    w(g, "var");
-                }
+                w(g, "let");
                 continue;
             }
             default:
@@ -1053,7 +1045,7 @@ static void generate_define(struct generator * g, struct node * p) {
     write_newline(g);
     write_comment(g, p);
 
-    if (q->type == t_routine) {
+    if (g->options->js_esm || q->type == t_routine) {
         writef(g, "~M/** @return {boolean} */~N"
                   "~Mfunction ~W() {~+~N", p);
     } else {
@@ -1278,38 +1270,39 @@ static void generate(struct generator * g, struct node * p) {
 
 static void generate_class_begin(struct generator * g) {
     if (g->options->js_esm) {
-        w(g, "// deno-lint-ignore-file~N"
-             "import ~P from './base-stemmer.mjs'~N"
+        w(g, "import { ~P } from './base-stemmer.js'~N"
+             "/** @constructor */~N"
              "~N"
-             "/** @typedef {{ stemWord(word: string): string }} Stemmer */~N"
-             "~N"
-             "/** @type {{ new(): Stemmer }} */~N"
-             "~C ~n = function() {~+~N"
-             "~M~D base = new ~P();~N");
+             "export const ~n = (() => {~+~N");
     } else {
-        w(g, "/**@constructor*/~N"
-             "~D ~n = function() {~+~N"
-             "~M~C ~P = require('./base-stemmer.js');~N"
-             "~M~D base = new ~P();~N");
+        w(g, "/** @constructor */~N"
+             "var ~n = function() {~+~N");
     }
+    w(g, "~M~C base = new ~P();~N");
     write_newline(g);
 }
 
 static void generate_class_end(struct generator * g) {
-    w(g, "~N");
-    w(g, "~M/**@return{string}*/~N");
-    w(g, "~Mthis['stemWord'] = function(/**string*/word) {~+~N");
-    w(g, "~Mbase.setCurrent(word);~N");
-    w(g, "~Mthis.stem();~N");
-    w(g, "~Mreturn base.getCurrent();~N");
-    w(g, "~-~M};~N");
-    w(g, "~-};~N");
+    write_newline(g);
     if (g->options->js_esm) {
-        w(g, "~N"
-             "export default ~n~N");
+        // providing a name for the class is optional
+        w(g, "~Mreturn class ~n {~+~N");
+        w(g, "~M/** @return{string} */~N");
+        w(g, "~MstemWord(/** string */word) {~+~N");
+        w(g, "~Mbase.setCurrent(word);~N");
+        w(g, "~Mstem();~N");
+        w(g, "~Mreturn base.getCurrent();~N");
+        w(g, "~-~M};~N");
+        w(g, "~-~M};~N");
+        w(g, "~-})();~N");
     } else {
-        w(g, "~N"
-             "if (typeof module === 'object' && module.exports) module.exports = ~n;~N");
+        w(g, "~M/** @return{string} */~N");
+        w(g, "~Mthis['stemWord'] = function(/** string */word) {~+~N");
+        w(g, "~Mbase.setCurrent(word);~N");
+        w(g, "~Mthis.stem();~N");
+        w(g, "~Mreturn base.getCurrent();~N");
+        w(g, "~-~M};~N");
+        w(g, "~-};~N");
     }
 }
 
@@ -1389,7 +1382,7 @@ static void generate_members(struct generator * g) {
             case t_integer:
                 w(g, "~M~D /** number */ ");
                 write_varname(g, q);
-                w(g, ";~N");
+                w(g, " = 0;~N");
                 wrote_members = true;
                 break;
             case t_boolean:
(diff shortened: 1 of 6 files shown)


ojwb:
FWIW, I think we've already concluded that we don't need to keep a "global" JS variant as an option, but if we do keep it the reason would be aiming to support older web browsers, so it would seem better to have it more like the current output and avoid introducing use of `const` and `let` into it.

compiler/generator_js.c
```diff
@@ -1278,38 +1278,29 @@ static void generate(struct generator * g, struct node * p) {
...
     w(g, "~-};~N");
+    write_newline(g);
     if (g->options->js_esm) {
-        w(g, "~N"
-             "export default ~n~N");
+        w(g, "export { ~n };~N");
     } else {
-        w(g, "~N"
-             "if (typeof module === 'object' && module.exports) module.exports = ~n;~N");
+        w(g, "globalThis.~n = ~n;~N");
```
AA-Turner, line 1303: This `"globalThis.{LANG}Stemmer = {LANG}Stemmer` line is not strictly required, as `var` has global scope, but is included for explicitness.

AA-Turner:
Thanks for the review @ojwb. I've reverted much of this PR and it now is a minimally correct implementation of JS Modules for Snowball.

I think that the refactoring to JS classes would be useful, but is orthagonal to this work and can be picked up in a later PR.

> FWIW, I think we've already concluded that we don't need to keep a "global" JS variant as an option

The 'global' variant would be useful for Sphinx, as we currently don't use JS Modules for our internal JavaScript, but we can work around changes made by Snowball as the upstream (and indeed already do as of the 3.0.1 release).

> but if we do keep it the reason would be aiming to support older web browsers, so it would seem better to have it more like the current output and avoid introducing use of `const` and `let` into it.

To note, commit 510d933740c4e1564c293db69b5d689210150d2b ("Modernise the JS output a bit") already introduced `const` in the JS output. This is a nominally backwards-incompatible change in terms of syntax, but 'const' is recognised in all browsers from 2012, and correctly supported by all browsers from 2017 ([source](https://caniuse.com/const)).

The bigger problem with that use of `const` is that it breaks scoping, meaning that the `BaseStemmer` class is no longer exported to the global scope. This PR adds an explicit `globalThis` export, restoring support for using `BaseStemmer` via the global scope.

A

ojwb:
> To note, commit [510d933](https://github.com/snowballstem/snowball/commit/510d933740c4e1564c293db69b5d689210150d2b) ("Modernise the JS output a bit") already introduced `const` in the JS output. This is a nominally backwards-incompatible change in terms of syntax, but 'const' is recognised in all browsers from 2012, and correctly supported by all browsers from 2017 ([source](https://caniuse.com/const)).

5a9d01977e68d609f885ae86553add2485f45c53 reverted to not generating `const` by default before that was in a release.  I see there's still a `const` on `BaseStemmer` in `base-stemmer.js` though - that was an oversight.

FWIW, https://caniuse.com/esm seems to suggest that ESM is supported by everything from 2018 on.  I guess it requires updates to user code whereas generating `const` instead of `var` doesn't.

I still think my preferred option is to only generate ESM as having to support multiple JS variants is extra maintenance work.  #183 suggested people wanting support for older implementations could then automatically convert the ESM version which seems a reasonable approach given there's likely to be a minification step anyway:

> The simplest way of doing this would probably be to have a single source of truth with ESM, ES6 classes, and other modern features. You can then build that intermediate output to a downstream target with [SWC](https://github.com/swc-project/swc) or similar. Alternately, only officially support the ESM output and require consumers to build their own downstream versions from that as necessary, maybe with documentation detailing how to do that.

If we are going to support outputting multiple JS variants, then the widest benefit seems to be from to have ESM for modern implementations, and a really low baseline version which should work everywhere.  (That also means I won't have to arbitrate between JS devs with differing opinions of that's too old to worry about who propose or oppose changes to the baseline output!)

AA-Turner:
> I still think my preferred option is to only generate ESM as having to support multiple JS variants is extra maintenance work.

Should I update this PR to unconditionally generate JS Modules, and remove the 'global' code?

> I guess it requires updates to user code whereas generating const instead of var doesn't.

Yes, this is the crux. `const`, `let`, and even `class` syntax are transparent, as the only published interface is `(new Stemmer()).stemWord('spam')`. Using modules requires updating downstream code to explicitly import the requested stemmers.

> there's likely to be a minification step anyway

To note, Sphinx provides both the minified and non-minified versions in generated HTML. See e.g. [`language_data.js`](https://www.sphinx-doc.org/en/master/_static/language_data.js) and [`english-stemmer.js`](https://www.sphinx-doc.org/en/master/_static/english-stemmer.js). Because we include both the base stemmer and stemmer for the natural language in one file, we will have to work out how to deal with JS Modules (we can't require users to suddenly install a JS toolchain!).

> If we are going to support outputting multiple JS variants

The problem of 'a really low baseline version which should work everywhere' is that you then have to arbitrate vendor and browser quirks. In Sphinx, we support the latest major browsers (Firefox, Safari, Chrome/Edge) only. Along these lines, it may be prudent to support the ['baseline'](https://web.dev/baseline) JS features, which is an externally-adjudicated set of widely-available features.

I would (selfishly) still advocate for a 'ES Modules' vs 'Global' flag. The difference is the `import ... from ...` and `export {NAME}` vs `globalThis.{NAME} = {NAME}` lines, which is relatively small. I'm not the maintainer here, though, so you're well within your rights to drop support for the 'global' variant.

A

compiler/header.h
```diff
@@ -388,7 +386,7 @@ struct options {
     FILE * output_h;
     byte syntax_tree;
     byte comments;
-    byte js_esm;
+    bool js_esm;
```
ojwb, line 389: Perhaps we should consider moving over to using C's `bool` type (we do now require a C99 compiler to build the compiler) but I don't think it's helpful to change a single place to use `bool` like this.

AA-Turner:
```suggestion
    byte js_esm;
```

compiler/header.h
```diff
@@ -1,13 +1,11 @@
 #include <stdio.h>
+#include <stdbool.h>  // bool
```
AA-Turner, line 2:
```suggestion
```

AA-Turner:
> If I understood [JS/TS ESM version (modern JS module system vs global scope polution) #183](https://github.com/snowballstem/snowball/issues/183) correctly, the global version won't work with deno, so to test with that you'd want to be able to run only the ESM tests.

I believe, though untested, that the problem here is the *implicit* global export. This would be testable locally by adding a `use strict` directive to enable strict mode. This proposal has *explicit* global exports by way of assigning properties on the global object, so I think Deno (or other strict mode code) would work here.

AA-Turner:
> I'm a bit more persuaded that "modern global" is a reasonable option having thought about this some more, but it does seem much of the additional maintenance overhead is likely to be just in having two variants to take care of, especially as I only really dabble in JS. Only supporting ESM is still seeming rather appealing.

I've pushed a commit to fully switch to JS Modules and remove all vestiges of the 'global' mode. We will probably patch this in Sphinx to keep the global mode (at least for the time being), but the patch is very small (change the 'export' and 'import' lines), so there is greater disutility from having Snowball carry the machinery for two variants.

A

AA-Turner:
P.S.: if this branch is merged, please use a squash-merge, the history is now very untidy!

ojwb:
> > If I understood [JS/TS ESM version (modern JS module system vs global scope polution) #183](https://github.com/snowballstem/snowball/issues/183) correctly, the global version won't work with deno, so to test with that you'd want to be able to run only the ESM tests.
> 
> I believe, though untested, that the problem here is the _implicit_ global export. This would be testable locally by adding a `use strict` directive to enable strict mode. This proposal has _explicit_ global exports by way of assigning properties on the global object, so I think Deno (or other strict mode code) would work here.

I tested adding `"use strict";` to the top of `stemwords.js` and `make check_js` passes (tested at 533eb0ca0b3bcab899fc13c6e5b30a9b18265691) so it seems you may well be right.

I also had a quick poke at adding a CI job using deno, but haven't got it to work yet (trying with ESM so it wasn't a strictness problem with globals).  I'll have another go once we've resolved this, but it might just need someone who knows more about it than me.

PR #235, opened 2025-05-18
Remove 'module.exports' in JavaScript

AA-Turner:
A much smaller alternate PR to unblock Sphinx in the immediate instance. `module.exports` prevents the files from being loaded in the browser.

A

### DIFF
--- a/compiler/generator_js.c
+++ b/compiler/generator_js.c
@@ -1308,8 +1308,7 @@ static void generate_class_end(struct generator * g) {
         w(g, "~N"
              "export default ~n~N");
     } else {
-        w(g, "~N"
-             "if (typeof module === 'object' && module.exports) module.exports = ~n;~N");
+        w(g, "~N");
     }
 }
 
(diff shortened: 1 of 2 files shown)


ojwb:
Unfortunately this causes CI to fail as it breaks the code which selects the requested stemmer in `stemwords.js`:

```
function create (name) {                                                        
    var lc_name = name.toLowerCase();                                           
    if (!lc_name.match('\\W') && lc_name != 'base') {                           
        try {                                                                   
            const Stemmer = require(lc_name + '-stemmer.js');                   
            return new Stemmer();                                               
        } catch (error) {                                                       
        }                                                                       
    }                                                                           
    console.log('Unknown stemming language: ' + name + '\n');                   
    usage();                                                                    
    process.exit(1);                                                            
}                                                                               
```

In the browser, shouldn't `typeof module` return `'undefined`' and so `module.exports` not actually get used?  That seems to be the case from testing with firefox at least.

AA-Turner:
Hmm, perhaps the problems I was seeing were as we run our tests via Node but normal execution is via the browser. I'll investigate.

Closing this in the meantime.

A

PR #238, opened 2025-05-19
Allow parallel execution for the 'check_*' targets in the Makefile

### LINKED ISSUE #237: `make check_*` fails when parallelised
I kept getting failures on `make -j30 check_js`, whereas `make check_js` worked, albeit painfully slowly.

I believe the issue is that each `check_{LANG}_%` target uses the same temporary files for each natural language, which are then overwritten causing `diff` to create spurious output.

A

AA-Turner:
Closes #237.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,7 +102,7 @@ jobs:
 
     env:
       CC: ${{ matrix.CC || 'gcc' }}
-      MAKE: ${{ matrix.MAKE || 'make' }}
+      MAKE: ${{ matrix.MAKE || 'make' }} --output-sync --jobs=4
       STEMMING_DATA: 'snowball-data'
 
     steps:
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -523,40 +523,40 @@ check_koi8r: $(KOI8_R_algorithms:%=check_koi8r_%)
 check_utf8_%: $(STEMMING_DATA)/% stemwords$(EXEEXT)
 	@echo "Checking output of $* stemmer with UTF-8"
 	@if test -f '$</voc.txt.gz' ; then \
-	  gzip -dc '$</voc.txt.gz'|./stemwords$(EXEEXT) -c UTF_8 -l $* -o tmp.txt; \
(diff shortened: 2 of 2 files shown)


ojwb:
Thanks.

This doesn't seem to protect runs with different encodings and/or target languages from each other though, so with enough parallelism things will still presumably fail?

Also as-is it rather messes up my current development workflow for evaluating a proposed change to an algorithm - currently I can patch an algorithm, run the tests, and evaluate the changes in the one which failed by running snowball-data's stemmer-compare script giving it just the language name:

```
scripts/stemmer-compare german
```

With this patch applied I'd have to instead run something like this which is rather more annoying to need to type:

```
scripts/stemmer-compare german ../snowball/tmp_german.txt
```

That's easy enough to fix for the patch as in this PR since `stemmer-compare` can be adjusted to use the new temporary filename scheme, but that becomes more difficult if the temporary filename includes the encoding and target language as well.

I think I'll have to ponder how best to resolve this.

AA-Turner:
Personally, I'd try and have the tests run in-memory & avoid the temporary files entirely. This would also help with cross-platform development (I use Windows, and so I'm running all my contributions to `Snowball` through a virtualised Linux OS).

A second-best would be to e.g. use all relevant properties to uniqely identify the file (natural language, programming language, possibly character encoding?) and maybe also the hash of the input data, to ensure no file collisions.

Maybe this 'parallel mode' could be enabled iff jobs>1, though, to preserve your current workflow. 

A

ojwb:
My fix for Ada was to use grouped targets, which are new in GNU make 4.3 (released 2020-01-19 so more than 5 years ago), and the syntax used results in an error for a pattern rule because the `&` is parsed as a non-pattern target by older GNU make versions.

I guess we could mess around with an explicit lock file, or (probably) apply `.NOTPARALLEL` for ada, or require GNU make > 4.3.

Apparently "To detect this feature search for 'grouped-target' in the .FEATURES special variable", so perhaps we could `.NOTPARALLEL` ada only for older GNU make.

ojwb:
It seems `.NOTPARALLEL` can only apply to a whole makefile so instead I've adjusted to conditionally use a rule which works in usual situations for older make.  The case we don't handle with older make hasn't worked at all previously and nobody seems to have noticed before.

All the changes here are either merged or superseded by a different approach.

PR #240, opened 2025-05-28
Emit JS classes

AA-Turner:
cc @ojwb 

This is a sketch of emitting JavaScript classes in the snowball compiler. It doesn't currently pass & needs work, but I'm not comfortable proposing changes more invasive than this.

The challenge is with the callback functions in the among arrays. A minimal example is below -- I believe that the functions will need to be changed in some way to pass explicit state or context.

A different solution would be to make the methods and fields public, but that isn't in keeping with the other programming languages, or the current JS implementation.

A

```js
class Base {
    act(w) { console.log(w[0]()); }
}
class C extends Base {
    get #arr() { console.log("#arr"); return [this.#food]; }
    get #arr2() { console.log("arr2"); return [this.food2]; }
    #lobster = "lobster";
    lobster2 = "lobster (public)";
    #food() { console.log("#food"); return this.#lobster; }
    food2() { console.log("food2"); return this.lobster2; }
    spam() { console.log("spam"); this.act(this.#arr); }
    spam2() { console.log("spam2"); this.act(this.#arr2); }
}
let c = new C();
c.spam();  // raises TypeError: can't access private field or method
c.spam2();  // logs 'undefined' instead of 'lobster'
```

### DIFF
--- a/compiler/generator_js.c
+++ b/compiler/generator_js.c
@@ -78,13 +78,13 @@ static void write_savecursor(struct generator * g, struct node * p,
                              struct str * savevar) {
     g->B[0] = str_data(savevar);
     g->S[1] = "";
-    if (p->mode != m_forward) g->S[1] = "base.limit - ";
-    writef(g, "~Mconst /** number */ ~B0 = ~S1base.cursor;~N", p);
+    if (p->mode != m_forward) g->S[1] = "this.limit - ";
+    writef(g, "~Mconst /** number */ ~B0 = ~S1this.cursor;~N", p);
 }
 
 static void append_restore_string(struct node * p, struct str * out, struct str * savevar) {
-    str_append_string(out, "base.cursor = ");
-    if (p->mode != m_forward) str_append_string(out, "base.limit - ");
+    str_append_string(out, "this.cursor = ");
+    if (p->mode != m_forward) str_append_string(out, "this.limit - ");
     str_append(out, savevar);
     str_append_string(out, ";");
 }
@@ -97,7 +97,7 @@ static void write_restorecursor(struct generator * g, struct node * p, struct st
 
 static void write_inc_cursor(struct generator * g, struct node * p) {
     write_margin(g);
-    write_string(g, p->mode == m_forward ? "base.cursor++;" : "base.cursor--;");
+    write_string(g, p->mode == m_forward ? "this.cursor++;" : "this.cursor--;");
     write_newline(g);
 }
 
@@ -162,9 +162,9 @@ static void write_failure_if(struct generator * g, const char * s, struct node *
 /* if at limit fail */
 static void write_check_limit(struct generator * g, struct node * p) {
     if (p->mode == m_forward) {
-        write_failure_if(g, "base.cursor >= base.limit", p);
+        write_failure_if(g, "this.cursor >= this.limit", p);
     } else {
-        write_failure_if(g, "base.cursor <= base.limit_backward", p);
+        write_failure_if(g, "this.cursor <= this.limit_backward", p);
     }
 }
 
@@ -275,16 +275,16 @@ static void generate_AE(struct generator * g, struct node * p) {
             write_char(g, ')');
             break;
         case c_cursor:
-            w(g, "base.cursor"); break;
+            w(g, "this.cursor"); break;
         case c_limit:
-            w(g, p->mode == m_forward ? "base.limit" : "base.limit_backward"); break;
+            w(g, p->mode == m_forward ? "this.limit" : "this.limit_backward"); break;
         case c_lenof: /* Same as sizeof() for Javascript. */
         case c_sizeof:
             writef(g, "~V.length", p);
             break;
         case c_len: /* Same as size() for Javascript. */
         case c_size:
-            w(g, "base.current.length");
+            w(g, "this.current.length");
             break;
     }
 }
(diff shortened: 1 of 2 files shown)


ojwb:
The `among` function feature is awkward in some other target languages too.

I had a idea a while ago which I've not had a chance to try out yet for an approach which might also help here.

If the `action` in the among array is changed to be an integer then it could be (say) 0 when there's no function, and we assign increasing integers for each different routine used in an among.

We then also generate a dispatch method something like:

```
dispatch_among_function(i) {
    switch (i) {
        case 1: return r_VI();
        default: return r_LONG();
    }
}
```

This dispatch method would then be passed as an extra parameter to find_among()/find_among_b(), with `null` being passed for an among with no functions.

Some algorithms (esperanto, hindi) only use at most one function in any particular among, so for these we don't need to dispatch anything so it'd just be a tail-calling wrapper:

```
dispatch_among_function(i) {
    return r_CONSONANT();
}
```

Maybe there's even a way to just pass the JS function for the routine to find_among() (I have a vague memory you can call functions with extra parameters in JS but maybe I'm thinking of a different language).

For finnish, indonesian and lovins (though the last is really of academic interest only and isn't included by default) we have at least one among using two or more different functions, and these would need an actual dispatcher.  This may add a small overhead to checking a function in an among compared to the current approach, but switch on an integer should be a pretty efficient dispatch.

ojwb:
> Maybe there's even a way to just pass the JS function for the routine to find_among() (I have a vague memory you can call functions with extra parameters in JS but maybe I'm thinking of a different language).

We could add a dummy optional parameter (`i=0`) to the JS method generated for any such routine.  Also often a routine used in amongs is only used in that way and not called directly from code, and for that case we don't even need the default value, just to add a dummy parameter.

ojwb:
@AA-Turner Recent changes on master conflicted, so I've merged master to the branch and pushed.

Did you have any thoughts about instead generating the among and grouping arrays as `const` and at the top level?  Or is there another way to avoid them being created repeatedly?

If having them at the top level seems reasonable, I can port over the change from C/C# to store among functions as an integer in the among array which would allow all amongs to be `const` and outside the class.

AA-Turner:
> Did you have any thoughts about instead generating the among and grouping arrays as const and at the top level? Or is there another way to avoid them being created repeatedly?

Either top level or private member variables seem fine to me. I had hoped that the various JS engines would be clever enough to realise that a `static get` returning constant data is in effect a read-only constant, but alas no.

> among functions as an integer 

Does this solve the problem of passing around private class functions as first-class objects? One idea I had is to make the among functions `static` and give them an explicit context object, which would be a different approach to solving the problem, but somewhat invasive.

A

ojwb:
I've also tested the extra parameter passing thing - for Javascript we could pass the integer `af` (the among function index) to the dispatch function as a parameter instead of stashing it in a base class property which the dispatch function then reads - this is a patch to do that:

```
diff --git a/compiler/generator_js.c b/compiler/generator_js.c
index 2a8853cf..ff8393a5 100644
--- a/compiler/generator_js.c
+++ b/compiler/generator_js.c
@@ -1356,9 +1356,12 @@ static void generate_among_dispatcher(struct generator * g, struct among * x) {
     write_comment(g, x->node);
 
     g->I[0] = x->number;
-    w(g, "~N~M/** @return {boolean} */~N");
-    w(g, "~M#af_~I0() {~N~+");
-    w(g, "~Mswitch (this.af) {~N~+");
+    w(g, "~N~M/**~N");
+    w(g, "~M * @param {int} af~N");
+    w(g, "~M * @return {boolean}~N");
+    w(g, "~M */~N");
+    w(g, "~M#af_~I0(af) {~N~+");
+    w(g, "~Mswitch (af) {~N~+");
     for (int n = 1; n <= x->function_count; n++) {
         w(g, "~Mcase ");
         write_int(g, n);
diff --git a/javascript/base-stemmer.js b/javascript/base-stemmer.js
index b3dd3ce2..6bfeb331 100644
--- a/javascript/base-stemmer.js
+++ b/javascript/base-stemmer.js
@@ -9,7 +9,6 @@ class BaseStemmer {
         this.limit_backward = 0;
         this.bra = 0;
         this.ket = 0;
-        this.af = 0;
     }
 
     /**
@@ -301,8 +300,7 @@ class BaseStemmer {
             {
                 this.cursor = c + w[0].length;
                 if (w.length < 4) return w[2];
-                this.af = w[3];
-                if (call_among_func.call(this))
+                if (call_among_func.call(this, w[3]))
                 {
                     this.cursor = c + w[0].length;
                     return w[2];
@@ -374,8 +372,7 @@ class BaseStemmer {
             {
                 this.cursor = c - w[0].length;
                 if (w.length < 4) return w[2];
-                this.af = w[3];
-                if (call_among_func.call(this))
+                if (call_among_func.call(this, w[3]))
                 {
                     this.cursor = c - w[0].length;
                     return w[2];
```

This relies on being able to pass more parameters to functions than they expect though, and I've no idea if that's fine or frowned upon - e.g. in `js_out/hindi-stemmer.js` we call `find_among_b()` like so:

```
        if (this.find_among_b(a_0, this.#r_CONSONANT) === 0) return false;
```

and then `find_among_b()` will pass an integer parameter when it calls `this.#r_CONSONANT`, but this function doesn't expect a parameter.  We pass this parameter because it's needed when there's a real dispatch function, e.g. `af_6()` in `js_out/finnish-stemmer.js`.

Is there anything else that needs resolving here before this can be merged?

ojwb:
This looks good to me, and the testsuite passes, so I'm going to merge it rather than risk it sitting here for long enough that other merged changes cause conflicts (more likely than for many patches as it touches a lot of the JS-related code).

If passing `af` as a parameter seems better, or there's anything else that could be improved, we can deal with it as a new patch or patches.

PR #241, opened 2025-05-28
Use strict equality for JavaScript

AA-Turner:
See [MDN (strict equality)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality).

A

### DIFF
--- a/compiler/generator_js.c
+++ b/compiler/generator_js.c
@@ -57,6 +57,15 @@ static void write_margin(struct generator * g) {
     for (int i = 0; i < g->margin; i++) write_string(g, "    ");
 }
 
+static void write_relop(struct generator * g, int relop) {
+    // Relational operators are the same as C, save for (in-)equality.
+    switch (relop) {
+        case c_eq: write_string(g, " === "); break;
+        case c_ne: write_string(g, " !== "); break;
+        default: write_c_relop(g, relop);
+    }
+}
+
 static void write_comment(struct generator * g, struct node * p) {
     if (!g->options->comments) return;
     write_margin(g);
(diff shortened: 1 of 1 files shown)


compiler/generator_js.c
```diff
@@ -767,7 +776,7 @@ static void generate_assignto(struct generator * g, struct node * p) {
 static void generate_sliceto(struct generator * g, struct node * p) {
     write_comment(g, p);
     writef(g, "~M~V = base.slice_to();~N"
-              "~Mif (~V == '') return false;~N", p);
+              "~Mif (~V === '') return false;~N", p);
```
ojwb, line 779: Something that's already broken rather than a problem with your change here...

This code is using an empty string to signal a problem (`slice_check()` failed, which means a logic error in the stemmer's code) but if the ends of the slice are equal, sliceto **should** return an empty string.

It's a latent problem for the Snowball code we currently ship, but it's an incorrect implementation of the Snowball language and we really ought to fix it before it bites someone.

I can see 3 ways to fix it, but I'm not sure of the efficiency overheads of each so if you could give some thoughts that'd be appreciated:

* We could return `null` for the error case (and then test `=== null` here).
* We could return `undefined` for the error case (and then test `=== undefined` here).
* `slice_check()` could instead throw an exception which is caught by a try/catch in `stemWord()` leaving the return value to always be a string here (and eliminating the need to check for failure; this also similarly simplifies calling some other helpers which can also fail but only if `slice_check()` does.  We should never throw and catch in normal use, but I've no idea how much overhead there typically is in just passing through a try block in JS.

Effectively the same problem is also present for Python, with similar possible solutions (`None` instead of `null` or `undefined`) but I suppose the best approach might be different there.

AA-Turner: > I'm not sure of the efficiency overheads of each

I don't know enough about JS to talk about efficiency, sorry. I imagine that a check against a singleton will be faster than catching errors, though. In Python, an `is None` check is much faster.

ojwb: We should never actually catch an error though (if we do, there's a bug in the `.sbl` code, or perhaps in the compiler or runtime).

So the question is not whether throwing and catching an exception is slower than an `is None` check (which I'd agree is almost certainly the case), but rather whether just passing through a `try` block is faster or slower than needing to check return values.  FWIW, I've been unable to measure a difference with `timeit`.

Thanks for the input.  I've opened #242 to track this.

Issue #242, opened 2025-05-29
Buggy sliceto implementation for Go, JS, Python, Rust


ojwb:
_Split out of https://github.com/snowballstem/snowball/pull/241#discussion_r2111310561_

The code we currently generate for these target languages is using an empty string to signal a problem (`slice_check()` failed, which means a logic error in the stemmer's code) but if the ends of the slice are equal, sliceto **should** return an empty string.  Currently that will incorrectly be handled as if there had been an error.

It's a latent problem for the Snowball code we currently ship, but it's an incorrect implementation of the Snowball language and we really ought to fix it before it bites someone.

I can see two reasonable approaches to fix this:

* For weakly typed languages or where strings are nullable we can return `null` for the error case (and then test for that instead of an empty string).  This is essentially what we do for C where a `symbol*` is returned which is `NULL` on error.
* `slice_check()` could instead throw an exception which is caught by a try/catch in `stem()` or `stemWord()` or equivalent, or even allowed to propagate to inform the caller that there's an error (arguably better than quietly swallowing it).  This leaves the return value to always be a string, and eliminates the need to check for failure on every call which simplifies the generate code quite a bit. Java asserts in `slice_check()` which is effectively this approach (except when assertions are compiled away).  Pascal writes `Faulty slice operation.` and halts, which is a similar approach (and doesn't require exceptions).

C#'s `slice_check()` currently does this and then carries on with the operation which uses the slice (not madly keen on carrying on, but a diagnostic seems better than quietly swallowing the error):

```
System.Diagnostics.Trace.WriteLine("faulty slice operation");
```

ojwb:
For languages where both options are possible, which is faster is a question.  We should never actually hit the error case (if we do, there's a bug in the `.sbl` code, or perhaps in the compiler or runtime) so it's the overhead of the non-error case we care about.

For Python I remembered `timeit`, so I did a quick hack implementation for Python and profiled with:

```
$ cd python_check
$ PYTHONPATH=snowballstemmer python3 -m timeit -r 999 -s 'import snowballstemmer' -s 's = snowballstemmer.FrenchStemmer()' 's.stemWord("astrologies")'
```

For both versions the time per loop dances around in the 67.x usec range, with no clear overall difference.  The generated code (counting all the stemmers) is significantly smaller without the need to explicitly check return values for operations involving slices though - 735749 bytes vs 789620 bytes, so ~7% smaller.

[...]

PR #243, opened 2025-06-02
Added PHP generator

timwhitlock:
Hello.

I've been privately using this PHP stemmer for a while, but I pretty much wrote it blind by copying the JavaScript stemmer. I have no real knowledge of Snowball and I don't write C, but this does work and the test data passes.

I won't be offended if you don't want to merge my code, because I'm not a reliable maintainer. I'm happy to keep this as a fork, merging upstream when I have the time, but I thought I'd offer my PR anyway.

I should add that I have tried the [php-stemmer](https://github.com/amaccis/php-stemmer/tree/master) project that uses FFI, and it does run much faster under heavy usage. However the pure code solution is easier to deploy.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,9 @@ jobs:
         - name: javascript_node
           JSRUN: node
           apt_packages: 'nodejs'
+        - name: javascript_deno
+          JSRUN: 'deno --allow-read --allow-write'
+          apt_packages: 'nodejs'
         - name: rust
           RUST: rust
           apt_packages: 'rustc'
@@ -163,6 +166,10 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install -y ${{ matrix.apt_packages }}
+    - uses: denoland/setup-deno@v2
+      if: ${{ startsWith(matrix.JSRUN, 'deno') }}
+      with:
+        deno-version: vx.x.x
     - name: Install mingw64 packages
       if: matrix.mingw64_packages
       uses: msys2/setup-msys2@v2
@@ -200,6 +207,12 @@ jobs:
     - name: Test Javascript
       if: matrix.JSRUN
       run: $MAKE check_js JSRUN="${{ matrix.JSRUN }}"
+    - name: Lint and check Javascript code
+      if: ${{ startsWith(matrix.JSRUN, 'deno') }}
+      run: |
+        deno lint javascript/*.js js_out/*.js
+        # Currently there are some problems in find_among()/find_among_b().
+        deno check javascript/*.js js_out/*.js || true
     - name: Test Rust
       if: matrix.RUST
       run: $MAKE check_rust RUST="${{ matrix.RUST }}"
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 *.o
 /ada/bin/
 /ada/obj/
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -254,12 +254,12 @@ everything: all java csharp pascal js rust go python ada
 
 baseline-create: everything
 	rm -rf *.baseline
-	for d in src_c java csharp pascal js_out go python_out ada ; do cp -a $$d $$d.baseline ; done
+	for d in src_c java csharp pascal js_out rust go python_out ada ; do cp -a $$d $$d.baseline ; done
 	rm -rf *.baseline/*.o ada.baseline/obj pascal.baseline/*.ppu
 	find java.baseline -name '*.class' -delete
 
 baseline-diff:
-	@for d in src_c java csharp pascal js_out go python_out ada ; do diff -ru -x'*.o' -x'obj' -x'*.ppu' -x'*.class' $$d.baseline $$d ; done
+	@for d in src_c java csharp pascal js_out rust go python_out ada ; do diff -ru -x'*.o' -x'obj' -x'*.ppu' -x'*.class' -x'Cargo.lock' -x 'target' $$d.baseline $$d ; done
 
 .PHONY: all clean update_version everything baseline-create baseline-diff
 
--- a/ada/generate/generate.adb
+++ b/ada/generate/generate.adb
@@ -25,12 +25,12 @@ procedure Generate is
       New_Line (File);
       Put (File, "   type Language_Type is (");
       for Lang of Languages loop
-         Put (File, "L_" & To_Upper (Lang));
-         I := I + 1;
-         if I < Natural (Languages.Length) then
+         if I > 0 then
             Put_Line (File, ",");
             Put (File, "                          ");
          end if;
+         Put (File, "L_" & To_Upper (Lang));
+         I := I + 1;
       end loop;
       Put_Line (File, ");");
       New_Line (File);
--- a/ada/src/stemmer.adb
+++ b/ada/src/stemmer.adb
@@ -177,8 +177,8 @@ package body Stemmer with SPARK_Mode is
                   return;
                end if;
                Execute (Context, W.Operation, Status);
-               Context.C := C + Len;
                if Status then
+                  Context.C := C + Len;
                   Result := W.Result;
                   return;
                end if;
@@ -250,8 +250,8 @@ package body Stemmer with SPARK_Mode is
                   return;
                end if;
                Execute (Context, W.Operation, Status);
-               Context.C := C - Len;
                if Status then
+                  Context.C := C - Len;
                   Result := W.Result;
                   return;
                end if;
--- a/algorithms/english.sbl
+++ b/algorithms/english.sbl
@@ -193,6 +193,7 @@ define exception1 as (
 
         /* special changes: */
 
+        'skis'      (<-'ski')
         'skies'     (<-'sky')
 
         /* special -LY cases */
--- a/compiler/analyser.c
+++ b/compiler/analyser.c
@@ -5,38 +5,13 @@
 #include <string.h>  /* memmove */
 #include "header.h"
 
-typedef enum {
-    e_token_omitted = 0,
-    e_unexpected_token = 1,
-    e_string_omitted = 2,
-    e_unexpected_token_in_among = 3,
-    /* For codes above here, report "after " t->previous_token after the error. */
-    e_unresolved_substring = 14,
-    e_not_allowed_inside_reverse = 15,
-    e_empty_grouping = 16,
-    e_already_backwards = 17,
-    e_empty_among = 18,
-    e_adjacent_bracketed_in_among = 19,
-    e_substring_preceded_by_substring = 20,
-    /* For codes below here, tokeniser->s is printed before the error. */
-    e_redeclared = 30,
-    e_undeclared = 31,
-    e_declared_as_different_mode = 32,
-    e_not_of_type_x = 33,
-    e_not_of_type_string_or_integer = 34,
-    e_misplaced = 35,
-    e_redefined = 36,
-    e_misused = 37
-} error_code;
-
 /* recursive usage: */
 
 static void read_program_(struct analyser * a, int terminator);
 static struct node * read_C(struct analyser * a);
-static struct node * C_style(struct analyser * a, const char * s, int token);
+static struct node * new_string_command(struct analyser * a, int token);
 
-
-static void print_node_(struct node * p, int n, const char * s) {
+static void print_node_(const struct node * p, int n, const char * s) {
     printf("%*s%s", n * 2, s, name_of_token(p->type));
     if (p->name) {
         putchar(' ');
@@ -84,20 +59,7 @@ static const char * name_of_mode(int n) {
     exit(1);
 }
 
-static const char * name_of_type(int n) {
-    switch (n) {
-        case 'b': return "boolean";
-        case 's': return "string";
-        case 'i': return "integer";
-        case 'r': return "routine";
-        case 'R': return "routine or grouping";
-        case 'g': return "grouping";
-    }
-    fprintf(stderr, "Invalid type %d in name_of_type()\n", n);
-    exit(1);
-}
-
-static const char * name_of_name_type(int code) {
+static const char * name_of_type(int code) {
     switch (code) {
         case t_string: return "string";
         case t_boolean: return "boolean";
@@ -106,7 +68,7 @@ static const char * name_of_name_type(int code) {
         case t_external: return "external";
         case t_grouping: return "grouping";
     }
-    fprintf(stderr, "Invalid type code %d in name_of_name_type()\n", code);
+    fprintf(stderr, "Invalid type code %d in name_of_type()\n", code);
     exit(1);
 }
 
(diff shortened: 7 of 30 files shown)


ojwb:
`php/.DS_Store` shouldn't be under version control (https://en.wikipedia.org/wiki/.DS_Store#Problems).

I've added it to `.gitignore` on master to avoid it getting accidentally included in future (16155a8c9d86d136670a42ce511f9fa8fae7df58).

php/base-stemmer.php
```diff
@@ -0,0 +1,524 @@
...
+    protected function in_grouping( array $s, int $min, int $max ):bool {
+        if( $this->cursor >= $this->limit ){
+            return false;
+        }
+        $ch = $this->currentCharCodeAt($this->cursor);
+        if ($ch > $max || $ch < $min){
+            return false;
+        }
+        $ch -= $min;
+        if( ($s[ self::unsignedShiftRight($ch,3) ] & (0x1 << ($ch & 0x7))) === 0 ){
```
ojwb, line 49: The values being shifted are all in the range 0 to 255 so you can just use `>>` here.

php/base-stemmer.php
```diff
@@ -0,0 +1,524 @@
...
+        return false;
+    }
+
+
+    protected function eq_s( string $s ):bool {
+        $slength = self::lengthOf($s);
+        if ($this->limit - $this->cursor < $slength ){
+            return false;
+        }
+        if ( $this->currentSlice($this->cursor, $this->cursor + $slength) !== $s) {
```
ojwb, line 199: Presumably using `substr_compare()` would be more efficient here as it would avoid creating a temporary string (unless there's some funky lazy slicing going on):

https://www.php.net/manual/en/function.substr-compare.php

php/base-stemmer.php
```diff
@@ -0,0 +1,524 @@
...
+        $c = $this->cursor;
+        $l = $this->limit;
+
+        $common_i = 0;
+        $common_j = 0;
+
+        $first_key_inspected = false;
+
+        while( true ){
+            $k = $i + self::unsignedShiftRight($j-$i,1);
```
ojwb, line 236: `>>` should be fine here too.

ojwb:
I think you've been misdirected on one fundamental design point by basing this so closely on the Javascript generator.

If I understand correctly, the PHP strings here are encoded as UTF-8 internally, but you've chosen Snowball's cursor positions to count in Unicode code points (which is what Snowball uses for Javascript[*]) rather bytes (which is what PHP string index positions are).

I think you'll find things are both simpler and more efficient if the cursor position is instead in bytes and so aligned better with PHP strings as then you could just use `str_replace()`, etc.  This is how the other target languages which use UTF-8 work - Ada, Go, Rust, and (with suitable options) C.

----

[*] Javascript strings actually use UTF-16, though currently Snowball kind of ignores the existence of surrogate pairs and pretends all Unicode codepoints are the same width in UTF-16.  This is OK for all the languages we currently have algorithms for, but would need addressing if someone wrote a stemmer for a language using characters outside the BMP.

php/base-stemmer.php
```diff
@@ -0,0 +1,524 @@
...
+        $c_bra <= $this->ket and $this->ket += $adjustment;
+    }
+
+
+    protected function slice_to():string {
+        $result = '';
+        if( $this->slice_check() ){
+            $result = $this->currentSlice($this->bra, $this->ket);
+        }
+        return $result;
```
ojwb, line 428: Note: this is buggy in Javascript (#242) - the slice can be empty so `slice_to` can validly return an empty string and so we can't have an empty string being returned as the way to signal an error.

Probably slice_check() should just assert, since this can only happen for a bug in the algorithm, Snowball compiler or runtime, and an out of range value won't result in a buffer overrun like it can in C.

compiler/generator_php.c
```diff
@@ -0,0 +1,1406 @@
...
+            case 'n': write_string(g, g->options->name); continue;
+            case 'P': write_string(g, g->options->parent_class_name); continue;
+            case 'C': { // Constant.
+                w(g, "/* TODO */ const"); // ??
+                continue;
+            }
+            case 'D': { // Declare variable.
+                w(g, "/* TODO ?*/" ); // ??
+                continue;
+            }
```
ojwb, line 230: `~C` and `~D` have now been scrapped (they were JS specific, but we now only support generating ES6 so can just use `const` and `let` unconditionally).

compiler/generator_php.c
```diff
@@ -0,0 +1,1406 @@
...
+static void w(struct generator * g, const char * s) {
+    writef(g, s, NULL);
+}
+
+static void generate_AE(struct generator * g, struct node * p) {
+    const char * s;
+    switch (p->type) {
+        case c_name:
+            write_string(g,"$this->");
+            write_varref(g, p->name); break;
```
ojwb, line 252: The `$this->` should be output by `write_varref()` rather than by everything that calls it - that's the difference between `write_varref()` (a reference to a variable) and `write_varname()` (just the name of the variable, e.g. for when we declare it).  For some languages the two functions (correctly) generate the same thing.

compiler/generator_php.c
```diff
@@ -0,0 +1,1406 @@
...
+        case c_number:
+            write_int(g, p->number); break;
+        case c_maxint:
+            write_string(g, "(-1>>>1)/*FIXME*/"); break;
+        case c_minint:
+            write_string(g, "(~(-1>>>1))/*FIXME*/"); break;
```
ojwb, line 258: Use `PHP_INT_MAX` and `PHP_INT_MIN`.

compiler/generator_php.c
```diff
@@ -0,0 +1,1406 @@
...
+            write_string(g, s); generate_AE(g, p->right); write_char(g, ')'); break;
+        case c_divide:
+            /* Snowball specifies integer division with semantics matching C,
+             * so we need to use `Math.trunc(x/y)` here.
+             */
+            write_string(g, "$this->trunc_FIXME(");
+            generate_AE(g, p->left);
+            write_string(g, " / ");
+            generate_AE(g, p->right);
+            write_char(g, ')');
```
ojwb, line 278: Looks like PHP's `intdiv()` has the correct semantics: https://www.php.net/manual/en/function.intdiv.php

compiler/generator_php.c
```diff
@@ -0,0 +1,1406 @@
...
+static void generate_GO(struct generator * g, struct node * p, int style) {
+    write_comment(g, p);
+
+    int a0 = g->failure_label;
+    struct str * a1 = str_copy(g->failure_str);
+
+    int end_unreachable = false;
+    int golab = new_label(g);
+    g->I[0] = golab;
+    write_comment(g, p);
```
ojwb, line 557: We already wrote the comment above so this looks like it'll give a duplicate comment (needs `-comments` to enable comments).

compiler/generator_php.c
```diff
@@ -0,0 +1,1406 @@
...
+        g->I[1] = v[i].result;
+        g->S[0] = i < x->literalstring_count - 1 ? "," : "";
+
+        w(g, "~M[");
+        write_literal_string(g, v[i].b);
+        w(g, ", ~I0, ~I1");
+        if (v[i].function != NULL) {
+            w(g, ", '");
+            write_varname(g, v[i].function); // callable
+            w(g, "'");
```
ojwb, line 1306: This presumably means we end up calling these functions starting from their name as a string.  That's OK but it might be worth looking at the approach now used for C (since extended to C# as well) to see if it's faster for PHP: c00e8d5cc89f15966bab222398a124c6d882c6bd

The basic idea of the new approach is to store an integer here (0 for no function), then for an among with functions we pass in a dispatch function to `find_among()`.  There are different reasons why this approach is better for C and C#, neither of which are probably relevant for PHP, but for PHP it would avoid having to look up the function by name on every call.

For an among where only one unique function is used (quite a common case) we can just pass that function to `find_among()` as the dispatch function.

For C and C# the number stored here is temporarily stored in the base class instance for the dispatch function to get at it without needing an extra parameter.  For PHP we might just be able to pass an extra parameter and have it ignored in the single unique function case.

ojwb:
We also need a new CI job to run `make check_php` (I suspect adding one would fail right now as CI now runs parallel make but this branch still uses temporary files for all check jobs - see note above).

That makes me wonder what PHP versions this is aiming to support.  We're using type annotations which are new in PHP7 IIRC (and PHP5 is past EOL anyway), but I think PHP8 expanded type annotations.

The docs should clearly state the minimum supported PHP version so users know what to expect, and we should ideally test a PHP version as close as easily possible to the oldest we aim to support and something as recent as easily possible.

compiler/generator_php.c
```diff
@@ -0,0 +1,1406 @@
...
+        w(g, "~Mreturn ");
+        p->right = NULL;
+    } else {
+        w(g, "~Mif (");
+        // We want the inverse of the snowball test here.
+        relop ^= 1;
+    }
+    generate_AE(g, p->left);
+    // Relational operators are the same as C.
+    write_c_relop(g, relop);
```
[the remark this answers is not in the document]
timwhitlock: Fixed in 748b717

compiler/generator_php.c
timwhitlock: Fixed in b92d555

compiler/generator_php.c
timwhitlock: Implemented in 4d520c6

compiler/generator_php.c
timwhitlock: Implemented in 4e78723

compiler/generator_php.c
timwhitlock: Removed in 4d53f9b

compiler/generator_php.c
```diff
@@ -0,0 +1,1406 @@
...
+    write_comment(g, p);
+
+    struct str * savevar = vars_newname(g);
+    g->B[0] = str_data(savevar);
+    writef(g, "~{~C~N"
+              "~M$~B0 = clone $this~N", p);
+
+    ++g->copy_from_count;
+    str_assign(g->failure_str, "clone ");
+    str_append(g->failure_str, savevar);
```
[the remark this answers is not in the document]
timwhitlock: Fixed in 80ded7e

compiler/generator_php.c
```diff
@@ -270,9 +270,9 @@ static void generate_AE(struct generator * g, struct node * p) {
             write_string(g, s); generate_AE(g, p->right); write_char(g, ')'); break;
         case c_divide:
             /* Snowball specifies integer division with semantics matching C,
-             * so we need to use `Math.trunc(x/y)` here.
+             * so we need to use `intdiv($x/$y)` here.
              */
-            write_string(g, "$this->trunc_FIXME(");
+            write_string(g, "intdiv(");
             generate_AE(g, p->left);
             write_string(g, " / ");
```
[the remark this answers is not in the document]
timwhitlock: Fixed in 4940103.

timwhitlock:
Thanks for the explanation. 

Out of interest, I just did a basic refactor of the base stemmer to assume single byte characters, (replacing `mb_substr` with `substr` etc..). Then I stemmed the whole English vocab file (as it should be equivalent, all being ASCII). What I got was a speedup from about 1.3 seconds, to about 1.1 seconds.

Then I ran the same test using using the FFI based php-stemmer. It took less than 0.1 seconds.

If performance is the concern here, then the PHP code route is probably dead in the water. I don't know that speeding it up with byte offsets will make much of a gain. Perhaps my trivial refactoring is not a sufficient example. Either way I won't be starting from scratch with this, and will likely move to the C bindings route for my own purposes.

ojwb:
I see your point and indeed fundamentally a pure-PHP version is unlikely to ever match calling a C implementation, at least with an interpreted PHP implementation.  However the existence of PHP bindings for the C version doesn't mean people won't care at all about the performance of the pure PHP version.  0.2 extra on 1.1 is 18% extra overhead, and as we improve Snowball's code generation that'll probably be 0.2 extra on 1.0 (20%), then 0.2 extra on 0.9 (22%), etc.

The total time to run `make check_php` is a concern for development too - I just tested and with the PHP bindings already generated and built it took:

```
real	15m50.537s
user	15m29.620s
sys	0m29.883s
```

Parallel make helps, and hopefully we can replace `arabic/voc.txt` with a smaller version, but keeping the time to run tests as low as we can is a concern.

I also worry taking a hybrid approach to handling Unicode compared to all the existing target languages will add to the effort needed to maintain this.

I don't think this requires starting from scratch - it's only the code which deals with cursor positions in strings.  Could you push a branch with your test version with `substr()` instead of `mb_substr()`?  I can take a look at some point.

timwhitlock:
I've pushed a branch to my fork called `single-byte-chars-test`. Obviously it only pass the tests for English.

ojwb:
I get about a 33% speedup from the branch with PHP 8.4 on my aging laptop which is still an x86 install.  Maybe it's architecture and/or PHP version dependent.

I had a go at hacking the branch to fully handle the cursor in byte offsets.  I have it working for some stemmers, but a few fail (e.g. danish and dutch_porter) so it seems some particular case isn't getting handled correctly but I'm failing to spot what right now.

The hacked up branch is a bit slower than your naive conversion (based on timings for English), but still faster than having the cursor work in character offsets.  I suspect we can optimise too as some of my hacked up code has scope for improving (e.g. I think among can just compare bytes rather than decoding UTF-8).

I've pushed a `php-generator` branch.

timwhitlock:
Thanks a lot for this. For English I'm getting 0.83 seconds vs my original 1.3 seconds.

I'm way out of my depth here, so I can't offer any improvements.

timwhitlock:
I've always used [XDebug's profiler](https://xdebug.org/docs/profiler), but the fiddly bit is viewing the data in CacheGrind. There may be better options since I last refreshed my knowledge on the topic.

The profiler will obviously see a lot of time spent in utility functions like `charCodeAt`, `inc_cursor`, etc.. but I imagine they will be hard to optimise much further.

amaccis:
Hi guys,
while looking at the code in this PR, I was wondering why you're not preparing a composer-ready package.
There’s no composer.json nor composer.lock, and the base-stemmer.php doesn’t seem to follow PSR standards.
Is this a deliberate choice? I had considered adding the PHP generator myself (though it’s no longer necessary thanks to this PR), and if I had, I probably would’ve write a fully composer-compatible package, also for making it available via Packagist.
What are your thoughts?

timwhitlock:
My thoughts are that you could contribute composer-compatible, PSR-compliant code if/when my hacked up version gets merged into the main project. I won't be able to continue as a contributor, as I'm way out of my depth on the subject matter and I don't have the time. Personally I'm looking to pursue the C bindings route for my own projects anyway, as it's ten times faster.

amaccis:
Ok, after the PR is merged I'll check if there's anything I can add to improve the code.

ojwb:
I noticed the among and grouping tables in each subclass are member variables, which means each instantiated object will have their own copy of these tables, and that copy will be created at construction time (possibly even built from scratch for each object), but they are constant data so could be declared as class constants and then only one copy would exist which would reduce memory use and the cost of creating a stemmer object.

That seems to be an improvement so I tried it in 2ff78a027c785b38d322fbf9a86ed2459dcaaae5 and it works.

The patch there uses typed class constants, which [apparently requires PHP 8.3](https://www.php.net/manual/en/language.oop5.constants.php), and it indeed doesn't work with 8.1.  If we make them untyped (i.e. also apply 55c71dec086bccae7938a4f91f928f9e13c407b1) then it works back to PHP 7.4 (which is the current minimum for other reasons).

Is there much benefit to making them typed here?  They're internal data so it doesn't affect the API exposed to users of the stemmers.

If there is a benefit, is requiring PHP 8.3 reasonable?  [It's the oldest in active support](https://www.php.net/supported-versions.php), but 8.1 and 8.2 are still getting security fixes.

(If we do keep 7.4 as the minimum requirement I'll add a CI job for it - there's a PPA of rebuilds of older PHP versions which makes that easy to do.)

timwhitlock:
Many thanks for all your work on this. Works a charm for me. I get about 5 mins for all the tests on my Mac M1. The English test on its own is down to about .8 seconds.

PR #245, opened 2025-06-15
Polish Stemmer v3

mitya57:
This PR supersedes #159 and #220.

I rewrote the Polish stemmer almost from scratch. The main differences compared to previous approaches are:

- I ignore derivational suffixes and focus on inflectional suffixes. Also I ignore the `naj-` superlative prefix. This allows me to keep the code cleaner.
- The list of endings is more systematic and ordered. I tried to cover as many different forms and declension/conjugation patterns as possible.
- The functions are grouped using `or`, not using `do`.
- Added a new step to remove trailing *kreska*, which indicates softness of consonant (like Russian ь) and is missing in most oblique cases because they have other letters on the end.

I decided to make a new PR, first because this is a rewrite so most of old comments are irrelevant, and second to make it easier to compare this approach with the previous ones.

This PR has only one FIXME comment. I tried to resolve it by making p0 hardcoded (`$p0 = 2`), but that behaves differently in UTF-8 and ISO-8859-2 encodings, so I did not go this way after all.

### DIFF
--- a/algorithms/polish.sbl
+++ b/algorithms/polish.sbl
@@ -17,10 +17,7 @@ externals (stem)
 
 routines (
   mark_regions
-  adjectival
-  verb
-  short_verb
-  verb2
+  verb_or_adjective
   noun
   normalize_consonant
 )
@@ -39,47 +36,7 @@ define mark_regions as (
 )
 
 backwardmode (
-  define adjectival as (
-    // This function handles adjectives (including comparative/superlative forms)
-    // as well as participles.
-    [substring] among (
-      'y'          // singular masculine nominative (nowy)
-      'ego' 'iego' // singular masculine genitive (nowego, polskiego)
-      'emu' 'iemu' // singular masculine dative (nowemu, polskiemu)
-      'ym'  'im'   // singular masculine instrumental (nowym, polskim)
-      'ej'  'iej'  // singular feminine genitive (nowej, polskiej)
-      'ych' 'ich'  // plural genitive (nowych, polskich)
-      'ymi' 'imi'  // plural instrumental (nowymi, polskimi)
-        (
-          delete
-          try (
-            [substring] among (
-              'aj{ak}c'  // participle suffix (czytający)
-              '{ak}c'    // participle suffix (lecący)
-              'iejsz'    // comparative suffix (piękniejszy)
-              'sz'       // comparative suffix (lepszy)
-                (delete)
-              'sz{ak}c'  // participle suffix (noszący)
-                (<- 's')
-            )
-          )
-        )
-      // We cannot remove endings like -ą and -e unconditionally, because these
-      // letters appear in too many contexts. But we can safely remove them if we
-      // know that our word is a participle or a comparative/superlative form.
-      'aj{ak}ca'    '{ak}ca'    'iejsza'    'sza'    // singular feminine nominative (czytająca, lecąca, piękniejsza, lepsza)
-      'aj{ak}c{ak}' '{ak}c{ak}' 'iejsz{ak}' 'sz{ak}' // singular feminine accusative (czytającą, lecącą, piękniejszą, lepszą)
-      'aj{ak}ce'    '{ak}ce'    'iejsze'    'sze'    // singular neuter nominative (czytające, lecące, piękniejsze, lepsze)
-        (delete)
-      // Handle participles like nosząca, prosząca.
-      'sz{ak}ca'
-      'sz{ak}c{ak}'
-      'sz{ak}ce'
-        (<- 's')
-    )
-  )
-
-  define verb as (
+  define verb_or_adjective as (
     do (
       setlimit tomark p1 for ([substring]) among (
         // conditionals:
@@ -120,6 +77,13 @@ backwardmode (
       'ajcie'  // imperative 2nd person plural (czytajcie)
       'cie'    // imperative 2nd person plural (chodźcie)
         (delete)
+      'sz{ek}' // present 1st person singular (noszę)
+// Also an adjectival form below (singular feminine accusative).
+// Currently noszą -> no.  FIXME When to apply each rule?
+//      'sz{ak}' // present 3rd person plural (noszą)
+        (<- 's')
+      '{ek}'   // present 1st person singular (lecę)
+        (delete)
 
       // There are short verbs whose root consists of only one consonant, e.g. być, żyć.
       // Stemming them to one letter would merge them with these letters used in other
@@ -132,25 +96,50 @@ backwardmode (
       'li{s'}cie'
       '{l/}y{s'}cie'
         (<- '{l/}')
-    )
-  )
 
-  define verb2 as (
-    // -szą can be also part of comparative forms like piękniejszą,
-    // so we need to process it after we are done with adjectival forms.
-    [substring] among (
-      'sz{ek}' // present 1st person singular (noszę)
-      'sz{ak}' // present 3rd person plural (noszą)
-        (<- 's')
-      '{ek}'   // present 1st person singular (lecę)
+      // Adjectives (including comparative/superlative forms)
+      // as well as participles.
+      'y'          // singular masculine nominative (nowy)
+      'ego' 'iego' // singular masculine genitive (nowego, polskiego)
+      'emu' 'iemu' // singular masculine dative (nowemu, polskiemu)
+      'ym'  'im'   // singular masculine instrumental (nowym, polskim)
+      'ej'  'iej'  // singular feminine genitive (nowej, polskiej)
+      'ych' 'ich'  // plural genitive (nowych, polskich)
+      'ymi' 'imi'  // plural instrumental (nowymi, polskimi)
+        (
+          delete
+          try (
+            [substring] among (
+              'aj{ak}c'  // participle suffix (czytający)
+              '{ak}c'    // participle suffix (lecący)
+              'iejsz'    // comparative suffix (piękniejszy)
+              'sz'       // comparative suffix (lepszy)
+                (delete)
+              'sz{ak}c'  // participle suffix (noszący)
+                (<- 's')
+            )
+          )
+        )
+      // We cannot remove endings like -ą and -e unconditionally, because these
+      // letters appear in too many contexts. But we can safely remove them if we
+      // know that our word is a participle or a comparative/superlative form.
+      'aj{ak}ca'    '{ak}ca'    'iejsza'    'sza'    // singular feminine nominative (czytająca, lecąca, piękniejsza, lepsza)
+      'aj{ak}c{ak}' '{ak}c{ak}' 'iejsz{ak}' 'sz{ak}' // singular feminine accusative (czytającą, lecącą, piękniejszą, lepszą)
+      'aj{ak}ce'    '{ak}ce'    'iejsze'    'sze'    // singular neuter nominative (czytające, lecące, piękniejsze, lepsze)
         (delete)
+      // Handle participles like nosząca, prosząca.
+      'sz{ak}ca'
+      'sz{ak}c{ak}'
+      'sz{ak}ce'
+        (<- 's')
     )
   )
 
   define noun as (
     setlimit tomark p1 for ([substring]) among (
       'a' 'o'                   // singular nominative (książka, lato)
-      'i' 'y' 'u' 'ia'          // singular genitive (książki, kobiety, stołu, słonia)
+      // 'y' removed as an adjectival ending.
+      'i' /*'y'*/ 'u' 'ia'          // singular genitive (książki, kobiety, stołu, słonia)
       'owi' 'iowi'              // singular dative (stołowi, słoniowi)
       '{ak}' 'i{ak}' 'em' 'iem' // singular instrumental (książką, możliwością, stołem, słoniem)
       'e' 'iu'                  // singular locative (stole, słoniu)
@@ -182,10 +171,8 @@ define stem as (
   (
     hop 2
     backwards (
-      verb
-      or adjectival
+      verb_or_adjective
       or noun
-      or verb2
     )
   ) or (
     backwards normalize_consonant
--- a/algorithms/polish.sbl
+++ b/algorithms/polish.sbl
@@ -17,9 +17,10 @@ externals (stem)
 
 routines (
   mark_regions
-  verb_or_adjective
+  remove_endings
   noun
   normalize_consonant
+  R1
 )
 
 integers ( p1 )
@@ -36,7 +37,9 @@ define mark_regions as (
 )
 
 backwardmode (
-  define verb_or_adjective as (
+  define R1 as ($p1 <= cursor)
+
+  define remove_endings as (
     do (
       setlimit tomark p1 for ([substring]) among (
         // conditionals:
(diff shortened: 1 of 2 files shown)


ojwb:
Thanks.

> This PR has only one FIXME comment. I tried to resolve it by making p0 hardcoded ($p0 = 2), but that behaves differently in UTF-8 and ISO-8859-2 encodings, so I did not go this way after all.

Also different for UTF-8 vs wide characters.

My immediate thought was Czech which has syllabic consonants, but according to https://en.wikipedia.org/wiki/Polish_language#Consonant_distribution : _Unlike languages such as Czech, Polish does not have syllabic consonants – the nucleus of a syllable is always a vowel._

Maybe it's worth trying to think about whether there are rules about these consonant clusters we can use though.

If you want the cursor two Unicode characters in, then `next 2` gives that.  E.g. see `mark_regions` in `german.sbl` which enforces a minimum of 3 characters before `p1`.

Are you going to be at debcamp/debconf next month?

mitya57:
> Maybe it's worth trying to think about whether there are rules about these consonant clusters we can use though.

I don't think there is a law here.
- *chcieć* and *brać* used to have a yer sound, which was regularly dropped. An alternative form *chocieć* existed for some time, though.
- But not every consonant cluster arose this way: for example, *grać* is a clipped version of *igrać*, and did not have a yer.
- Verbs like *mieć*, *być*, *pić*, *wyć*, *ryć*, *żyć* do not have any clusters at all, just one consonant. There are not many such verbs, though (these are all that I managed to come up with).

> If you want the cursor two Unicode characters in, then `next 2` gives that.

Thank you, I will give it a try. Number 2 was chosen quite arbitrarily, maybe I should also play with 1 and 3 and look at the results.

> Are you going to be at debcamp/debconf next month?

Unfortunately, no…

ojwb:
> > Maybe it's worth trying to think about whether there are rules about these consonant clusters we can use though.
> 
> I don't think there is a law here.

Fair enough.  I noticed Wikipedia says "Polish can have word-initial and word-medial clusters of up to four consonants" but that's probably not helpful unless it's possible to have five or more initial consonants and we should break after four in those cases.

>* Verbs like _mieć_, _być_, _pić_, _wyć_, _ryć_, _żyć_ do not have any clusters at all, just one consonant. There are not many such verbs, though (these are all that I managed to come up with). Edit: after further thinking I can also name _myć_ and _nyć_.

Producing single letter stems, even if linguistically correct, can be problematic - there just aren't very many single letters so unwanted conflation is a bigger risk - e.g. consider discussion of "the letter b", musical notes: "in the key of B minor", chemical elements: "B is the symbol for Boron", itemisation: "In case b, we see [...]", "blood type: B", "vitamin B", "Hepatitis B", people's initials: "B. Wilson", etc.

There's of course unwanted conflation in search results between all of these already, but stemming forms of _być_ to _b_ will likely mean searching for them produces a lot more noise and the benefits of conflating the different forms will tend to be significantly reduced if not wiped out entirely.

So it might be better to just not conflate forms of such words, or if possible map them to a longer stem.  Grouping some forms together on to a longer stem (or onto several longer stems) could also be an option (probably terrible example just to illustrate what I mean: for _być_, conflating _byłem_, _byłam_, _byłom_, _byłyśmy_, etc on stem _był_).

algorithms/polish.sbl
```diff
@@ -0,0 +1,200 @@
...
+      'li{s'}my'
+      '{l/}y{s'}my'
+      'li{s'}cie'
+      '{l/}y{s'}cie'
+        (<- '{l/}')
+      // Handle the imperative 2nd person plural form (chodźcie) here as well.
+      // We cannot do that earlier, because that would also affect the 2nd person
+      // plural past tense forms.
+      'cie'
+        (delete)
```
ojwb, line 147: This results in an empty stem for input _cie_.  It seems it's even a Polish dialect word: https://en.wiktionary.org/wiki/cie#Polish

The rules above can also result in single letter stems - e.g. _liście_ -> _ł_ from the sample vocabulary.  Perhaps we should apply `p0` here too (but see the other comment I'm about to add).

algorithms/polish.sbl
```diff
@@ -0,0 +1,200 @@
...
+  do mark_regions
+  backwards (
+    verb
+    or short_verb
+    or adjectival
+    or noun
+    or normalize_consonant
+    or verb2
+  )
+)
```
ojwb, line 200: It seems we want the `p0` limit pretty much everywhere except where we want `p1`, in which case we can just apply it at the top level like so (and then `setlimit tomark p0 for ([substring])` -> `[substring]` elsewhere):

```
define stem as (
  do mark_regions
  // Make sure we don't produce too short outputs.  The "backwards" will
  // set the backwards limit to the current cursor position.
  hop 2
  backwards (
    verb
    or short_verb
    or adjectival
    or noun
    or normalize_consonant
    or verb2
  )
)
```

This results in two changes in the sample vocabulary (and also fixes the `cie` case I noted above) - before:

```
$ printf '%s\n' 'cie' 'liście' 'oś'|./stemwords -l pl -p
cie -> 
liście -> ł
oś -> os
```

after:

```
$ printf '%s\n' 'cie' 'liście' 'oś'|./stemwords -l pl -p
cie -> cie
liście -> liś
oś -> oś
```

In terms of changes to groupings, the only effect is:

```
{ os osi osiem osoby osy | oś }
```

That's slightly worse it seems, so perhaps we don't want to apply the limit to `normalize_consonant` (and it seems reasonable that morphing the consonant at the end the stem can happen within the limit).  Currently we also remove acute accents from inputs which are individual characters:

```
$ printf '%s\n' 'ć' 'ń' 'ś' 'ź'|./stemwords -l pl -p
ć -> c
ń -> n
ś -> s
ź -> z
```

Not sure how common individual accented letters will be in Polish text, and of the 8 only _z_ seems to be a Polish word, but I suspect it's better not to remove accents in these cases.

So as well as the change above, we could move `normalize_consonant` after `verb2` (since the suffixes they check for are mutually exclusive so the order doesn't make a difference, but grouping `verb2` with the other checks allows applying the limit just once) and not apply the 2 character limit to it.  Then add `not atlimit` in `normalize_consonant` so it doesn't modify _ć_, etc.

Here's a version to show what I mean:

https://github.com/ojwb/snowball/commit/73696f2e83878c3b1facd38b90cd2ae40eef9b73

algorithms/polish.sbl
```diff
@@ -0,0 +1,200 @@
...
+        (delete)
+    )
+  )
+
+  define verb2 as (
+    // -szą can be also part of comparative forms like piękniejszą,
+    // so we need to process it after we are done with adjectival forms.
+    setlimit tomark p0 for ([substring]) among (
+      'sz{ek}' // present 1st person singular (noszę)
+      'sz{ak}' // present 3rd person plural (noszą)
```
ojwb, line 156: As far as I can see we can never match `'sz{ak}'` here because it's also handled in `adjectival` (with action `delete` instead of `<- 's'`).  Testing the example word here, it gets stemmed to `no` not `nos`.

algorithms/polish.sbl
```diff
@@ -0,0 +1,200 @@
...
+        (<- 's')
+      '{ek}'   // present 1st person singular (lecę)
+        (delete)
+    )
+  )
+
+  define noun as (
+    setlimit tomark p1 for ([substring]) among (
+      'a' 'o'                   // singular nominative (książka, lato)
+      'i' 'y' 'u' 'ia'          // singular genitive (książki, kobiety, stołu, słonia)
```
ojwb, line 166: It looks like `'y'` here can't match because it's also handled in `adjectival`.  The action there is `delete` plus possible additional removals rather than just `delete`, but quite likely that's OK.

algorithms/polish.sbl
ojwb: From on a quick test, it looks like always doing `<- 's'` for `'sz{ak}' is worse than always doing `delete`.

Based on the example words in the comments (R1 starts at the `|` in `nos|zą` and `lep|szą`), I tried leaving the `s` in place for shorter words:

```
      'sz{ak}'  
        (R1 and delete or <-'s')
```

I couldn't tell if that was better (it's at least not obviously worse than always removing it) - might be worth you taking a look at.

mitya57: > I tried leaving the s in place for shorter words:

Both adjectives and verbs can be short and long. But I decided to check your heuristic that short words are most probably verbs. I took words matching `^[a-z]{,3}szą$` from the sample vocabulary:

- *cieszą* — verb
- *dalszą* — adjective
- *duszą* — noun
- *gorszą* — adjective
- *gruszą* — noun
- *kuszą* — noun
- *lepszą* — adjective
- *mszą* — noun
- *muszą* — verb
- *naszą* — pronoun (declines as an adjective)
- *noszą* — verb
- *nowszą* — adjective
- *pieszą* — adjective
- *piszą* — verb
- *proszą* — verb
- *ruszą* — verb
- *skuszą* — verb
- *słyszą* — verb
- *suszą* — noun
- *unoszą* — verb
- *waszą* — pronoun (declines as an adjective)
- *wnoszą* — verb
- *zmuszą* — verb
- *znoszą* — verb

(Nouns in this list have *sz* as part of their root, not as a suffix.)

So, indeed most of the short forms in `-shą` will be verbs. But with any such heuristic there will be many false positives.

What I can say for sure is that forms starting with *naj-* and having *-sz-* + adjectival ending on the end are definitely superlative forms of adjectives and should not be treated as verbs. But my level of snowball knowledge does not allow me to express this rule (I did some attempts earlier but failed).

mitya57: And a short review of changes after applying the suggested change `(R1 and delete or <-'s')` on top of your branch:

source | old | new  | part of speech | comment
------ | --- | ---  | -------------- | -------
cieszą | cie | cies | verb           | looks good
duszą  | du  | dus  | noun           | ideally should be dusz, but better than before
gruszą | gru | grus | noun           | ideally should be grusz, but better than before
kuszą  | ku  | kus  | noun           | ideally should be kusz, but better than before
muszą  | mu  | mus  | verb           | looks good
naszą  | na  | nas  | pronoun        | ideally should be nasz, but better than before
noszą  | no  | nos  | verb           | looks good
pieszą | pie | pies | adjective      | looks good (not a comparative form, sz is part of the root)
piszą  | pi  | pis  | verb           | looks good
proszą | pro | pros | verb           | looks good
ruszą  | ru  | rus  | verb           | ideally should be rusz, but better than before
skuszą | sku | skus | verb           | looks good
słyszą | sły | słys | verb           | ideally should be słysz, but better than before
suszą  | su  | sus  | noun or verb   | ideally should be susz (in both cases), but better than before
waszą  | wa  | was  | pronoun        | ideally should be wasz, but better than before
wnoszą | wno | wnos | verb           | looks good
zmuszą | zmu | zmus | verb           | looks good
znoszą | zno | znos | verb           | looks good 

I don’t see any downsides of this change.

mitya57:
Your modifications look good to me, thank you! I merged them into my branch, and made two commits on top of them.

algorithms/polish.sbl
ojwb: > What I can say for sure is that forms starting with _naj-_ and having _-sz-_ + adjectival ending on the end are definitely superlative forms of adjectives and should not be treated as verbs. But my level of snowball knowledge does not allow me to express this rule

I think you're wanting `reverse` which swaps the cursor and the limit and flips the direction.  Untested, but maybe something like:

```
'sz{ak}'  
        ((R1 or reverse 'naj') and delete or <-'s')
```

Not 100% sure I've inserted that in the right place, but hopefully it shows how you'd use it.

ojwb: Oh, except that we set the backwards limit to be two characters from the start of the word at the top level, so with `reverse` we will test for `naj` there not at the start of the word.

I'll have a think about how best to do this.

ojwb: I was taking a look at this but there don't seem to be any words in `polish/voc.txt` which start _naj_ and reach this rule.

And thinking about it, the R1 definition means that for any word starting _naj_, R1 will start after the _j_ so testing for _naj_ here is redundant.  Were you actually suggesting testing for _naj_ somewhere else?

Marking R1 for words from your `^[a-z]{,3}szą$` list:

- *cies|zą* — verb
- *dal|szą* — adjective
- *dus|zą* — noun
- *gor|szą* — adjective
- *grus|zą* — noun
- *kus|zą* — noun
- *lep|szą* — adjective
- *mszą|* — noun
- *mus|zą* — verb
- *nas|zą* — pronoun (declines as an adjective)
- *nos|zą* — verb
- *now|szą* — adjective
- *pies|zą* — adjective
- *pis|zą* — verb
- *pros|zą* — verb
- *rus|zą* — verb
- *skus|zą* — verb
- *słys|zą* — verb
- *sus|zą* — noun
- *un|oszą* — verb
- *was|zą* — pronoun (declines as an adjective)
- *wnos|zą* — verb
- *zmus|zą* — verb
- *znos|zą* — verb

Ideally we delete _-szą_ for an adjective and replace with _-s_ for a verb.

I notice that most of the adjectives have _-szą_ in R1 because they start consonant vowel consonant - e.g. _dal|szą_.  The only exception seems to be _pies|zą_, so in general the current R1 test is very good there.

It's similarly good for verbs - only _un|oszą_ seems to be mishandled.

If _was|zą_ declines as an adjective then it's handled wrongly, but handling of pronouns is unlikely to be important unless they get conflated with other unrelated words.

IIUC nouns should be handled with replace with _-s_ like verbs and all the nouns here are.

ojwb:
> I've pushed a trivial fix for a warning.

CI failed to find the polish data for some reason.  The previous build passed and this change clearly couldn't cause problems so don't worry about this failure as far as this PR is concerned (I will see if I can work out why it isn't finding the data).

ojwb:
Apologies for abusing your PR branch to debug the CI issue but it needs quite particular circumstances which would be tricky for me to recreate alone.

Unfortunately I just don't see how to make it work in this case though.

We need to know the repo URL and branch name for the PR branch.  We can get the branch name, but the repo URL doesn't seem to be anywhere.  We can use `$GITHUB_ACTOR` to get a username which is the right one when someone opens a PR or that same person pushes to it.  However if a maintainer pushes to someone else's PR branch then the maintainer's user name is in `$GITHUB_ACTOR` instead.

I'd hoped it'd be in an environment variable or there would be a remote pointing at it, but there doesn't seem to be.

The closest I can see is the email address on the (automatically generated) merge commit:

```
$ git -C "$GITHUB_WORKSPACE" show "pull/$GITHUB_REF_NAME"
commit 4830554d3c2044210cb9643438212f8bec54f0fd
Author: Dmitry Shachnev <mitya57@users.noreply.github.com>
Date:   Thu Jul 10 14:31:56 2025 +0000

    Merge c0054b62971b855b4c0590f308e218feec0ccd22 into 9fb4c01fdc355e9deeed0ff2b4671d3faf3eb239
```

I'm not sure if this address is always just the github user name before the `@` or is it might be a different email address for the user (if it could be it isn't useful to us).

I'll try to reset the branch to a sensible state.

algorithms/polish.sbl
mitya57: Okay, I will try to do it, probably next week.

PR #250, opened 2025-08-11
Add Dart generator.

ryanheise:
This was mainly derived from the Java generator, with the among function approach copied from the C-Sharp / C implementations. In Dart, minint and maxint are defined as +/- (2^53 - 1). Technically the -1 may not be needed needed for either max or min (see https://github.com/dart-lang/sdk/issues/41717) but maybe there is no need to test the limits here if this is adequate for purpose. I'm not 100% on whether I've used unreachable as intended, but it passes on all test data.

To publish the distribution, run `dart pub publish` from within the same directory that contains `pubspec.yaml` . This file contains the metadata for the pub package. I don't think there is any longer a way to set the "author" field except for the author to authenticate themselves when running `dart pub publish`.

Before publishing, you can also do:

```console
$ dart pub analyze   # do linter checks
$ dart pub publish --dry-run  # do the validation process without publishing
```

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -24,6 +24,14 @@ JAVA ?= java -ea
 java_src_main_dir = java/org/tartarus/snowball
 java_src_dir = $(java_src_main_dir)/ext
 
+DART ?= dart run --enable-asserts
+dart_src_main_dir = dart/lib
+dart_src_dir = $(dart_src_main_dir)/ext
+dart_runtime_dir = dart/lib/src
+dart_gen_dir = dart/lib/gen
+dart_example_dir = dart/example
+dart_package_dir = dart
+
 MONO ?= mono
 MCS ?= mcs
 csharp_src_main_dir = csharp/Snowball
@@ -104,6 +112,7 @@ COMPILER_SOURCES = compiler/space.c \
 		   compiler/driver.c \
 		   compiler/generator_csharp.c \
 		   compiler/generator_java.c \
+		   compiler/generator_dart.c \
 		   compiler/generator_js.c \
 		   compiler/generator_pascal.c \
 		   compiler/generator_python.c \
@@ -124,6 +133,17 @@ JAVARUNTIME_SOURCES = java/org/tartarus/snowball/Among.java \
 		      java/org/tartarus/snowball/SnowballProgram.java \
 		      java/org/tartarus/snowball/SnowballStemmer.java \
 		      java/org/tartarus/snowball/TestApp.java
+
+DARTRUNTIME_SOURCES = dart/lib/src/snowball.dart \
+		      dart/lib/src/algorithms.dart
+
+DART_PACKAGE_SOURCES = dart/lib/snowball.dart
+
+DART_TEST_SOURCES = dart/example/test_app.dart
+
+DART_PACKAGE_FILES = dart/pubspec.yaml \
+		     dart/analysis_options.yaml \
+		     dart/.gitignore
 
 CSHARP_RUNTIME_SOURCES = csharp/Snowball/Among.cs \
 			 csharp/Snowball/Stemmer.cs \
@@ -175,6 +195,8 @@ C_LIB_HEADERS = $(libstemmer_algorithms:%=$(c_src_dir)/stem_UTF_8_%.h) \
 C_OTHER_SOURCES = $(other_algorithms:%=$(c_src_dir)/stem_UTF_8_%.c)
 C_OTHER_HEADERS = $(other_algorithms:%=$(c_src_dir)/stem_UTF_8_%.h)
 JAVA_SOURCES = $(libstemmer_algorithms:%=$(java_src_dir)/%Stemmer.java)
+DART_SOURCES = $(libstemmer_algorithms:%=$(dart_src_dir)/%_stemmer.dart) \
+	$(dart_runtime_dir)/algorithms.dart
 CSHARP_SOURCES = $(libstemmer_algorithms:%=$(csharp_src_dir)/%Stemmer.generated.cs)
 PASCAL_SOURCES = $(ISO_8859_1_algorithms:%=$(pascal_src_dir)/%Stemmer.pas)
 PYTHON_SOURCES = $(libstemmer_algorithms:%=$(python_output_dir)/%_stemmer.py) \
@@ -218,6 +240,7 @@ clean:
 	      $(C_LIB_SOURCES) $(C_LIB_HEADERS) $(C_LIB_OBJECTS) \
 	      $(C_OTHER_SOURCES) $(C_OTHER_HEADERS) $(C_OTHER_OBJECTS) \
 	      $(JAVA_SOURCES) $(JAVA_CLASSES) $(JAVA_RUNTIME_CLASSES) \
+	      $(DART_SOURCES) \
 	      $(CSHARP_SOURCES) \
 	      $(PASCAL_SOURCES) pascal/stemwords.dpr pascal/stemwords pascal/*.o pascal/*.ppu \
 	      $(PYTHON_SOURCES) \
@@ -234,21 +257,23 @@ clean:
 	-rmdir $(js_output_dir)
 
 update_version:
-	perl -pi -e 's/(SNOWBALL_VERSION.*?)\d+\.\d+\.\d+/$${1}$(SNOWBALL_VERSION)/' \
+	perl -pi -e 's/(SNOWBALL_VERSION.*?)\d+\.\d+\.\d+/$${1}$(SNOWBALL_VERSION)/;' \
+		-e 's/\d+\.\d+\.\d+(.*?SNOWBALL_VERSION)/$(SNOWBALL_VERSION)$$1/;' \
 		compiler/header.h \
 		csharp/Snowball/AssemblyInfo.cs \
+		dart/pubspec.yaml \
 		python/setup.py
 
-everything: all java csharp pascal js rust go python ada
+everything: all java dart csharp pascal js rust go python ada
 
 baseline-create: everything
 	rm -rf *.baseline
-	for d in src_c java csharp pascal js_out rust go python_out ada ; do cp -a $$d $$d.baseline ; done
+	for d in src_c java dart csharp pascal js_out rust go python_out ada ; do cp -a $$d $$d.baseline ; done
 	rm -rf *.baseline/*.o ada.baseline/obj pascal.baseline/*.ppu
 	find java.baseline -name '*.class' -delete
 
 baseline-diff:
-	@for d in src_c java csharp pascal js_out rust go python_out ada ; do diff -ru -x'*.o' -x'obj' -x'*.ppu' -x'*.class' -x'Cargo.lock' -x 'target' $$d.baseline $$d ; done
+	@for d in src_c java dart csharp pascal js_out rust go python_out ada ; do diff -ru -x'*.o' -x'obj' -x'*.ppu' -x'*.class' -x'Cargo.lock' -x 'target' $$d.baseline $$d ; done
 
 .PHONY: all clean update_version everything baseline-create baseline-diff
 
@@ -289,6 +314,9 @@ tests/%.o: tests/%.c
 stemtest$(EXEEXT): $(STEMTEST_OBJECTS) libstemmer.a
 	$(CC) $(CFLAGS) $(LDFLAGS) -o $@ $^
 
+dart/lib/src/algorithms.dart: dart/generate_algorithms.pl libstemmer/modules.txt
+	dart/generate_algorithms.pl $(libstemmer_algorithms) > $@
+
 csharp_stemwords$(EXEEXT): $(CSHARP_STEMWORDS_SOURCES) $(CSHARP_RUNTIME_SOURCES) $(CSHARP_SOURCES)
 	$(MCS) -unsafe -target:exe -out:$@ $(CSHARP_STEMWORDS_SOURCES) $(CSHARP_RUNTIME_SOURCES) $(CSHARP_SOURCES)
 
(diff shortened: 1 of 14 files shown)


ryanheise:
Yes, I am aware of it, although it was producing incorrect results and also crashed on certain inputs, hence I wrote this PR. I used the feedback on #156 as a guide to get certain things right, and ensured it passes on all test files in snowball-data. Integer division is correctly used. For min/max int, see my PR description and the provided link. It wasn't actually clear to me what min/max int are used for, but there are some gotchas that are mentioned in that link. Dart has different limits when targeting JS vs targeting native binaries. Since JS has a more restrictive range, I selected those as the safest min/max. Different from what you might expect, there is not one extra value on the negative side, because the JS numbers are actually floating point, and the min/max integers are simply the most extreme floating point numbers that have an exact integer equivalent. This turns out to be the symmetrical on the positive and negative side. The link I provided shows a gotcha with the interpretation of what is "min/max". Theoretically, I could add +1 to max and -1 to min and those would also be representable as ints, however it would be unsafe to use those ints as an upper bound in a for loop, since the loop condition could never become false. Depending on how maxint is actually used, that might or might not be OK. I didn't have any information on that when creating this port.

GNUmakefile
```diff
@@ -125,6 +134,17 @@ JAVARUNTIME_SOURCES = java/org/tartarus/snowball/Among.java \
...
+DARTRUNTIME_SOURCES = dart/lib/src/snowball.dart \
+		      dart/lib/src/algorithms.dart
+
+DART_PACKAGE_SOURCES = dart/lib/snowball.dart
+
+DART_TEST_SOURCES = dart/example/test_app.dart
+
+DART_PACKAGE_FILES = dart/pubspec.yaml \
+		     dart/analysis_options.yaml \
+		     dart/.gitignore
```
ojwb, line 146: It doesn't seem to make sense to ship `.gitignore` files, and we don't ship any existing ones.

dart/lib/src/algorithms.dart
```diff
@@ -0,0 +1,106 @@
...
+import '../ext/porter_stemmer.dart';
+import '../ext/portuguese_stemmer.dart';
+import '../ext/romanian_stemmer.dart';
+import '../ext/russian_stemmer.dart';
+import '../ext/serbian_stemmer.dart';
+import '../ext/spanish_stemmer.dart';
+import '../ext/swedish_stemmer.dart';
+import '../ext/tamil_stemmer.dart';
+import '../ext/turkish_stemmer.dart';
+import '../ext/yiddish_stemmer.dart';
```
ojwb, line 36: This file needs to be generated from `libstemmer/modules.txt` (see "Don't hardcode algorithm names" in `CONTRIBUTONG.rst`).

ojwb: Hmm, it looks like there's actually already a script to generate it in your patch, so I think you just need to remove this file from git and add it to `.gitignore`.

ojwb:
> In Dart, minint and maxint are defined as +/- (2^53 - 1). Technically the -1 may not be needed needed for either max or min (see [dart-lang/sdk#41717](https://github.com/dart-lang/sdk/issues/41717)) but maybe there is no need to test the limits here if this is adequate for purpose.

I think 2^53 makes more sense here.  The [snowman](https://snowballstem.org/compiler/snowman.html) says "maxint — the maximum positive number" and that integer operations have "the same [...] semantics as C": in C `INT_MIN`/`INT_MAX` are the lowest/highest representable value rather than one inside that limit.

Looking at your link, the argument for using `2^53 - 1` given there is that a loop like `for (var i = 0; i <= limit; i++) { ... }` would never terminate, but consider that in the C equivalent, `i++` when `i` has value `INT_MAX` would invoke undefined behaviour (even if C signed integer overflow were defined to wrap, the loop wouldn't ever terminate).

dart/lib/src/algorithms.dart
ryanheise: That's fair. I originally had in mind having some substitution magic similar to `SNOWBALL_VERSION`, perhaps `SNOWBALL_LANGUAGES`, that would allow applying that substitution to an existing file that is checked in. However, technically if the entire file is generated it doesn't need to be checked in.

I've made the change.

ryanheise:
> Given `maxint` doesn't see much use in practice I'd suggest not to overthink this and just go for `2^53` which I think makes most sense in the context of the existing support for other target languages (or follow Javascript's lead if the logic for `-1>>>1` applies to Dart as well).

If I'm not going to overthink it, then I would just go with what I've already one. `-(2^53 - 1)` to `2^53 - 1` corresponds to the Javascript `Number.MIN_SAFE_INTEGER` and `Number.MAX_SAFE_INTEGER` constants.

If any of the approaches you listed is acceptable to you, then I see no reason why this wouldn't also be acceptable, until you pin down a precise definition of what minint/maxint should actually be. After all, if you want to use this for finding the smallest value, there is no reason to restrict yourself even to the largest countable integer `2^53`, as there are even larger "un"countable integers than that, that would work in this particular use case. If you need it to align with 32 bit arithmetic, then neither of those approaches would be correct, and so since this hasn't been pinned down yet, I would just go with what I already have, and it can always be changed later. The time to consider that may be when snowball starts using maxint for various things that clarify what the exact constraint should be.

GNUmakefile
ryanheise: I'm open to suggestions, but the most similar language would be Python which also has a standard way to distribute packages, similarly python packages have a standard directory structure with a configuration to define metadata for the package management system (package name, description, version, author, dependencies).

Dart has its own version of something like that, with a configuration file to define metadata for the package management system, and standard files put at expected paths (hence the Dart make rule renames `COPYING` to `LICENSE` and `NEWS` to `CHANGELOG.md` when preparing for distribution.) Even if a user wishes to use this Dart library directly from the filesystem, a Dart library is required to have this same directory structure or it can't be used, so this is what the make rule prepares. That same directory structure is in a format that can be officially published to the Dart package manager as well, just as the Python one can be (and has been) published to the Python package manager.

Looking at what the make rules does for Python, it generates a wheel file which is basically a zip file containing files in the expected directory structure for a valid python package.

For my Dart rule, it generates a tar ball of a directory conforming to the expected structure for a valid Dart package. This seems analogous to me.

ojwb:
> I suspect the former can be fixed by generating just `~/=` instead of `/=` (we already use `~/` instead of `/`).

I tested using `~/=` on top of my branch with the unit tests merged in and it does seem to fix this part.

Are you happy for me to just add commits to your branch for simple stuff such as this?  I could open PRs against your repo instead if you'd prefer that.

ryanheise:
Sorry for the delay to reply, yes I'm fine with you adding a commit.

ojwb:
Merging is probably less disruptive to reviewing, and I think I'd squash for the final merge to master anyway.

Don't mind whether you fix the conflict or I do (it seemed like something I shouldn't just do to your branch without raising it first).

ryanheise:
OK, done.

ryanheise:
Fixed.

ryanheise:
Sorry, just noticed another issue.

ryanheise:
So in the `sizelen.sbl` test it does this:

```
  $s = 'test'
  $(lenof s == 4) or fail='lenof string variable incorrect'
  $(sizeof s == 4) or fail='sizeof string variable incorrect'
```

The Dart implementation reports a length of 0 rather than 4, probably because I am not making `s` and `current` "point" to the same string in memory. I don't have enough time tonight, so I'll look into this tomorrow.

ojwb:
The "string-`$`" feature can be a bit of a pain to implement, and the implementation has been buggy in many of the generators.

Ideally an operation like `$s 'x'` (test if `s` has prefix _x_) should be O(1) rather than O(max(`size`, `sizeof s`)).  That's fairly easy to achieve in target languages where it's possible to swap around the values of string variables (e.g. C because strings are really just pointers, Java where variables are references to objects) since we can do something like this:

```
{
  string tmp = current;
  current = s;
  // code "inside" the string-$
  s = current; // Needed if update operations on current might create a new string (e.g. immutable string, COW, or reallocation as the string grows)
  current = tmp;
}
```

(To make the string handling part here clearer, I've omitted saving, resetting and restoring the cursor, limit, etc, and in practice the saving and restoring of `current` is sometimes done along with saving and restoring the `cursor`, etc by copying the base class or context object.)

If swapping around pointers or references is not possible, you'll probably need to copy the strings which may well look very similar to the above, just the underlying operations being invoked are different.

This feature is not used in any of the currently shipped stemmers, so it's probably not worth adding a layer of indirection to try to achieve O(1) if that indirection would add overhead for all string operations, but it'd be better not to copy strings where it's easy to avoid.

Martin Porter's snowball implementation of the Schinke Latin stemmer uses string-$, and it would be hard to rewrite it to not do so, so it is a potentially useful feature for future algorithms.  We don't currently ship this Latin stemmer though, because it produces two stems (noun and verb) for each input which doesn't fit the pattern of the other stemmers.

We could potentially handle this in the compiler by tracking whether operations are being applied to the current string or a variable, and then just generating code based on that (e.g. by mapping node types in the parse tree: `c_len` to `c_lenof`, `c_size` to `c_sizeof`, `c_insert` to (new node type) `c_insert_on_string`, etc).  The wrinkle here is that string-$ is not simply lexically scoped (if you're familiar with Perl, it's more like `local` than `my`) - if a routine is called inside the string-$ it should also work on the string variable, but the same routine could be called outside string-$ and then should work on the current string so we'd need to be prepared to generate multiple versions of such routines (which could just be duplicating them in the parse tree).  That seems feasible to do but perhaps not worthwhile until it's more widely used (Latin's userbase isn't what it used to be!) though it would avoid this being one of the fiddlier parts of a new target language.

ryanheise:
Thanks for the pointer (no pun intended). The latest commit should fix this.

Originally I copied the Java / C Sharp implementations which use mutable string buffers, but then decided it would be easier to just use immutable strings and updated the rest of the code accordingly, except I overlooked this one.

In theory, it would of course be possible to implement mutable string buffers in Dart, but immutable string approach I've used is probably decent given that it uses built-in string functions that are implemented in optimised C. Also, Dart doesn't have `realloc()` while something like:

```dart
// dart
current = current.replaceRange(c_bra, c_ket, s);
```

will probably use `realloc` under the hood in the C implementation, as well as definitely escape Dart's array bounds checking. Probably minor optimisations to the Dart implementation are possible but that can always be done later.

ryanheise:
> will probably use `realloc` under the hood

Actually no, it would have to use `malloc`. Although it would at least be faster than a manual Dart implementation due to avoiding bounds checking and having a better compiler.

ojwb:
Java actually uses `char[]` arrays for variables since ~1 week ago - we switched to using `char[]` for the current string in v3.0.0 which it turns out broke string-$ for Java.  Using the same type seemed the best way to fix that, and also makes switching more efficient.

Immutable strings probably aren't great for code generated from Snowball because every command which modifies the current string will need to allocate a new string and copy the most of the current string to it, whereas most modifications are either removing or replacing characters at the end of the string, and replacements are usually the same length or shorter, so can happen in place.

Seems reasonable to leave this for now though.

ryanheise:
> Seems reasonable to leave this for now though.

Yep, I'd like to have a look at optimising it some time later.

compiler/generator_dart.c
```diff
@@ -931,17 +931,15 @@ static void generate_dollar(struct generator * g, struct node * p) {
               "~Mfinal SnowballProgram ~B0 = SnowballProgram.from(this);~N", p);
 
     ++g->copy_from_count;
-    str_assign(g->failure_str, "copy_from(");
-    str_append(g->failure_str, savevar);
-    str_append_string(g->failure_str, ");");
+
     writef(g, "~Mcurrent = ~V;~N"
```
ojwb, line 935: This looks wrong as now the current string won't get restored on failure if I follow how the generated code will look.

I'll add a testcase for this situation and see.

ojwb: Yes, it mishandles this case.  Go and Rust seem to have the same problem.  I'll see if I can fix them all.

ojwb: > I noticed (incorrectly??) that `failure_str` was only used locally anyway, sort of as a temp variable [...]

`g->failure_str` may get used while generating the code that's lexically inside the string-$, which is generated by:

```
    generate(g, p->left);
```

> you might want to check if Python exhibits the same problem since I mirrored what it was doing.

I've added test coverage for this to the runtime tests which get run for each target language, and the new testcase passes for Python.

However, the implementation here may be broken in a different way for everything except C - C uses the `g->failure_str` that was set on entry - that seems like the right thing to do to me, though I've not yet managed to construct a suitable testcase which passes for C and fails for the others.

The history here is that C was the first and only target language for some time and originally generated the failure code on demand, only tracking an integer code to tell it where to fail to.  Then Java got added but used the approach of pre-generated failure code into a string, and other target languages all followed Java.

Earlier this year I discovered the C approach was buggy as it generates that code using the forwards/backwards mode of where the failure was triggered rather than where it was handled, so I changed it to work more like all the other target languages.  The extra handling C has here might be an unneeded left-over from the old approach, but I don't think so.

I'll push the testcase I have so far with fixes, then poke the related problem some more.

ojwb: OK, I've redone the string-$ implementation for everything (just minor tweaks for C, but everything else now works like C with a boolean flag to detect if the subcommand failed, in which case we now propagate that failure upwards).  I've added a regression test for this, merged master to this branch, and fixed dart similarly.  CI is running, but it passed for my integration branch which has an empty diff to this branch now so it ought to pass too.

I think this may be ready to merge now but I'll take a final look over it, probably on Monday.

PR #253, opened 2025-09-14
Fix java code generation for $ operator

### LINKED ISSUE #252: Java string variable representation
Snowball generates invalid Java code for this reduced testcase:

```
externals (stem)
strings (s)
define stem as $s=''
```

This case isn't currently exercised by any of the stemmers we ship - to reproduce, simply replace e.g. `algorithms/arabic.sbl` with the above and run `make java`:

```
$ make java
cd java && javac  org/tartarus/snowball/ext/arabicStemmer.java
org/tartarus/snowball/ext/arabicStemmer.java:26: error: incompatible types: StringBuilder cannot be converted to char[]
            current = S_s;
                      ^
org/tartarus/snowball/ext/arabicStemmer.java:28: error: cannot find symbol
            limit = current.length();
                           ^
  symbol:   method length()
  location: variable current of type char[]
2 errors
make: *** [GNUmakefile:602: java/org/tartarus/snowball/ext/arabicStemmer.class] Error 1
```

The problem looks to have been introduced by the changes to use `char[]` for the current string (caf9dceb26bf8a1c633d4655401b76b144cad92e).  String variables are still `java.lang.StringBuilder`.

I wonder if we should also switch string variables to be `char[]`.  We'd then need to track their lengths for ourselves, so e.g.:

```
    private java.lang.StringBuilder S_s = new java.lang.StringBuilder();
```

might become:

```
    private char[] S_s = new char[8];
    private int L_s = 0;
```

@rmuir You implemented the change to use `char[]` - thoughts on this?

rmuir:
The internal `current` variable is private and not accessible by the stemmer, and is of type `char[]` which won't work via assignment from a StringBuilder.

Call `setCurrent()` so that the generator creates working valid code.

Closes #252

@ojwb Given none of the current algorithms exercise this, I'm hesitant to try to "optimize" this case in fear I might introduce a bug. With a modern java version this shouldn't be too bad, an unnecessary String object is allocated, but escape analysis should work to reduce any overhead as long as `setCurrent()` can be inlined, I think.

### DIFF
--- a/compiler/generator_java.c
+++ b/compiler/generator_java.c
@@ -886,9 +886,8 @@ static void generate_dollar(struct generator * g, struct node * p) {
     str_assign(g->failure_str, "copy_from(");
     str_append(g->failure_str, savevar);
     str_append_string(g->failure_str, ");");
-    writef(g, "~Mcurrent = ~V;~N"
-              "~Mcursor = 0;~N"
-              "~Mlimit = current.length();~N", p);
+    /* TODO: not optimal */
+    writef(g, "~MsetCurrent(~V.toString());~N", p);
     generate(g, p->left);
     if (!g->unreachable) {
         write_margin(g);


rmuir:
An alternative in the future to speed this case up would be to add a `SnowballProgram.setCurrent(StringBuilder)` to complement the existing `setCurrent(String)` and `setCurrent(char[], int)`. 

It would guarantee size of the internal `char[]` with `StringBuilder.getLength()` and then copy with `StringBuilder.getChars()` to prevent any unnecessary allocation.

It could be also a more generally useful addition to the Java API, to make things easier in case the end-user has a `StringBuilder` already that they want to stem.

ojwb:
Thanks for the quick response.

> The internal `current` variable is private and not accessible by the stemmer

FWIW, it used to be protected (and string-$ used to work for Java).  The change to using `char[]` also made it private.

The main thing I don't like about where we are now is that it seems a cleaner design to use the same type for string variables and the current string.

There is admittedly very limited real world use of string-$ currently - the only instance I know is Martin's implementation of Schinke's Latin stemmer, but that algorithm returns separate noun and verb stems which doesn't fit into the "one word in, one stem out" pattern of the other stemmers so we don't include it in the distributed set of stemmers (also Latin stemming isn't in very high demand).

So in practical terms this being inefficient is not currently a big deal, but switching to work on a string variable and back should really be O(1) - e.g. switching to work on `s` could be something like `char[] saved_current = current; current = S_s;` then save and set the length, cursor, etc, and back could be `S_s = current; current = saved_current;` then restore the length, cursor, etc.  Each of `saved_current = current;` and `current = S_s;` should just make a new reference to an existing object, not actually copy any string contents (unless my mental model of Java is wrong) so these operations should not depend on the length of any of the strings involved.

BTW there is test coverage for this case on the way (which is how I discovered the problem) in https://github.com/ojwb/snowball/tree/runtime-test-suite (warning: currently the runtime tests get run by a rather hacky mechanism in CI and there isn't yet a clean way to run them locally - I plan to sort that out, but I only started working on this on Friday!)

> `SnowballProgram.setCurrent(StringBuilder)`

I'd agree this seems a reasonable addition to the user API, whether or not it's useful here.

rmuir:
I agree, I think the main issue is, it would be some work to clean up.

Most stemmers don't use StringBuilder at all: today it is only Danish, Dutch, and Finnish. StringBuilder and String are even more complicated in modern java: if you are afraid of  copying, just look at their code: these string classes also do on-the-fly utf-8/utf-16 conversions in various operations, with the goal of transparently "compacting" the used memory: 

https://github.com/openjdk/jdk/blob/5271448b3a013b2e3edcd619a4a3b975b292dae1/src/java.base/share/classes/java/lang/AbstractStringBuilder.java

IMO (and biased by benchmarks we've done with lucene), a simple `char[]` + `int length` will perform the best, for stuff like stemming where you process many short strings. The fancy stuff the wrapper objects do only gets in the way and causes slowdowns.

I think your complaints about multiple string representations (StringBuffer, StringBuilder, AbstractStringBuilder, String, CharSequence, `char[]`, ...) is just a fault of the language itself. 

Today: at least for the common cases (anything but Danish, Dutch, and Finish), the stemming can work without allocating any heap objects which avoids the GC pain points.

ojwb:
The change here generates code that doesn't actually work correctly.

This amended example shows the problem:

```
externals (stem)
strings (s)
define stem as ($s='x' =s)
```

This should return output `x` for any input, but actually returns an empty string with your patch.  The problem is inside the `$s` the `='x'` sets the current string to `x` but when we exit the `$s` the variable `s` is not updated.  If we made string variables `char[]` then the current string would reference the same `char[]` as the variable inside the `$s`; otherwise we need to explicitly copy back.

I think I'll try coding up using `char[]` and see how things look.

> there isn't yet a clean way to run them locally

I have improved this somewhat now and you can run the runtime tests locally.  Still only on the branch currently.  Documented near the end of `GNUmakefile`:

```
# Runtime test integration is currently a bit clunky, and you need to switch
# your tree to a different state to run runtime tests.
#
# make clean setup_runtime_tests
#
# Then run them like so:
#
# make STEMMING_DATA=tmp_runtime_tests_snowball_data ALGORITHMS=tests/runtime MODULES=tmp_runtime_tests_snowball_data/modules.txt other_algorithms= check
#
# This will run the runtime tests with the C stemmers - to run for another
# language use e.g `check_python` instead of `check`.
#
# Once you're done, switch the tree back to the normal state:
#
# make clean clean_runtime_tests
```

ojwb:
Using `char[]` for string variables requires changes relating to helper functions such as `slice_from()` where we use the same helper currently for the string literal case and the string variable case.  Both `String` and `StringBuilder` implement `CharSequence` so the helper just takes that.

If string variables change to be `char[]` we can't just pass that for `CharSequence`.

https://docs.oracle.com/javase/8/docs/api/java/lang/CharSequence.html says `Segment` implements `CharSequence` and seems would be useful here, but I see it's `javax.swing.text.Segment` so seems to be part of swing which I think is GUI stuff - is that OK to use here?

ojwb:
I've come up with a working patch for the `char[]` approach (#254) which passes the in-progress runtime tests.

It's more complex than the change here, but the change here doesn't pass the runtime tests and I think fixing that would make it rather more involved.

PR #254, opened 2025-09-24
Java: Fix string-$

### LINKED ISSUE #252: Java string variable representation

ojwb:
We now use char[] for string variables as well as the current string, which makes it much simpler to switch to working on a string variable and back.

Fixes #252
Closes #253

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -121,6 +121,7 @@ RUNTIME_HEADERS  = runtime/api.h \
 		   runtime/header.h
 
 JAVARUNTIME_SOURCES = java/org/tartarus/snowball/Among.java \
+		      java/org/tartarus/snowball/CharArraySequence.java \
 		      java/org/tartarus/snowball/SnowballProgram.java \
 		      java/org/tartarus/snowball/SnowballStemmer.java \
 		      java/org/tartarus/snowball/TestApp.java
--- a/compiler/generator.c
+++ b/compiler/generator.c
@@ -1961,6 +1961,8 @@ extern struct generator * create_generator(struct analyser * a, struct options *
 #ifndef DISABLE_PYTHON
     g->max_label = 0;
 #endif
+    g->java_import_arrays = false;
+    g->java_import_chararraysequence = false;
     return g;
 }
 
--- a/compiler/generator_java.c
+++ b/compiler/generator_java.c
@@ -265,7 +265,7 @@ static void generate_AE(struct generator * g, struct node * p) {
             w(g, p->mode == m_forward ? "limit" : "limit_backward"); break;
         case c_lenof: /* Same as sizeof() for Java. */
         case c_sizeof:
-            writef(g, "~V.length()", p);
+            writef(g, "L~V", p);
             break;
         case c_len: /* Same as size() for Java. */
         case c_size:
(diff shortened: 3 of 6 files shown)


ojwb:
> Publishing these java artifacts in this way is NOT easy and you support a lot of different language bindings, so I'm definitely not asking for that, I'm just explaining what is 'typical'.

Thanks.  I'm definitely keen to avoid getting into maintaining target-language packages - I think that's better done by active users of those languages.

> You may find it in JDK classes, but AFAIK the plan is still to remove it:
>
> -  https://openjdk.org/jeps/154
> -  https://mail.openjdk.org/pipermail/discuss/2025-June/006514.html

Hmm, that JEP is marked "Closed/Withdrawn", and the thread seems to conclude it was probably an April Fool's joke (albeit one which highlights some arguably valid concerns in a humourous way).

> Honestly, unless there is a really good reason, I would remove implements `Serializable` from these classes entirely.

This support was added to Snowball by #43 from @master because apparently _"Classes need to be serializable in order to use them in [Apache Spark](https://spark.apache.org/) and similar frameworks."_ so it sounds like there at least was a pretty good reason in 2016, but possibly it is no longer useful (it seems interesting that another Apache project apparently can't be used with Apache Spark).

I'd really prefer not to have to mediate if this is an area of dispute in the Java community, but if there's a clear consensus I'm happy to go with it.

If we are going to keep it, it would be useful for someone to volunteer to maintain the UID values, or at least give us some easy to follow rules for when to update them.

ojwb:
> Think about it, what state would you be serializing? What state does a stemmer have?

Not sure if that's a rhetorical question, but the obviously externally observable state is the currently set string.

There may also be Java variables corresponding to Snowball variables, which are mostly hidden state of the object (currently at least - conceptually many could be mapped to local variables in the generated Java functions, but they're global in the Snowball program so that's not trivial to do.  We actually have machinery to identify variables only used in one routine, but are missing something to verify that the value of the variable isn't required to persist from one call of the routine to another - it never actually is in non-buggy real world Snowball programs, but we really ought not just assume that).

So outside of `setCurrent()/stem()/getCurrent()` sequences a Snowball stemmer's state likely is not useful to serialise.

However I can see that you might want all objects involved to be serialisable so that the state of a group of objects can be snapshotted by a generic serialisation mechanism, even though some of those objects won't have an interesting state at the points where that happens (but bear in mind I have little idea what Apache Spark does or how it works, or about Java serialisation more generally).  You'd end up serialising some useless data, but only a small amount and the convenience may matter more than minimising the serialisation size.

Anyway, I'm not against removing `implements Serializable`, etc but I'd really like to hear from Oleg as the original contributor of this.

> Keeping the `Serializable` marker interface and the ID does not do any harm, so I think the discussion should be a separate issue.

Indeed (especially as I've now merged this).  I'll open one.

Issue #244, opened 2025-06-12
Javascript improvements


ojwb:
Gathering together things that have been suggested over various issues and PRs:

- [ ] JSDoc types which work with TypeScript (maybe done or at least mostly there now?) e.g. https://github.com/snowballstem/snowball/pull/221/files/f72e3e890587d55861e181c7cc81292471985a11#diff-9a44c8895826a0b981973faa0f4b2f28f018aabfde923d03a406f2a47e82a9f8
- [ ] Allow selecting a stemmer by passing an ISO-639 code (https://github.com/clearlylocal/snowball/blob/esm/esm/core/getStemmerByLocale.mjs but that just hard-codes them all whereas they should be generated from `libstemmer/modules.txt`)
- [ ] #221 used `export default` which seems to avoid the need to calculate the class name of the stemmer as well as the module name; the merged patch calculates both.
- [ ] Currently Snowball externals are generated as private methods (so treated just like Snowball routines) and then a fixed `stemWord()` public method is generated which sets the current word, calls `#stem()`, and returns the current word.  This all seems a bit specific to the stemmers we ship rather than supporting the Snowball language in full.  Conceptually wrong, but not all that important in practice.  If there's a cleaner way to achieve this that would be nice though - e.g. we could apply the setting and getting of the current word to all externals and have a way to specify we want to map Snowball `stem` to Javascript `stemWord()` (e.g. `this.stemWord = this.stem;`)
- [ ] #221 Refactored `stemwords.js` to be fully async.  Not entirely convinced that's helpful - at least to me it makes it harder to follow so worse in its partial role as an example of how to use the API, but maybe if you write a lot of JS async code it's actually easier to read in that form.  It doesn't seem to be measurably faster as async either (it can probably parallelise IO with stemming a bit, but there is presumably extra overhead from the parallelism too).
- [ ] Would it be better to pass `af` from `find_among` to dispatch functions as a parameter: https://github.com/snowballstem/snowball/pull/240#issuecomment-2951638537

ojwb:
`export default` is now used (5c37d044783231e5574c871edf861b73b56efc8f)

ojwb:
> CI now runs `deno check` to check the types with `|| true` after the command for now so we can see the current errors but CI still passes.

23f156f7f28ce6503f81935a9610a76cdf7381ad added `@ts-expect-error` magic comments and drops the `|| true` so regressions will now cause CI to fail.

ojwb:
[...]

PR #258, opened 2025-10-13
Javascript: generate public functions for all externals

simlrh:
Adds a public function to the JS class for each external routine, which does the same getting & setting of the current word as `stemWord`.

~I've left `stemWord` untouched for backwards compatibility - not sure if that's the right thing to do, it creates a reference to a non-existent `#stem` function if the program you're writing isn't a stemmer, although that's not a problem in practice.~

The `stemWord` function is now added only if there is an external with the exact name `stem`.

Addresses one of the minor points in #244

### DIFF
--- a/compiler/generator_js.c
+++ b/compiler/generator_js.c
@@ -1145,12 +1145,9 @@ static void generate_define(struct generator * g, struct node * p) {
         w(g, "~Mreturn this.getCurrent();~N");
         w(g, "~-~M}~N");
 
-        if (!strcmp((const char *) q->s, "stem")) {
+        if (SIZE(q->s) == 4 && memcmp(q->s, "stem", 4) == 0) {
             w(g, "~N");
-            w(g, "~M/**@return{string}*/~N");
-            w(g, "~MstemWord(/**string*/word) {~+~N");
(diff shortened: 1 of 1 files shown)


ojwb:
> This would also mean the compiler would know not to try to localise these variables (there's work in progress to automatically detect Snowball variables which can be made local to one function in the generated target language code).

FYI, this variable localisation is now merged.

> Will get to changes today, thanks for the review

Did you do this?  I don't see a push since.

If it's just the single change I suggested and that looks good to you, I can just fold that in.

simlrh:
I haven't yet sorry, and there's a bug in my `stemWord`, I will spend some time first thing Monday finishing up.

ojwb:
> Complicating the code didn't seem worth it to save the overhead of a single function call.

Seems reasonable.  If the code is minimised it'll probably eliminate that anyway, and an optimising JS interpreter may be able to.  It happens for every word stemmed, so may be worth profiling at some point just to check if it's worth trying harder to change, but really the key thing is it should be no worse than before your patch (since `stemWord = this.stem` should just mean that `stemWord` is exactly the same function as before).

PR #260, opened 2025-11-27
Add sesotho Stemmer

MrLebjane:
This is a stemmer for Sesotho (Southern Sotho), a Bantu language spoken in South Africa and Lesotho.


algorithms/sesotho.sbl
```diff
@@ -0,0 +1,109 @@
...
+  )
+)
+
+/* --- MAIN STEMMER --- */
+define stem as (
+  do mark_regions
+  do remove_nominal_suffixes
+  do remove_verb_suffixes
+  do remove_noun_prefixes
+)
```
ojwb: Just a minor nit, but please include a newline character on the final line (github's red icon here means there isn't one).  It's better to include as some text processing tools can behave unhelpfully without one (and some editors will add automatically add one which can create noise in future PRs).

libstemmer/modules.txt
```diff
@@ -40,6 +40,7 @@ swedish         UTF_8,ISO_8859_1        swedish,sv,swe
 tamil           UTF_8                   tamil,ta,tam
 turkish         UTF_8                   turkish,tr,tur
 yiddish         UTF_8                   yiddish,yi,yid
+sesotho         UTF_8                   sesotho,st,sot
```
ojwb: Judging from the stemmer code, this language seems to use only ASCII characters so we can put `UTF_8,ISO_8859_1` here.

algorithms/sesotho.sbl
```diff
@@ -0,0 +1,110 @@
...
+
+/* --- Remove noun class prefixes --- */
+define remove_noun_prefixes as (
+  [substring] among(
+    'mo'   'ba'   'me'   'le'
+    'ma'   'se'   'boi'   'li'
+  )
+  /* Only delete if there's a vowel after the cursor position */
+  (gopast v delete)
```
ojwb: This can reduce words to a single letter, which is rather aggressive and tends to be problematic - there are only 26 latin letters so only 26 possible one letter stems - also single letters occur in text for other reasons - people's initials, music notes, chemical elements, blood types, itemised lists ("(a)", "(b)", ...), roman numerals, text talking about the letter itself, etc - and it is probably better not to reduce to a single character stem even if that is the true linguistic root.

In your sample vocabulary, the following multi-character words have a single character stem:

```
bao                           o
ea                            e
eang                          e
eile                          e
ii                            i
lee                           e
leo                           o
mae                           e
mau                           u
mee                           e
meea                          e
moea                          e
moeeng                        e
moo                           o
oa                            o
oela                          o
oile                          o
see                           e
seea                          e
seo                           o
ua                            u
unyana                        u
```

I don't know what makes sense linguistically here, but we could require at least two characters, for example:

```
  ((v next) or (next gopast v)  delete)  
```

Or if these prefixes should be followed by a non-vowel:

```
  (non-v gopast v delete)
```

algorithms/sesotho.sbl
```diff
@@ -0,0 +1,110 @@
...
+
+/* --- Integer for tracking position --- */
+integers ( pV )
+
+/* --- Mark vowel region --- */
+define mark_regions as (
+  $pV = limit
+  do (
+    gopast v  setmark pV
+  )
```
ojwb: Similar to the comment below, this will set `pV` just one character in for words which start with a vowel, which can result in single character stems from rules which use `pV` as their limit.

ojwb:
There seem to be some unwanted conflations.  It's hard to completely avoid this happening, but maybe there are conditions or exceptions which could help avoid this?

For example, the following words from your `voc.txt` all stem to _tho_ ("word  stem  ; comment") - I used wiktionary and google translate for meanings (but the latter in particular isn't always reliable):

```
batho                         tho  ; "people"
letho                         tho  ; "nothing"
litho                         tho  ; "members"
motho                         tho  ; "person"
motho                         tho  ; duplicate voc.txt entry!
setho                         tho  ; "member"
thoeng                        tho  ; "so-called"
```

Assuming the translations are accurate, "person"/"people" and "member"/"members" are reasonable to conflate in those pairs (but probably not with each other unless they all actually mean both and that's getting lost in individual word translation to English); "nothing" and "so-called" don't seem at all related.

MrLebjane:
Yes correct, in this case the end result was that "tho" would be the unique stem for people/person. Litho as members does refer to non humans also. I don't think such inconsistencies would happen often, I could increase the word length limit for stemming but would result in understemming for most cases.

"Le" and "eng" are a very common prefix and suffix but not all words have them as that,  for example "letho" and "thoeng" are unique words, "le" and "eng" don't function as a prefix and suffix there, so yes there is some inconsistencies where common patterns don't apply which is a tradeoff.

libstemmer/modules.txt
ojwb: The new vocabulary list has valid Sesotho words containing `š`.  It's apparently only used in Lesotho Sesotho, but it's not in iso-8859-1 so we can't trivially support this stemmer for that encoding.

algorithms/sesotho.sbl
ojwb: With the cleaner `voc.txt` these words are conflated onto single letter stems:

```
a                             a

e                             e
mee                           e
eang                          e
eile                          e
meea                          e
moea                          e
seea                          e
moeeng                        e

iana                          i

o                             o
oa                            o
bao                           o
owa                           o
seo                           o
mooi                          o
oela                          o
oile                          o

ua                            u
liu                           u
mau                           u
unyana                        u
```

I suspect at least some of these aren't Sesotho words, e.g. SEO is probably "Search Engine Optimisation", IANA probably "Internet Assigned Numbers Authority", "Liu" a Chinese surname and "bao" a Chinese bun, but these are all terms that might be searched for in some domains of use of a Sesotho search engine, and reducing them single character stems will result in some irrelevant results being retrieved.

As best I can make out, none of these are **useful** to stem in this way, so it seems imposing a higher minimum on pV wouldn't stop useful stemming, but would prevent some problems.

I tried simply avoiding producing single character stems - that changes the grouping of stems like this:

<ul>
<li>A total of 19 words changed stem
<li>1 words changed stem but aren't interesting
<li>3 splits of groups of stems:
<div>
<div>{ <span title="bao">bao</span> | <span title="moo">mooi</span> | <span title="o">o</span> | <span title="oa">oa</span> | <span title="oel">oela</span> | <span title="oile">oile</span> | <span title="ow">owa</span> | <span title="seo">seo</span> }</div><div>{ <span title="e">e</span> | <span title="eang">eang</span> | <span title="eile">eile</span> | <span title="mee">mee meea</span> | <span title="moe">moea moeeng</span> | <span title="see">seea</span> }</div><div>{ <span title="liu">liu</span> | <span title="mau">mau</span> | <span title="ua">ua</span> | <span title="uny">unyana</span> }</div></div>
</ul>

(You can mouse-over to see the stems, so e.g. "mee" and "meea" are still conflated, now onto stem "mee".)

Are any of the splits cases where it was actually useful to conflate any of the words?  I'm struggling to find Sesotho translations for most of these words (maybe they're mostly foreign words or names).

libstemmer/modules.txt
MrLebjane: Yes in South Africa instead of š we use sh.

algorithms/sesotho.sbl
MrLebjane: 55% of those words are sesotho words including bao and seo. Mee is not a sesotho word. For any of these sesotho words it was not useful to conflate. I think limiting to 2 letter word for stemming would be better yes.

ojwb: Thanks, I'll push that change.

> Mee is not a sesotho word

I'll drop it from the list then so it doesn't confuse any future testing.

ojwb:
> Yes correct, in this case the end result was that "tho" would be the unique stem for people/person. Litho as members does refer to non humans also. I don't think such inconsistencies would happen often, I could increase the word length limit for stemming but would result in understemming for most cases.
> 
> "Le" and "eng" are a very common prefix and suffix but not all words have them as that, for example "letho" and "thoeng" are unique words, "le" and "eng" don't function as a prefix and suffix there, so yes there is some inconsistencies where common patterns don't apply which is a tradeoff.

Understemming errors are typically less of a problem than overstemming errors (because overstemming results in a search matching words unrelated to those in the query, whereas understemming just moves us back towards the situation we are in without a stemmer).

Sometimes this situation can be addressed via extra conditions on suffix removal - for example, "don't remove _-eng_ if the character before is _o_" would avoid _thoeng_ -> _tho_.  I can't see a suitable rule of that type here though - a lot of words in the vocabulary end _-oeng_.  I think we may just have to put up with the overstemming, and if particularly problematic cases are reported we can add specific exceptions (or perhaps find a pattern to the exceptions).

PR #263, opened 2025-12-12
Added more consistent word endings to Lithuanian stemmer

uaimedna:
The current Lithuanian stemmer implementation causes problems for semi international words and their plural forms such as:

kompiuteris, printeris, lazeris (computer, printer, lazer)

While these aren't fully Lithuanian words they are very commonly used when searching for products in Lithuanian.

To motivate this change further even the current examples in algorithm are also broken for real native Lithuanian words. For example:
sesers, seseris, seseriai produce different stems. (sister in english)

Another example of current stemmer generating different stems:
moteris, moteriai (woman in english)

I have produced a small dataset of examples of word stems before changes:

(Native lithuanian)
seseris -> seser
seseriai -> ses
moteris -> moter
moteriai -> mot
moterimi -> mot
moterų -> mot

(Semi native)
kompiuteris -> kompiuter
kompiuteriai -> kompiut
kompiuteriui -> kompiuter
printeris -> printer
printeriai -> print
printeriui -> printer

And after the fix: 

(Native lithuanian)
seseris -> ses
seseriai -> ses
moteris -> mot
moteriai -> mot
moterimi -> mot
moterų -> mot

(Semi native)
kompiuteris -> kompiut
kompiuteriai -> kompiut
kompiuteriui -> kompiut
printeris -> print
printeriai -> print
printeriui -> print

### DIFF
--- a/algorithms/lithuanian.sbl
+++ b/algorithms/lithuanian.sbl
@@ -44,7 +44,7 @@ backwardmode (
     setlimit tomark p1 for ([substring]) among (
       // Daiktavardžiai (Nouns)
       // I linksniuotė (declension I)
-      'as' 'ias' 'is' 'ys' 'eris' // vyras, kelias, brolis, gaidys, seseris
+      'as' 'ias' 'is' 'ys'        // vyras, kelias, brolis, gaidys
       'o'    'io'                 // vyro, kelio
       'ui'   'iui'                // vyrui, keliui
       '{ak}' 'i{ak}' '{ik}'       // vyrą, kelią, brolį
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks for raising this.

I ran a script to evaluate the effects on our sample vocabulary of 86105 words (this was extracted from a wikipedia dump so is likely to contain some non-Lithuanian words, typos, proper nouns, and junk terms, but the stemmer will receive these in real use so how they're handled is somewhat relevant):

https://oligarchy.co.uk/snowball/pr263-compare-lt.html

This analysis looks at the changes between how words are grouped by the stemmer before and after your proposed change.

I don't speak or read Lithuanian, but running it through Firefox's built-in translation feature it doesn't look great.

The first merge is good, the second I'm getting "{ Mystical } + { Mystery }" for which seems questionable if the machine translation is accurate, the third fails to translate usefully.

The 20 splits seem generally unhelpful (a few don't translate usefully; #15 gives "{ mackerel | broker }" which is a good split if correct).

The graphs show the cases where there is movement between groups.  These seem largely neutral with the odd one perhaps worse.

Can you have a look through the output?  BTW you can mouse-over to see the stems.

uaimedna:
Just to clarify: the current stemmer also does not conflate them correctly. I have marked the ancient form in red and modern form used in everyday language with blue:

image

After a quick google search and reading through texts that use the old form of 'seseriai' it seems to be heavily used in biblical or faith context. So I would argue that the split of stem is useful from data retrieval standpoint.

Regarding other words using the same pattern: I have looked through the current wiki dataset and found only one more word following the same pattern that is native Lithuanian: 'dukteriai':

image

Although ideally the two groups should be merged it currently still is an improvement.

Overall these two words are really old and rarely used and would most likely not even appear in any real world data retrieval scenario so I would suggest not adding them as an exception (also considering that the current changes would still Improve the stemming of these words).

mitya57:
Hi all! I am not a native speaker of Lithuanian but I know some basics, so I can clarify some things:
- *s* and *š* are distinct letters. One of them should not be normalized to the other.
- The word *vyšnia* and all its forms should have *š* letter in the root. Spellings with *s* are probably attempts to fit into 7-bit ASCII, which is not really required here.
- *u* and *ū* are also distinct letters, despite the fact that the only difference is the vowel length. What you quoted above about acute, grave, and macron/tilde accents does not apply to the long vowel marker which is part of *ū*.
- However, what you quoted above does apply to the acute accent in *mė́nuoj* and to the tilde accent in *žmuõj*. Indeed, such forms only appear in dictionaries and textbooks. We can normalize out such accents, but I think we can also assume that the input does not have them.
- For completeness: the long vowel *ū* can also have accents over it, e.g. *sū́nūs* (Nom. pl. of ‘son’).

PR #264, opened 2025-12-23
Python: Add type hints and py.typed marker

adamtheturtle:
This PR adds type hints to the Python snowballstemmer package and includes a `py.typed` marker file for PEP 561 compliance.

## Changes

- Add type hints to `among.py`, `basestemmer.py`, `stemwords.py`, `testapp.py`
- Update `create_init.py` to generate typed `__init__.py` with annotations for `algorithms()` and `stemmer()` functions
- Add `py.typed` marker file for PEP 561 compliance
- Update `setup.py` with `package_data` to include `py.typed` in installed package
- Update `MANIFEST.in` and `GNUmakefile` to include `py.typed` in distribution

## Benefits

This allows users of the snowballstemmer package to benefit from static type checking with mypy, pyright, and other type checkers. IDEs will also provide better autocompletion and documentation.

## Compatibility

- Uses `from __future__ import annotations` for forward reference support
- Uses `TYPE_CHECKING` guards to avoid runtime import overhead
- Compatible with Python 3.3+ (the existing minimum supported version)

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -225,7 +225,8 @@ PHP_RUNTIME_SOURCES = php/base-stemmer.php
 # Python
 
 PYTHON_RUNTIME_SOURCES = python/snowballstemmer/basestemmer.py \
-		         python/snowballstemmer/among.py
+		         python/snowballstemmer/among.py \
+		         python/snowballstemmer/py.typed
 
 PYTHON_SAMPLE_SOURCES = python/testapp.py \
 		        python/stemwords.py
--- a/python/MANIFEST.in
+++ b/python/MANIFEST.in
@@ -2,6 +2,7 @@ include *.rst
 include modules.txt
 include setup.*
 recursive-include src *.py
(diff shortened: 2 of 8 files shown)


ojwb:
> @adamtheturtle closed this by deleting the head repository last week

I'm going to assume this means you don't want to continue with this.  If you actually do, please restore your fork and address the points raised.

> This allows users of the snowballstemmer package to benefit from static type checking with mypy, pyright, and other type checkers. IDEs will also provide better autocompletion and documentation.

That's rather shallow and really doesn't explain some of the changes - for example, you add type hints to `stemwords.py` which is an example/test program rather than part of the API the package provides.

(I suspect the summary is LLM generated, and that really isn't very useful here - we're going to fully read any patch we merge anyway, so they really key part is to explain exactly what you're actually trying to achieve with a proposed change.  At least with the current state of LLMs that is much better done directly by the person who made the changes.  If it was human written, please focus on tell us **why** rather than _what_.)

Issue #223, opened 2025-04-02
German: Investigate removing -ge- and -zu-


ojwb:
_Splitting out from [comment in #139](https://github.com/snowballstem/snowball/issues/139#issuecomment-2749995129)_:
> I think we can however probably handle the related case of removing infix -ge- (and also -zu-).  It seems there are a fixed and fairly small set of possible separable verb prefixes and if we check for one of these followed by `ge` or `zu` as a new step after removing suffixes (and skip this new step if we removed a suffix which means the word can't be a verb or derived from a verb) then the results look plausible.  This gives ~260 merges for `ge` and ~220 for `zu`.  I haven't checked more than the first few so it might be we need to exclude some prefixes or adjust the conditions to avoid affecting some cases we aren't actually an infix after a separable verb prefix but look like one.

ojwb:
I've found a problem case for _-ge-_ removal, which is verbs which start which _ge-_ after the separable prefix, e.g. _ausgehen_ and _wiedergeben_, which are also cases which make safely removing a _ge-_ prefix difficult - it seems those two problems are closely connected.  There is at least one difference though - in the infix case we know that _-ge-_ could only have been added if the word looks like a past participle.  I'm not sure that actually helps much though since the possible endings for a past participle (_-t_, _-et_, _-en_) also occur as endings for other verb forms.  Perhaps Martin's conclusion that removing _ge-_ and _-ge-_ are both "_almost intractable problems_" stands.

Martin does however note:

>_Infixed <B>zu</B> can be removed algorithmically, but this minor feature is not shown here_.

Investigating this, I found a case where it would be incorrect to remove _-zu-_ even if we check for a separable verb prefix before it: _zusammenzucken_.  This is the only example I've spotted, but I haven't looked much.  The key question is whether there's an example where removal would cause unwanted conflation or prevent desirable existing conflation (for this case, _zusammenck_ is not linguistically correct but is OK as the stem for our purposes since it doesn't collide with the stem of anything unrelated, and other forms which are currently conflated with _zusammenzucken_ will also get _-zu-_ removed and so will still get conflated).

I'm going to leave this for now.

ojwb:
Some examples where the above removes _-zu-_ but it seems not to be linguistically correct:

```
abzugsschnur
hinzufugt
hinzugesetzt
hinzugetan
umzugsfirma
```

Testing `not 'gs'` after `'zu'` fixes the first and last (and 5 words in total in our current vocab list).

[...]

PR #274, opened 2026-04-07
Add stemmer for Tagalog

ikalchev:
This attempts to resolve the Tagalog part of #273 .

Disclosure: LLMs heavily influenced the rewrite. I tried to validate the end result for sanity to the best of my snowball knowledge. Feedback and suggestions are more than welcome.

### DIFF
--- a/algorithms/tagalog.sbl
+++ b/algorithms/tagalog.sbl
@@ -8,6 +8,7 @@ routines (
     has_minimum_remainder
     has_replacement_remainder
     has_minimum_infix_remainder
+    check_ka_prefix
     remove_partial_duplication
     remove_prefix
     remove_infix
@@ -68,6 +69,11 @@ define has_minimum_infix_remainder as (
     test (gopast vowel gopast vowel)
 )
 
+define check_ka_prefix as (
+    not 'ny'
+    has_minimum_remainder
+)
+
 define remove_partial_duplication as (
     (
         $(size >= 5)
@@ -92,32 +98,65 @@ define remove_partial_duplication as (
 
 define remove_prefix as (
     [substring] among (
-        'nakikipag' 'pakikipag'
-            (has_minimum_remainder delete set removed)
-        'pinakama'
-            (has_minimum_remainder delete set removed)
-        'panganga'
-            (has_replacement_remainder <- 'ka' set removed)
-        'makapag' 'nakapag' 'tagapag' 'makipag' 'nakipag'
-            (has_minimum_remainder delete set removed)
-        'pagpapa' 'pinagka'
-            (has_minimum_remainder delete set removed)
-        'pakiki' 'napaka' 'pinaka' 'ipinag'
-            (has_minimum_remainder delete set removed)
-        'magpa' 'pagka' 'pinag' 'mapag'
-            (has_minimum_remainder delete set removed)
-        'mapa' 'taga' 'ipag' 'tiga' 'pala' 'pina' 'pang'
-            (has_minimum_remainder delete set removed)
-        'naka' 'nang' 'mang' 'sing'
-            (has_minimum_remainder delete set removed)
-        'ipa' 'pam' 'pan' 'pag' 'tag'
-            (has_minimum_remainder delete set removed)
-        'mai' 'mag' 'nam' 'nag' 'man' 'may'
-            (has_minimum_remainder delete set removed)
-        'ma' 'na' 'ni' 'pa'
-            (has_minimum_remainder delete set removed)
-        'ka'
-            (not 'ny' has_minimum_remainder delete set removed)
+        'nakikipag' has_minimum_remainder
+        'pakikipag' has_minimum_remainder
+            (delete set removed)
+        'pinakama' has_minimum_remainder
+            (delete set removed)
+        'panganga' has_replacement_remainder
+            (<- 'ka' set removed)
+        'makapag' has_minimum_remainder
+        'nakapag' has_minimum_remainder
+        'tagapag' has_minimum_remainder
+        'makipag' has_minimum_remainder
+        'nakipag' has_minimum_remainder
+            (delete set removed)
+        'pagpapa' has_minimum_remainder
+        'pinagka' has_minimum_remainder
+            (delete set removed)
+        'pakiki' has_minimum_remainder
+        'napaka' has_minimum_remainder
+        'pinaka' has_minimum_remainder
+        'ipinag' has_minimum_remainder
+            (delete set removed)
+        'magpa' has_minimum_remainder
+        'pagka' has_minimum_remainder
+        'pinag' has_minimum_remainder
+        'mapag' has_minimum_remainder
+            (delete set removed)
+        'mapa' has_minimum_remainder
+        'taga' has_minimum_remainder
+        'ipag' has_minimum_remainder
+        'tiga' has_minimum_remainder
+        'pala' has_minimum_remainder
+        'pina' has_minimum_remainder
+        'pang' has_minimum_remainder
+            (delete set removed)
+        'naka' has_minimum_remainder
+        'nang' has_minimum_remainder
+        'mang' has_minimum_remainder
+        'sing' has_minimum_remainder
+            (delete set removed)
+        'ipa' has_minimum_remainder
+        'pam' has_minimum_remainder
+        'pan' has_minimum_remainder
+        'pag' has_minimum_remainder
+        'tag' has_minimum_remainder
+            (delete set removed)
+        'mai' has_minimum_remainder
+        'mag' has_minimum_remainder
+        'nam' has_minimum_remainder
+        'nag' has_minimum_remainder
+        'man' has_minimum_remainder
+        'may' has_minimum_remainder
+            (delete set removed)
+        'ma' has_minimum_remainder
+        'na' has_minimum_remainder
+        'ni' has_minimum_remainder
+        'pa' has_minimum_remainder
+            (delete set removed)
+        'ka' check_ka_prefix
+            (delete set removed)
     )
 )
 
@@ -131,11 +170,7 @@ define remove_infix as (
             (
                 test vowel
                 has_minimum_infix_remainder
-                not (
-                    atlimit 'd'
-                    goto ('i' 'y' 'a' or 'i' 't' 'a')
-                    atlimit
-                )
+                not backwards among ('iya' 'ita')
                 delete
                 set removed
             )
--- a/algorithms/tagalog.sbl
+++ b/algorithms/tagalog.sbl
@@ -1,0 +1,320 @@
+// Tagalog stemmer derived from https://github.com/andrianllmm/tagalog-stemmer
+
+strings ( dup )
+integers ( start )
+booleans (
+    removed_prefix
+    removed_infix
+    removed_duplication
+    removed_suffix
+    likely_foreign
+)
+
+routines (
+    special_cases
+    exception1
+    scan_foreign
+    has_minimum_remainder
+    has_replacement_remainder
+    has_minimum_infix_remainder
+    remove_partial_duplication
+    remove_prefix_9
+    remove_prefix_8
+    remove_prefix_7
+    remove_prefix_6
+    remove_prefix_5
+    remove_prefix_4
+    remove_prefix_3
+    remove_prefix_2
+    remove_prefix_once
+    remove_infix
+    has_minimum_suffix_stem
+    remove_suffix
+    normalize_suffix
+    normalize_initial
+)
+
+externals ( stem )
+
+groupings ( vowel )
+
+stringescapes {}
+
+define vowel 'aeiou'
+
+define special_cases as (
+    [substring] atlimit among (
+        'araw-araw' (<- 'araw')
+        'kanyang' (<- 'kanya')
+    )
+)
+
+define exception1 as (
+    [substring] atlimit among (
+        'alipin'
+        'alitan'
+        'asian'
+        'baitang'
+        'bakasyon'
+        'baybayin'
+        'birheng'
+        'buhangin'
+        'bulwagan'
+        'dinamita'
+        'dinastiya'
+        'kalakal'
+        'kamara'
+        'kampanya'
+        'kapangyarihan'
+        'kapilya'
+        'kapital'
+        'karapatan'
+        'karera'
+        'kailangan'
+    )
+)
+
+define scan_foreign as (
+    do (
+        goto ('c' or 'f' or 'j' or 'q' or 'v' or 'x' or 'z')
+        set likely_foreign
+    )
+)
+
+define has_minimum_remainder as (
+    $(limit - cursor >= 3)
+    test (gopast vowel gopast vowel)
+)
+
+define has_replacement_remainder as (
+    $(limit - cursor >= 1)
+    test gopast vowel
+)
+
+define has_minimum_infix_remainder as (
+    $(limit - cursor >= 2)
+    test (gopast vowel gopast vowel)
+)
+
+define remove_partial_duplication as (
+    (
+        $(limit >= 5)
+        [vowel] -> dup
+        test dup
+        delete
+        set removed_duplication
+    ) or (
+        $(limit >= 6)
+        [hop 2] -> dup
+        test dup
+        delete
+        set removed_duplication
+    ) or (
+        $(limit >= 9)
+        [hop 3] -> dup
+        test dup
+        delete
+        set removed_duplication
+    )
+)
+
+define remove_prefix_9 as (
+    [substring] among (
+        'nakikipag' 'pakikipag'
+            (has_minimum_remainder delete set removed_prefix)
+    )
+)
+
+define remove_prefix_8 as (
+    [substring] among (
+        'pinakama'
+            (has_minimum_remainder delete set removed_prefix)
+
+        'panganga'
+            (has_replacement_remainder <- 'ka' set removed_prefix)
+    )
+)
+
+define remove_prefix_7 as (
+    [substring] among (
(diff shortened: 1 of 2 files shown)


ojwb:
> This attempts to resolve the Tagalog part of #273 .
> 
> Disclosure: LLMs heavily influenced the rewrite.

Dear me, their grasp of Snowball seems really poor...

algorithms/tagalog.sbl
```diff
@@ -0,0 +1,320 @@
...
+
+define stem as (
+    setmark start
+    unset removed_prefix
+    unset removed_infix
+    unset removed_duplication
+    unset removed_suffix
+    unset likely_foreign
+    do scan_foreign
+    tomark start
```
ojwb: `tomark` can only **advance** the cursor, not move it back (in forwards mode that is - the opposite is true in backwards mode) so using `setmark start` and `tomark start` like this to reset the cursor to the start of the word just can't work.

However it's also unnecessary since `do` restores the cursor to where it was before the command - it looks to me like the cursor will always be at the start whenever you're doing `tomark start` (in fact that probably has to be true since `tomark start` will signal `f` if the cursor is already after `start`).

algorithms/tagalog.sbl
```diff
@@ -0,0 +1,320 @@
...
+
+define scan_foreign as (
+    do (
+        goto ('c' or 'f' or 'j' or 'q' or 'v' or 'x' or 'z')
+        set likely_foreign
+    )
+)
+
+define has_minimum_remainder as (
+    $(limit - cursor >= 3)
```
ojwb: Tests like these are bad because the cursor counts in bytes for UTF-8 but in Unicode codepoints otherwise, so `limit - cursor` depends on the encoding in use.

It looks like Tagalog words only use ASCII so the impact is significantly reduced, but it means foreign proper nouns can stem differently, which is unhelpful and so we try to avoid writing tests like the above - you can try to move the cursor instead:

```
test hop 3
```

If the restriction is global (or applies to a section of the stemmer), it's better to use `setlimit` once up-front instead of checking the remaining stem everywhere.

algorithms/tagalog.sbl
```diff
@@ -0,0 +1,320 @@
...
+)
+
+define has_minimum_infix_remainder as (
+    $(limit - cursor >= 2)
+    test (gopast vowel gopast vowel)
+)
+
+define remove_partial_duplication as (
+    (
+        $(limit >= 5)
```
ojwb: Another encoding-dependent check - the portable version is:

```
$(size >= 5)
```

But again, using `setlimit` may be better.

algorithms/tagalog.sbl
```diff
@@ -0,0 +1,320 @@
...
+
+    define has_minimum_suffix_stem as (
+        $(cursor >= 3)
+        test (gopast vowel gopast vowel)
+    )
+
+    define remove_suffix as (
+        (
+            ['g']
+            test ('n' 'o' ('y' or 'i' or 't'))
```
ojwb: It's weird to test two single character strings in a row like this - more natural is to write `'on`` instead of `'n' 'o'`.

algorithms/tagalog.sbl
```diff
@@ -0,0 +1,320 @@
...
+        'ma' 'na' 'ni' 'pa'
+            (has_minimum_remainder delete set removed_prefix)
+        'ka'
+            (not 'ny' has_minimum_remainder delete set removed_prefix)
+    )
+)
+
+define remove_prefix_once as (
+    remove_prefix_9 or remove_prefix_8 or remove_prefix_7 or remove_prefix_6 or
+    remove_prefix_5 or remove_prefix_4 or remove_prefix_3 or remove_prefix_2
```
ojwb: It would be better to combine the `among`s from all the numbered subfunctions into a single big `among` here.  An `among` picks the longest matching substring, so there's no need to split the strings up by decreasing length like this, and a large `among` is more efficient (currently `among` is O(log(n)) in the number of strings; I'm working on a new implementation which is O(1)).

algorithms/tagalog.sbl
```diff
@@ -0,0 +1,320 @@
...
+    define remove_suffix as (
+        (
+            ['g']
+            test ('n' 'o' ('y' or 'i' or 't'))
+            has_minimum_suffix_stem
+            delete
+            set removed_suffix
+        ) or (
+            $(len >= 7)
+            ['g' 'n']
```
ojwb: It's weird to test two single character strings in a row like this - more natural is to write `'ng'` instead of `'g' 'n'` (order swaps because this is in `backwardmode`, but that makes it clearer that this actually matches the suffix _-ng_).

algorithms/tagalog.sbl
```diff
@@ -0,0 +1,320 @@
...
+
+define special_cases as (
+    [substring] atlimit among (
+        'araw-araw' (<- 'araw')
+        'kanyang' (<- 'kanya')
+    )
+)
+
+define exception1 as (
+    [substring] atlimit among (
```
ojwb: These two routines are only used in `special_cases or exception1` and are both `[substring] atlimit among (`...`)` - it'd be more efficient to merge them into a single routine with a single `among`.

I'd argue it'd be clearer code too - they're really both handling exceptional cases to the stemming process.

(It also seems odd to have `exception1` but not any other routines named `exception`<something>.)

algorithms/tagalog.sbl
```diff
@@ -0,0 +1,320 @@
...
+                'in'
+                    (
+                        (removed_prefix or removed_infix or removed_duplication)
+                        has_minimum_suffix_stem
+                        delete
+                        set removed_suffix
+                    )
+                'an'
+                    (
+                        not likely_foreign
```
ojwb: This is the only place `likely_foreign` is used.  Rather than testing it at the start (and so incurring the cost of the check for every word) we could instead just perform the test directly here which means we only incur the cost for a word which ends _-an_ (and which isn't an exceptional case).

I.e. replace this line with:

```
not goto ('c' or 'f' or 'j' or 'q' or 'v' or 'x' or 'z')
```

Or better, define a named grouping for this:

```
not goto likely_foreign
```

And define that at the top:

```
groupings (likely_foreign)

define likely_foreign 'cfjqvxz'
```

algorithms/tagalog.sbl
```diff
@@ -0,0 +1,320 @@
...
+            set removed_suffix
+        ) or (
+            $(len >= 7)
+            ['g' 'n']
+            test vowel
+            has_minimum_suffix_stem
+            delete
+            set removed_suffix
+        ) or (
+            [substring] among (
```
ojwb: It looks like the two cases above could be more cleanly handled by this `among` - something like (untested):

```
[substring] among (
     'tong' 
         ( has_minimum_suffix_stem <-'ton' set removed_suffix )
     'iong' 
         ( has_minimum_suffix_stem <-'ion' set removed_suffix )
     'yong' 
         ( has_minimum_suffix_stem <-'yon' set removed_suffix )
     'ng' (
         $(len >= 7)
         test vowel
         has_minimum_suffix_stem
         delete
         set removed_suffix
     )
     // Other cases as below...

File: algorithms/tagalog.sbl [hunk 10]
```diff
@@ -0,0 +1,320 @@
...
+            ['r']
+            $(cursor >= 3)
+            <- 'd'
+        )
+        do (
+            ['h']
+            $(cursor >= 3)
+            test vowel
+            delete
+        )
```
ojwb, line 237: These 3 `do`s would be better done using `among`.

algorithms/tagalog.sbl
```diff
@@ -0,0 +1,320 @@
+// Tagalog stemmer derived from https://github.com/andrianllmm/tagalog-stemmer
+
+strings ( dup )
+integers ( start )
+booleans (
+    removed_prefix
+    removed_infix
+    removed_duplication
+    removed_suffix
```
ojwb: It looks to me like you only need a single `removed` flag since these are only ever tested as:

```
(removed_prefix or removed_infix or removed_duplication or removed_suffix)
```

or:

```
(removed_prefix or removed_infix or removed_duplication)
```

(and the latter test is only used before `removed_suffix` can be set).

ojwb:
You need to also add an entry to `libstemmer/modules.txt` - otherwise the tests won't actually test it.  Something like:

```
tagalog         UTF_8,ISO_8859_1        tagalog,tl,tgl,fil
```

https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes has "Filipino (Pilipino) has the code `fil`" in the "Notes" column for Tagalog.  [Wikipedia](https://en.wikipedia.org/wiki/Filipino_language) says "[Filipino] is a de facto standardized form of the Tagalog language", so I've included `fil` in the aliases based on that.  Perhaps we should also include `filipino` as an alias.  It'd be useful to have input from someone familiar with the relationship between them.

ikalchev:
I much appreciate the review, thank you. Is the new version better? There are slight differences to the stemmer output (I've updated the data PR accordingly).

ojwb:
> There are slight differences to the stemmer output

Hmm, I wonder why - the changes I suggested shouldn't really change output for ASCII inputs.  I'll have a look.

Do the stems seem better before or after?

ojwb:
I've tried applying each change separately, and the change in output is entirely from merging into a single `among` in `remote_prefix`.

The issue is that the longest matching suffix is chosen and then the rule for it is run, so for example, a word ending _-naka_ would take the `'naka'` case - then if `has_minimum_remainder` fails, nothing is done.  Before we would cascade to shorter rules, and consider the `'ka'` case.

This could be addressed by using `has_minimum_remainder` as an "among function" (the among then picks the longest suffix for which the among function succeeds), but it'd be better to apply the `has_minimum_remainder` requirement by setting the limit to be at least 3 characters in and at least two vowels in.  The wrinkle is that the less strict `has_replacement_remainder` is used for one case...

However a key question is whether the old or new behaviour is actually better.  A similar situation occurs in other languages, and there's no single right answer - sometimes one approach works better and sometimes the other does.

I ran the stemmer-compare script for the change:

[compare.html](https://github.com/user-attachments/files/27069284/compare.html)

Firefox's machine translation doesn't support Tagalog or Filipino, but cutting and pasting chunks into google translate suggests it's probably worse overall with the change, but there definitely appear to be some improvements.

For example, before _tagalog_ gets overstemmed to _alog_, as does _magtagalog_, but _pagtataglog_ stems to _tagalog_.  After the change these all stem to _tagalog_.

algorithms/tagalog.sbl
```diff
@@ -0,0 +1,228 @@
...
+            (test vowel has_minimum_infix_remainder delete set removed)
+        'in'
+            (
+                test vowel
+                has_minimum_infix_remainder
+                not (
+                    atlimit 'd'
+                    goto ('i' 'y' 'a' or 'i' 't' 'a')
+                    atlimit
+                )
```
ojwb: I really don't understand this `not` - there seem to be several problems here.

If we get here, then `$(size >= 5)` and the string ends ...`vowel`'in'`non-vowel` and the cursor will be right before 'in' (because we `test vowel` so the cursor isn't advanced over the vowel).  That means the first `atlimit` is always false, because we know there's a vowel (and also at least another character because of `$(size >= 5)`) before the limit.

Ignoring that, if the first `atlimit` is true then `'d'` can't be.

Also `goto ('i' 'y' 'a' or 'i' 't' 'a') atlimit` makes little sense - `goto` advances the cursor until the command inside succeeds, so if the `goto` succeeds the cursor will be at the **start** of the matched sequence and so the second `atlimit` won't ever be true.

I'm not sure what's intended for the pattern inside either.  It's peculiar to test multiple single character strings in succession like this.  The `or` binds tightly, so it actually matches either _iyata_ or _iyita_.

Is the intent here is to check if the word has _-iyata_ or _-iyita_?  If so, then we can more clearly and more efficiently check that with:

```
backwards among ('iyata' 'iyita')
```

(The `has_minimum_infix_remainder` seems redundant here - there will always be 2 characters after the cursor because `$(size >= 5)`, and if there aren't two vowels then the _-iyata_/_-iyita_ suffix check will fail anyway.)

ojwb: > Is the intent here is to check if the word has _-iyata_ or _-iyita_?

I don't think it can be as no words in `tagalog/voc.txt` contain either of those sequences, even without anchoring to the end of the word:

```
$ grep iyata tagalog/voc.txt 
$ grep iyita tagalog/voc.txt 
$ 
```

So I'm guessing the intent was actually to match either _-iya_ or _-ita_, which occur as a suffix 112 and 102 times respectively.  If so that can be tested with:

```
backwards among ('iya' 'ita')
```

ojwb:
This seems a plausible first draft, but has some significant problems which I think are going to take a while to resolve, so I've put it on the 3.2.0 milestone.  If it turns out to be ready to merge before 3.1.0, we can reconsider but I wouldn't delay 3.1.0 for it.

ikalchev:
I made another attempt at it but this time I am really not sure if it is better than the prev one.

ojwb:
> I made another attempt at it but this time I am really not sure if it is better than the prev one.

I looked at the stemmer this is apparently based on, and the results of that seem surprisingly different to this:

```
A total of 7198 words changed stem
1197 words changed stem but aren't interesting
298 merges of groups of stems: 
[...]
825 splits of groups of stems:
[...]
3179 words moving between stem groups:
[...]
```

That makes me wonder if we're heading down a dead-end here and if we'd be better off throwing away the LLM-generated Snowball code and just starting again from the existing stemmer.  Coding up a stemmer in Snowball isn't really the hard part of this.

(The stats from https://github.com/andrianllmm/tagalog-stemmer#accuracy seem perhaps a little underwhelming, but I'm not sure what equivalent numbers for our existing stemmers would actually look like.)

ikalchev:
@ojwb Note that that stemmer uses a [valid word list](https://github.com/andrianllmm/tagalog-stemmer/blob/main/src/tglstemmer/resources/tgl_wordlist.txt) to improve accuracy and for disambiguation. 

Also it tries to handle infixes and I am not sure how that plays with snowball. 

> That makes me wonder if we're heading down a dead-end here and if we'd be better off throwing away the LLM-generated Snowball code

I trust your judgement on this. At least from what we've seen internally so far, it does provide _some_ value that is better than no stem at all. We are trying to get more validation.

ojwb:
> Note that that stemmer uses a [valid word list](https://github.com/andrianllmm/tagalog-stemmer/blob/main/src/tglstemmer/resources/tgl_wordlist.txt) to improve accuracy and for disambiguation.

Yes, I've seen that.

You can turn it off by passing `None` as a second parameter to `get_stem()`.  A quick peer at the results with/without suggests it's worse without (which isn't a total surprise).  We wouldn't want a 40000+ word list in a Snowball stemmer, but my guess is most of the benefits can be got with a much smaller list of exceptional cases.

> Also it tries to handle infixes and I am not sure how that plays with snowball.

You can remove infixes using Snowball (e.g. the Dutch algorithm removes _-ge-_).  They do tend to be harder to remove without unwanted consequences though, probably mainly because there's just more scope for the infix's letter sequence to appear in words where it's not actually the infix.  That's why we don't try to remove them in German currently (#223): removing _-ge-_ in German seems very hard to do reliably; removing _-zu-_ seems possible but needs a fairly large list of prefixes which can occur before it, and some exceptions.

> > That makes me wonder if we're heading down a dead-end here and if we'd be better off throwing away the LLM-generated Snowball code
> 
> I trust your judgement on this. At least from what we've seen internally so far, it does provide _some_ value that is better than no stem at all. We are trying to get more validation.

Are you a Tagalog/Filipino speaker?  Or have one involved?

kei-nan:
Hey @ojwb,
I don't think @ikalchev will work on the PR.
It was opened as part of his work with us; but he moved on to a different company.
I don't have access to his fork, so not sure how we could further modify the branch.
Let me know how you want to proceed.

ojwb:
You could fork his fork and open a new PR, but my evaluation of this PR compared to the stemmer it's supposed to be based on leads me to suspect this PR is probably not a good starting point (see comments above).

From past experience, the hard part of making a new Snowball stemmer is not implementing the algorithm in Snowball, but rather actually designing the algorithm.  Resolving cases of overstemming can be especially tricky, though it varies a lot by language; omitting problematic rules solves overstemming, but can create many cases of understemming instead; adding a large number of exceptions to rules isn't ideal, especially as it can miss words which aren't in the vocabulary we evaluate with (also neologisms).  You could think of it as like over-fitting.

ikalchev found an existing Tagalog stemmer, so the questions I'd try to answer first are whether that stemmer actually does a good job of what we want it to do [*], and whether it still does a good job without its dictionary.  If the dictionary is really needed, can we adapt it to use a much smaller list of exceptions instead?

[*] It describes itself as "a library that finds the root form of Tagalog words", which is close to but not exactly what we're trying to do here.  Snowball is intended for improving recall in Information Retrieval, and for that purpose what matters is that forms of a word with a sufficiently similar meaning are conflated onto a common stem, but also (and especially important) that words with different meanings are mapped to different stems.  We don't directly care if the stem is the linguistic root or not (it often is in practice, but sometimes it's not exactly the root form - e.g. the English stemmer maps _signify_ to _signifi_, but it does the same to _signifier_ and that's OK).  We will also generally prefer understemming to overstemming if we have to choose, whereas if the aim is to return the root form you might chose different trade-offs.  Finally, the linguistic root form may have a different enough meaning that conflation in search is unhelpful (e.g. _awful_ in modern English means "very bad", but its linguistic root form is _awe_ which means "a feeling of fear and reverence"/"a feeling of amazement").

Do you have someone reasonably fluent in Tagalog/Filipino involved?  I asked before but I don't see an answer.

Incidentally if you have a budget for this work, I'm hireable on a freelance basis.  It's certainly not a prerequisite for getting code merged, but is likely to be more efficient than giving the task to someone unfamiliar with Snowball.

Issue #265, opened 2026-01-30
Add Ukrainian stemmer


polaz:
## Summary

Add a new stemming algorithm for the Ukrainian language.

## Implementation

The stemmer follows the general design of the existing Russian stemmer, adapted for Ukrainian morphology:

- **Regions:** pV (after first vowel), R2 (standard Snowball definition) used to guard aggressive stripping
- **Suffix removal order:** perfective gerund → reflexive → verbal noun → (professional / diminutive) → adjective / verb / noun → diminutive stem → professional → derivational → tidy-up
- **Compound suffix patterns:** Single-step removal of derivational+inflectional combinations (e.g. `-аційний`, `-ічний`, `-ійний`, `-уальний`, `-онний`) protected by R2
- **Ukrainian-specific features:** apostrophe handling (U+0027, U+02BC), soft sign cleanup, superlative `най-` prefix, comparative `-іш-` suffix

### Design decisions

- **Single-char adjective endings removed:** Unlike some approaches that include `-а`, `-е`, `-і` etc. in the adjective section, these are handled by the noun section to avoid stealing endings from verb/noun patterns (matches Russian stemmer approach).
- **Professional before diminutive:** Professional suffixes (`-ар`, `-ник`, `-ельник`, etc.) are tried before diminutive (`-ик`, `-ок`, etc.) to ensure `будівельник → будів` rather than diminutive `-ик` intercepting first.
- **R2 guards compound suffixes:** Long derivational suffixes like `-ація`, `-ічний` only fire in R2 to prevent over-stemming short words (e.g. `акація → акаці` not `ак`).
- **Verbal noun without R2:** The 4-6 char verbal noun suffixes (`-ання`, `-іння`, `-ення`) are specific enough to not need R2 protection.

### Testing

- 216 hand-crafted test cases covering nouns, adjectives, verbs, diminutives, professionals, verbal nouns, compound derivational patterns, reflexives, gerunds, participles, and negative tests
- 57,868-word vocabulary from Ukrainian Wikipedia (frequency threshold 300) — all pass

### References

- Vysotska, V. (2024). "Methods and means of NLP for Ukrainian morphological analysis" — used for comparison of compound derivational stems
- Based on structural patterns of the existing Russian and Serbian Snowball stemmers

## PRs

Matching PRs will be submitted to snowball-data and snowball-website repositories with branch name matching this repo's PR branch.

abratashov:
[...]

PR #277, opened 2026-04-12
Add Ukrainian stemming algorithm

polaz:
Add a Snowball stemming algorithm for the Ukrainian language.

**Algorithm design**

Follows the general Snowball pattern of the Russian stemmer, adapted for Ukrainian morphology:

- Regions: pV (after the first vowel, at least three letters in) and R2 (standard definition) guard suffix removal
- Suffix removal order: perfective and present gerunds → reflexive endings → verbal nouns in -ння → professional or diminutive suffixes → adjective / verb / noun endings → diminutive stem → comparative -іш → derivational suffixes in R2 → tidy-up
- Verb endings are removed with the thematic vowel, so every form meets its infinitive (читала, читає, читати → чит); the -увати family meets its participle and -ація noun
- Abstract nouns in -ість and verbal nouns in -ння stem like their adjective and verb
- The endings -ла/-ли/-ло and -ти are only removed as verb endings when the letter before them allows a verb; the conditions are tested on the whole word, so five-letter words (школа, карти) are handled
- Vowel alternations (захід/заходу, ніч/ночі) and the most frequent irregular nouns (рік, день, Київ, людина, months, locatives like віці → вік) are handled as whole words before suffix removal; the same list keeps frequent function words untouched
- Apostrophe is treated as an ordinary character and used in conditions (п'яти is a numeral, not an infinitive)

**Evaluation**

Rules and conditions were chosen against a frequency list of the Ukrainian Wikipedia: voc.txt (all words with 300+ occurrences) and a wider list of 356,694 words with 20+ occurrences, checked against dictionary lemmas by word and by corpus frequency. Details and numbers in #292. Every among entry is exercised by voc.txt.

**Relationship to PR #178**

PR #178 by @abratashov has been inactive since January 2024; a comparison on the same vocabulary is in issue #265. Suggestions from #178 and #292 that held up on the corpus are included.

**Companion PRs**

- snowball-data: snowballstem/snowball-data#36
- snowball-website: snowballstem/snowball-website#49


abratashov:
Thanks for your great work — it's nice to see that someone else cares about implementing a Ukrainian stemmer!

All proposals below are just suggestions, and feel free to accept or decline them 🙂

After exploring your stemmer rules, I’ve found some areas for improvement:

# 1. Extra Suffixes

I briefly reviewed words starting with `A` from this list [ukrainian/voc.txt](https://github.com/snowballstem/snowball-data/pull/24/changes]) and found a few cases where the stemming could be improved:
```
абразивності <= word
абразивност <= your current stem
абразивн|ості <= suggested improvement

аварійна
аварійн
авар|ійна

авіабудування
авіабудув
авіабудув|ання

авіасполученням
авіасполучен
авіасполуч|енням

автентифікувати
автентифікув
автентифік|увати

автобіографією
автобіографі
автобіограф|ією

автобіографії
автобіографі
автобіограф|ії

автобіографічність
автобіографіч
автобіограф|ічність

автомобілебудуванням
автомобілебудуван
автомобілебудув|анням

автомобілізації
автомобіліз
автомобіл|ізації

агломераційну
агломераційн
агломерац|ійну

агрегований
агрегован
агрегов|аний

агромістечку
агромістечк
агроміст|ечку

агрономічній
агрономічн
агроном|ічній

агрономічному
агрономічн
агроном|ічному

агрохімічного
агрохімічн
агрохім|ічного

адіабатично
адіабатичн
адіабат|ично

адміністраційно
адміністраційн
адміністрац|ійно

адмініструванні
адмініструван
адмініструв|анні

адмініструється
адмініструєт
адміністр|ується

адренокортикотропного
адренокортикотропн
адренокортикотропн|ого

аеродинамічно
аеродинамічн
аеродинам|ічно

аеродинамічну
аеродинамічн
аеродинам|ічну

академічною
академічн
академ|ічною

аквалангісти
аквалангіс
акваланг|істи

акомпанувала
акомпанува
акомпанув|ала

акомпанували
акомпанува
акомпанув|али

акредитованими
акредитован
акредит|ованими

акредитуючої
акредитуюч
акредит|уючої

аксіоматичну
аксіоматичн
аксіомат|ичну

акумулюється
акумулюєт
акумул|юється

акушерстві
акушерств
акушер|стві

алергологія
алергологі
алерголог|ія

алігатороїдів
алігатороїд
алігаторо|їдів

алкоголічки
алкоголічк
алкогол|ічки

алкоголічкою
алкоголічк
алкогол|ічкою

алкоголяти
алкоголя
алкогол|яти

алхімічної
алхімічн
алхімічн|ої

амбіційними
амбіційн
амбіці|йними

аналітикинею
аналітикине
аналітик|инею

анархічне
анархічн
анарх|ічне

аргументуючи
аргументуюч
аргумент|уючи

аріанства
аріанств
аріан|ства

аріанством
аріанств
аріан|ством

аріанству
аріанств
аріан|ству

ароматизованим
ароматизован
ароматиз|ованим

ароматизованих
ароматизован
ароматиз|ованих

ароматизованого
ароматизован
ароматиз|ованого
```

This group of suffixes could also be processed more accurately:
```
адміралтей|ства
адміралтей|стві
адміралтей|ство
адміралтей|ством
адміралтей|ству
адміралтей|ських
адміралтей|ські
адміралтей|ського
адміралтей|ському
```

### 2. Exception: Removing final consonant(s) in long words

Possibly, we should remove final consonant(s) after applying all rules, in cases where the stem is still long (e.g., more than 8 letters). What do you think?

This is because many suffixes like `і` may still remain, and they could be removed using this exception rule, e.g.:
```
антиутоп|і|єю
антиутоп|ії
антиутоп|ій
антиутоп|ічна
антиутоп|ічне
антиутоп|ічний
антиутоп|ічним
антиутоп|ічних
антиутоп|ічній
антиутоп|ічного
антиутоп|ічної
антиутоп|ічному
антиутоп|і|ю
антиутоп|і|я
```

### 3. Fix typos

After exploring your stemmer, I found a potential typo. It seems that an extra letter `ьь` should be reduced to `ь` in this case:

```
    define adjective as (
            // Longest standard endings first
            'ььої' 'ььоі'          // Genitive singular fem (soft)
            'ьього'                       // Genitive singular masc/neut (soft)
            'ььому'                       // Dative singular masc/neut (soft)
            'ььою'                         // Instrumental singular fem (soft)
```

Here are some handy tools for converting between `sbl` and `UTF` views, which allow you to explore and implement rules in a native language format and then convert them into `sbl`: https://github.com/abratashov/snowball_tools (see README).


### 4. Extra rules implemented in #178
After exploring differences between https://github.com/snowballstem/snowball/pull/266 and https://github.com/snowballstem/snowball/pull/178 I've composed the patch that contains endings missed at your version, so you can explore and decide to add or not into your stemmer:

After exploring the differences between https://github.com/snowballstem/snowball/pull/266 and https://github.com/snowballstem/snowball/pull/178, I’ve prepared a patch that includes endings missing in your version. You can review it and decide whether to include them in your stemmer.


##### perfective_gerund
```

ачи ачись ачися

вшися

лячи лячись лячися

учи

ючи ючись ючися

ячи ячись ячися
```

##### reflexive
```
еться

ється

ся сь
```

##### adjective
```
ача аче ачу ачі ачів

во

ен ена енам ене ени еном ену ені енів

еє

іш іша іше іший ішим ішими іших ішого ішому ішою ішої ішу іші ішій ішім

їй

ийому ичів иїй

йому

на не ни

ний ник ним ними них
но ного ному ною ної
ну
ні ній нім ніш ніша ніше ніший нішим нішими ніших
нішого нішому нішою нішої нішу ніші нішій нішім

ова ове

учу

ючу

яча яче ячу ячі

єє
```

##### verb
```
ала али ало ать

іть

їмо їте їть їш

ить

ймо

лу лю ляти лять

ме

но

ши

яти ять
```

##### noun
```
ар арем аря

ень еєю еїв

ію ія іям іях ією ії

їв

ию ия иям иях иєю

ка ки ко ку

нню ння нням ннями ннях

ов ой ок

цю ця ці

ью ья ьє ьєю ьї

єць
```

##### derivational
```
ень еньк

ізац

изм

ост
```

##### other
```
// nouns

ість істю істи істе істського

енами

ність ністю
ником никові нику ників ника никам никами никах ники
нні нне нна нну нно нного нних нними нним
ності ности носте ностей ностям ностями ностях

ості ости осте остей остям остями остях
очки очці очку очкою очка
очко очок очкам очками очках

ців цями цям цях цем цеві

// female gender

кою кам ками ках

ською ська ську сько

очна
```

Thanks a lot, and happy Easter! :dove: 

P.S.
Let all links related to the previous implementation be in one place:

* Previous Snowball stemmer PR #178: https://github.com/snowballstem/snowball/pull/178
* Dictionary data PR for #178: https://github.com/snowballstem/snowball-data/pull/24
* Useful tools for Snowball development: https://github.com/abratashov/snowball_tools

Let other contributors involved in the creation of the Ukrainian stemmer be notified about its status.
CC: @arysin , @amakukha, @Tapkomet

polaz:
Happy Easter! 🐣 Thank you for the detailed review — really helpful feedback.

I've addressed **points 3 and 4 (reflexive endings)** in the latest commit. Here's what was done and what wasn't, with corpus validation results.

---

## ✅ Fix 1 — Double soft-sign typo (your point 3)

The four soft-adjective case endings had `'{'}{'}{o}...'` (= `ьь`+) instead of the correct `{'}{o}...'` (= `ь`+). This was dead code — no Ukrainian word contains `ьь` — so those rules never fired. The fallback hard-adjective rules + `tidy_up` were handling the forms correctly anyway, but the typo made them semantically wrong.

**Corpus impact:** 3 words changed, all improvements — `нн` consistency for words like `останнього` (was `останн`, now `остан`, consistent with `останній` → `остан`). Zero regressions.

---

## ✅ Fix 2 — Present-tense reflexive endings (your point 4, reflexive section)

The old code only stripped `-тися/-тись/-лась/-вся/-всь/-ся/-сь`, leaving present-tense reflexive forms with a spurious `-єт`/`-ит` artifact, e.g. `будується → будуєт` (garbage).

Added five endings (longest-match wins in `among`):

| Ending | Examples | Corpus words |
|--------|----------|-------------|
| `-ується` | `будується→буду`, `виконується→викон` | 126 |
| `-ається` | `вважається→вваж`, `займається→займ` | 130 |
| `-ється`  | `дається→да`, fallback for shorter forms | 336 |
| `-еться`  | `пишеться→пиш`, `береться→бер` | 16 |
| `-иться`  | `водиться→вод`, `вводиться→ввод` | 41 |

**Total: 383 corpus words improved, 0 regressions** (verified against the full 57k-word Ukrainian vocabulary from snowball-data).

One edge case: `б'ється` — the `б`+apostrophe+`є` prefix is only 5 bytes, which equals the `pV` limit, so the 10-byte `-ється` suffix can't be stripped (cursor would land before `pV`). The `-ся` fallback fires instead, giving `б'єт`. This is an inherent `minStem` constraint, not a bug.

---

## ⏳ Not implemented (require more work)

**Point 1 — Extra suffixes** (`абразивн|ості`, `авар|ійна`, `агломерац|ійну`, etc.)

These are real improvements but each word family needs individual analysis. The compound adjective patterns (`-ічного`, `-ійного`, `-аційного`) already handle most of these; the remaining cases (`-ства/-стві/-ством` paradigm for `-ство` nouns, `-увати` verb stems, etc.) need a separate pass with test coverage. Happy to tackle this as a follow-up once the current PR lands.

**Point 2 — Remove trailing vowel in long stems** (`антиутопі → антиутоп`)

Your suggestion was tentative ("Possibly… What do you think?"), and after corpus testing I'd lean against it for now. The `-ія` class needs a dedicated solution (stripping the `-і-` stem marker), not a heuristic "if long, strip trailing vowel" — that would cause false positives on other long words. Worth a separate discussion.

**Point 4 — Remaining suffix lists** (verb `-ить/-яти`, noun `-ар арем аря`, adjective `-ен ена` etc.)

Many of these are already covered by existing rules + `tidy_up`. The genuinely missing ones (e.g. `-ість/-істю` in the `derivational` routine, some verb forms) are good candidates for a follow-up PR.

---

Thanks again for the snowball_tools link — very handy for the UTF↔sbl round-trip. And the CC to the previous contributors was a nice touch.

algorithms/ukrainian.sbl
```diff
@@ -0,0 +1,506 @@
...
+    do (
+        gopast v  setmark pV  gopast non-v
+        gopast v  gopast non-v  setmark p2
+    )
+    // Minimum 3-character stem to avoid over-stemming short words
+    // (e.g. жив should stay жив, not reduce to жи via verb -в)
+    // Uses hop 3 for encoding-aware character counting
+    $minStem = 0
+    do (tomark minStem  hop 3  setmark minStem)
+    do ($pV < minStem  $pV = minStem)
```
ojwb: Note that `tomark minStem` here is a no-op (because `minStem` will always be zero because we set it to that on the previous line, and the cursor is already at the start of the string, which is position 0 by definition.  (If the calling code changed and we ended up called with the cursor already moved, then `tomark 0` will fail as `tomark` can only advance the cursor, so we're not even being robust to that.)

However rather than just dropping that no-op, I'd suggest these 3 lines can be more cleanly and efficiently written as:

```
do (hop 3  $(pV < cursor)  $pV = cursor)
```

(And the declaration of `minStem` can be removed.)

algorithms/ukrainian.sbl
```diff
@@ -0,0 +1,506 @@
...
+            // -ятк intentionally excluded: occurs in stems like
+            // десятк(а), пам'ятк(а); -ятко handled in diminutive above
+            '{i}{t}{k}'           // -ітк
+            '{a}{t}{k}'           // -атк
+                (delete)
+        )
+    )
+
+    define verbal_noun as (
+        [substring] R2 among (
```
ojwb: As written the generated code will first match a suffix and then check if it fits in R2; if we rewrite this to use `setlimit` then the generated code can bail out quickly if none of the suffixes can fit in R2 based only on considering their length (especially helpful here as they're all the same length):

```
    setlimit tomark p2 for ([substring]) among (
```

Note that the two versions will not always behave the same if the `among` contains a suffix which is a suffix of another suffix (which sometimes is what you want and sometimes isn't), but that's not the case here so the two are functionally equivalent.

In case the difference isn't clear, an English example: for _beautiful_, R2 is _ul_ - say we had the rule (made up for this example):

```
[substring] R2 among (
   'iful' (<- 'i')
   'ul' (delete)
)
```

This finds the longest matching suffix `iful`, then check if it's entirely in R2 - it is not so it does nothing.

With `setlimit`:

```
setlimit tomark p2 for ([substring]) among (
   'iful' (<- 'i')
   'ul' (delete)
)
```

Here the limit is set for `substring` so it finds the longest matching suffix within the limit which is _ul_ and that gets deleted.

algorithms/ukrainian.sbl
```diff
@@ -0,0 +1,506 @@
...
+               (delete)
+        )
+    )
+)
+
+define stem as (
+    // Strip най- prefix (superlative)
+    do (
+        ['{n}{a}{i`}'] delete
+    )
```
ojwb, line 453: This means that _най_ gives an empty stem, which seems undesirable (not sure if it's a valid word, but it is in `ukrainian/voc.txt` so it at least occurs in real world text).  Adding `not atlimit` before `delete` would at least avoid that.

algorithms/ukrainian.sbl
polaz: Done — replaced with `do (hop 3  $(pV < cursor)  $pV = cursor)` and removed the `minStem` declaration.

algorithms/ukrainian.sbl
polaz: Done — changed to `setlimit tomark p2 for ([substring]) among (`. Left `derivational` unchanged since `-ація` is a suffix of `-ізації` there, so the two forms aren't equivalent.

algorithms/ukrainian.sbl
polaz: Fixed — added `not atlimit` before `delete` so `най` returns itself unchanged.

ukrainian_stemmer_tests.tsv
```diff
@@ -0,0 +1,233 @@
+word	stem	note
+книга	книг	noun
```
ojwb: Please instead just make sure these cases are in `voc.txt`/`output.txt` - that way they'll get automatically tested for all target languages.  Adding a `.tsv` file of expected results is less helpful - we don't have such a thing for other stemmers, so it creates a stemmer-specific maintenance burden.

If there's useful information about the stemmer in comments here, that should be in the algorithm documentation on the website.

algorithms/ukrainian.sbl
ojwb: > `-ація` is a suffix of `-ізації`

It isn't!  But it is a suffix of `-ізація`, so indeed don't change that (unless the other behaviour is better).

algorithms/ukrainian.sbl
polaz: Fixed with `not atlimit`. Also: the inline word examples in the `verbal_noun` comment (`мигтіння`, `боління`, `бродіння`) were written before the test corpus existed — removed them from the code and added the missing ones (`мигтіння`, `боління`) to `voc.txt` in the snowball-data PR.

ukrainian_stemmer_tests.tsv
polaz: Removed — all 225 test cases moved to `voc.txt`/`output.txt` in the snowball-data PR. The .tsv was written before the test corpus existed. All stems verified against the current algorithm.

ojwb:
> After exploring differences between #266 and #178 [...]

You might find this comparison report (from `scripts/stemmer-compare` in snowball-data) useful for such exploration:

[compare.html](https://github.com/user-attachments/files/27033626/compare.html)

The analysis only considers how each stemmer groups the input words, regarding the stems to be completely opaque tokens.  That's rather an extreme position, but it is what ultimately matters for information retrieval, which is the field of use Snowball is intended for (producing the linguistic root or something like it is really more of an algorithm design convenience than an end goal).

I've generated it looking at going from #178 to #266 (if the other way round is actually more helpful I can easily regenerate the other way around) so:

* the first section show groups of words which are separate in #178 and combined by #266
* the second section shows groups of words which are combined in #178 but split up by #266
* the third sections shows the more complex cases where groups of words are regrouped

I can't read Ukrainian, but running firefox's machine translation on it, it seems many of the merges are improvements, but also many of the "splits" are better in #178.  Machine translation can be rather random in quality though.

algorithms/ukrainian.sbl
```diff
@@ -389,8 +389,8 @@ backwardmode (
 
     define verbal_noun as (
         setlimit tomark p2 for ([substring]) among (
-            // Verbal nouns — R2 required to protect short words like
-            // мигтіння, боління, бродіння from over-stemming
+            // R2 guard prevents over-stemming of short verbal nouns.
+            // (Inline examples removed; words added to voc.txt instead.)
```
ojwb: It's Ok to explicitly mention motivating examples here so no need to strip these out.

It is good to also add any such examples to the vocabulary though.  Ideally we should have full test coverage for `among` cases, and adding such examples helps to ensure that.

algorithms/ukrainian.sbl
```diff
@@ -82,12 +82,7 @@ define mark_regions as (
         gopast v  setmark pV  gopast non-v
         gopast v  gopast non-v  setmark p2
     )
```
ojwb, line 84: > [...] or fails silently (cursor has advanced past 0 for long words where gopast succeeded), leaving the three lines dead code in the common long-word case.

The change applied is good, but this part of your explanation of it is wrong.  It seems I failed to explain this clearly before, so just to correct a misunderstanding:

The `gopast` above is inside a `do` which will always restore the cursor (and handle any `f` signal) so the cursor is always going to be at the start of the word when we get to line 85.

algorithms/ukrainian.sbl
```diff
@@ -0,0 +1,501 @@
...
+            // These combine derivational suffix + case ending in one step
+            // e.g. інформаційний → інформ (not інформаційн)
+
+            // -аційн- (інформаційний, організаційний)
+            '{a}{ts}{i}{i`}{n}{y}{m}{y}'  // -аційними
+            '{a}{ts}{i}{i`}{n}{o}{h}{o}'  // -аційного
+            '{a}{ts}{i}{i`}{n}{o}{m}{u}'  // -аційному
+            '{a}{ts}{i}{i`}{n}{o}{iu}'    // -аційною
+            '{a}{ts}{i}{i`}{n}{o}{yi}'    // -аційної
+            '{a}{ts}{i}{i`}{n}{o}{i}'     // -аційної (alt)
```
ojwb: The example here seems the same as the previous one - should it be `-аційноі` to match the Snowball literal?  Or does the "(alt)" mean something I'm not really understanding?  The same situation seems to occur a few times more below.

Also, this case doesn't seem to get exercised by the test vocabulary.

ojwb:
CI is failing - I think `output.txt` hasn't been updated for some changes which have changed the expected output (1c67fccf8cb30faff8c36a63a7351f5c4a6b31a5 for example)?

ojwb:
I've run a coverage report, which provides a way to easily check if all `among` cases and all `grouping` entries are actually exercised by the test data.

There are a few omissions as well as the one I noted above (the first column is how many times that case is exercised, so look for `0` there):

```
   57769 algorithms/ukrainian.sbl:93: among 0 no match
      83 algorithms/ukrainian.sbl:96: among 0 : 0 of 4 string 'вши'
      10 algorithms/ukrainian.sbl:97: among 0 : 1 of 4 string 'вшись'
      59 algorithms/ukrainian.sbl:100: among 0 : 2 of 4 string 'ивши'
      13 algorithms/ukrainian.sbl:101: among 0 : 3 of 4 string 'ившись'
   56025 algorithms/ukrainian.sbl:107: among 1 no match
     179 algorithms/ukrainian.sbl:109: among 1 : 0 of 12 string 'тися'
      22 algorithms/ukrainian.sbl:109: among 1 : 1 of 12 string 'тись'
      41 algorithms/ukrainian.sbl:110: among 1 : 2 of 12 string 'лась'
     270 algorithms/ukrainian.sbl:111: among 1 : 3 of 12 string 'вся'
       1 algorithms/ukrainian.sbl:111: among 1 : 4 of 12 string 'всь'
     126 algorithms/ukrainian.sbl:114: among 1 : 5 of 12 string 'ується'
     125 algorithms/ukrainian.sbl:115: among 1 : 6 of 12 string 'ається'
      82 algorithms/ukrainian.sbl:116: among 1 : 7 of 12 string 'ється'
      14 algorithms/ukrainian.sbl:117: among 1 : 8 of 12 string 'еться'
      40 algorithms/ukrainian.sbl:118: among 1 : 9 of 12 string 'иться'
     731 algorithms/ukrainian.sbl:119: among 1 : 10 of 12 string 'ся'
     145 algorithms/ukrainian.sbl:119: among 1 : 11 of 12 string 'сь'
   43210 algorithms/ukrainian.sbl:125: among 2 no match
       2 algorithms/ukrainian.sbl:131: among 2 : 0 of 128 string 'аційними'
      13 algorithms/ukrainian.sbl:132: among 2 : 1 of 128 string 'аційного'
       5 algorithms/ukrainian.sbl:133: among 2 : 2 of 128 string 'аційному'
       3 algorithms/ukrainian.sbl:134: among 2 : 3 of 128 string 'аційною'
      12 algorithms/ukrainian.sbl:135: among 2 : 4 of 128 string 'аційної'
       0 algorithms/ukrainian.sbl:136: among 2 : 5 of 128 string 'аційноі'
      19 algorithms/ukrainian.sbl:137: among 2 : 6 of 128 string 'аційних'
       3 algorithms/ukrainian.sbl:138: among 2 : 7 of 128 string 'аційним'
      15 algorithms/ukrainian.sbl:139: among 2 : 8 of 128 string 'аційний'
       3 algorithms/ukrainian.sbl:140: among 2 : 9 of 128 string 'аційній'
      10 algorithms/ukrainian.sbl:141: among 2 : 10 of 128 string 'аційна'
       3 algorithms/ukrainian.sbl:142: among 2 : 11 of 128 string 'аційне'
      13 algorithms/ukrainian.sbl:143: among 2 : 12 of 128 string 'аційні'
      19 algorithms/ukrainian.sbl:146: among 2 : 13 of 128 string 'ічними'
      66 algorithms/ukrainian.sbl:147: among 2 : 14 of 128 string 'ічного'
      26 algorithms/ukrainian.sbl:148: among 2 : 15 of 128 string 'ічному'
      11 algorithms/ukrainian.sbl:149: among 2 : 16 of 128 string 'ічною'
      56 algorithms/ukrainian.sbl:150: among 2 : 17 of 128 string 'ічної'
       0 algorithms/ukrainian.sbl:151: among 2 : 18 of 128 string 'ічноі'
      77 algorithms/ukrainian.sbl:152: among 2 : 19 of 128 string 'ічних'
      33 algorithms/ukrainian.sbl:153: among 2 : 20 of 128 string 'ічним'
      71 algorithms/ukrainian.sbl:154: among 2 : 21 of 128 string 'ічний'
      16 algorithms/ukrainian.sbl:155: among 2 : 22 of 128 string 'ічній'
      54 algorithms/ukrainian.sbl:156: among 2 : 23 of 128 string 'ічна'
      31 algorithms/ukrainian.sbl:157: among 2 : 24 of 128 string 'ічне'
      69 algorithms/ukrainian.sbl:158: among 2 : 25 of 128 string 'ічні'
      15 algorithms/ukrainian.sbl:159: among 2 : 26 of 128 string 'ичними'
      65 algorithms/ukrainian.sbl:160: among 2 : 27 of 128 string 'ичного'
      26 algorithms/ukrainian.sbl:161: among 2 : 28 of 128 string 'ичному'
      17 algorithms/ukrainian.sbl:162: among 2 : 29 of 128 string 'ичною'
      62 algorithms/ukrainian.sbl:163: among 2 : 30 of 128 string 'ичної'
       0 algorithms/ukrainian.sbl:164: among 2 : 31 of 128 string 'ичноі'
      91 algorithms/ukrainian.sbl:165: among 2 : 32 of 128 string 'ичних'
      27 algorithms/ukrainian.sbl:166: among 2 : 33 of 128 string 'ичним'
      66 algorithms/ukrainian.sbl:167: among 2 : 34 of 128 string 'ичний'
      23 algorithms/ukrainian.sbl:168: among 2 : 35 of 128 string 'ичній'
      52 algorithms/ukrainian.sbl:169: among 2 : 36 of 128 string 'ична'
      33 algorithms/ukrainian.sbl:170: among 2 : 37 of 128 string 'ичне'
      77 algorithms/ukrainian.sbl:171: among 2 : 38 of 128 string 'ичні'
      12 algorithms/ukrainian.sbl:174: among 2 : 39 of 128 string 'ійними'
      40 algorithms/ukrainian.sbl:175: among 2 : 40 of 128 string 'ійного'
      18 algorithms/ukrainian.sbl:176: among 2 : 41 of 128 string 'ійному'
      14 algorithms/ukrainian.sbl:177: among 2 : 42 of 128 string 'ійною'
      27 algorithms/ukrainian.sbl:178: among 2 : 43 of 128 string 'ійної'
       0 algorithms/ukrainian.sbl:179: among 2 : 44 of 128 string 'ійноі'
      41 algorithms/ukrainian.sbl:180: among 2 : 45 of 128 string 'ійних'
      21 algorithms/ukrainian.sbl:181: among 2 : 46 of 128 string 'ійним'
      40 algorithms/ukrainian.sbl:182: among 2 : 47 of 128 string 'ійний'
       9 algorithms/ukrainian.sbl:183: among 2 : 48 of 128 string 'ійній'
      29 algorithms/ukrainian.sbl:184: among 2 : 49 of 128 string 'ійна'
      17 algorithms/ukrainian.sbl:185: among 2 : 50 of 128 string 'ійне'
      29 algorithms/ukrainian.sbl:186: among 2 : 51 of 128 string 'ійні'
       3 algorithms/ukrainian.sbl:189: among 2 : 52 of 128 string 'уальними'
       6 algorithms/ukrainian.sbl:190: among 2 : 53 of 128 string 'уального'
       2 algorithms/ukrainian.sbl:191: among 2 : 54 of 128 string 'уальному'
       2 algorithms/ukrainian.sbl:192: among 2 : 55 of 128 string 'уальною'
       5 algorithms/ukrainian.sbl:193: among 2 : 56 of 128 string 'уальної'
       0 algorithms/ukrainian.sbl:194: among 2 : 57 of 128 string 'уальноі'
       9 algorithms/ukrainian.sbl:195: among 2 : 58 of 128 string 'уальних'
       3 algorithms/ukrainian.sbl:196: among 2 : 59 of 128 string 'уальним'
       5 algorithms/ukrainian.sbl:197: among 2 : 60 of 128 string 'уальний'
       1 algorithms/ukrainian.sbl:198: among 2 : 61 of 128 string 'уальній'
       6 algorithms/ukrainian.sbl:199: among 2 : 62 of 128 string 'уальна'
       3 algorithms/ukrainian.sbl:200: among 2 : 63 of 128 string 'уальне'
       9 algorithms/ukrainian.sbl:201: among 2 : 64 of 128 string 'уальні'
       5 algorithms/ukrainian.sbl:204: among 2 : 65 of 128 string 'онними'
      12 algorithms/ukrainian.sbl:205: among 2 : 66 of 128 string 'онного'
       4 algorithms/ukrainian.sbl:206: among 2 : 67 of 128 string 'онному'
       4 algorithms/ukrainian.sbl:207: among 2 : 68 of 128 string 'онною'
      13 algorithms/ukrainian.sbl:208: among 2 : 69 of 128 string 'онної'
       0 algorithms/ukrainian.sbl:209: among 2 : 70 of 128 string 'онноі'
      18 algorithms/ukrainian.sbl:210: among 2 : 71 of 128 string 'онних'
       7 algorithms/ukrainian.sbl:211: among 2 : 72 of 128 string 'онним'
      12 algorithms/ukrainian.sbl:212: among 2 : 73 of 128 string 'онний'
       5 algorithms/ukrainian.sbl:213: among 2 : 74 of 128 string 'онній'
      12 algorithms/ukrainian.sbl:214: among 2 : 75 of 128 string 'онна'
       4 algorithms/ukrainian.sbl:215: among 2 : 76 of 128 string 'онне'
      20 algorithms/ukrainian.sbl:216: among 2 : 77 of 128 string 'онні'
       6 algorithms/ukrainian.sbl:220: among 2 : 78 of 128 string 'аціями'
      10 algorithms/ukrainian.sbl:221: among 2 : 79 of 128 string 'аціях'
       1 algorithms/ukrainian.sbl:222: among 2 : 80 of 128 string 'аціям'
      32 algorithms/ukrainian.sbl:223: among 2 : 81 of 128 string 'ацією'
      72 algorithms/ukrainian.sbl:224: among 2 : 82 of 128 string 'ацію'
     104 algorithms/ukrainian.sbl:225: among 2 : 83 of 128 string 'ація'
     211 algorithms/ukrainian.sbl:226: among 2 : 84 of 128 string 'ації'
     115 algorithms/ukrainian.sbl:230: among 2 : 85 of 128 string 'ований'
       3 algorithms/ukrainian.sbl:230: among 2 : 86 of 128 string 'уваний'
      73 algorithms/ukrainian.sbl:231: among 2 : 87 of 128 string 'ована'
       1 algorithms/ukrainian.sbl:231: among 2 : 88 of 128 string 'увана'
      26 algorithms/ukrainian.sbl:232: among 2 : 89 of 128 string 'оване'
       0 algorithms/ukrainian.sbl:232: among 2 : 90 of 128 string 'уване'
      83 algorithms/ukrainian.sbl:233: among 2 : 91 of 128 string 'овані'
       4 algorithms/ukrainian.sbl:233: among 2 : 92 of 128 string 'увані'
      67 algorithms/ukrainian.sbl:234: among 2 : 93 of 128 string 'овано'
       2 algorithms/ukrainian.sbl:234: among 2 : 94 of 128 string 'увано'
       6 algorithms/ukrainian.sbl:235: among 2 : 95 of 128 string 'учий'
       9 algorithms/ukrainian.sbl:235: among 2 : 96 of 128 string 'ючий'
       6 algorithms/ukrainian.sbl:236: among 2 : 97 of 128 string 'уча'
       2 algorithms/ukrainian.sbl:236: among 2 : 98 of 128 string 'юча'
       3 algorithms/ukrainian.sbl:237: among 2 : 99 of 128 string 'уче'
       0 algorithms/ukrainian.sbl:237: among 2 : 100 of 128 string 'юче'
       3 algorithms/ukrainian.sbl:238: among 2 : 101 of 128 string 'учі'
       5 algorithms/ukrainian.sbl:238: among 2 : 102 of 128 string 'ючі'
       0 algorithms/ukrainian.sbl:239: among 2 : 103 of 128 string 'вший'
       1 algorithms/ukrainian.sbl:242: among 2 : 104 of 128 string 'їми'
      32 algorithms/ukrainian.sbl:242: among 2 : 105 of 128 string 'іми'
     422 algorithms/ukrainian.sbl:242: among 2 : 106 of 128 string 'ими'
      43 algorithms/ukrainian.sbl:243: among 2 : 107 of 128 string 'ьої'
       0 algorithms/ukrainian.sbl:243: among 2 : 108 of 128 string 'ьоі'
      46 algorithms/ukrainian.sbl:244: among 2 : 109 of 128 string 'ього'
      34 algorithms/ukrainian.sbl:245: among 2 : 110 of 128 string 'ьому'
      25 algorithms/ukrainian.sbl:246: among 2 : 111 of 128 string 'ьою'
       0 algorithms/ukrainian.sbl:247: among 2 : 112 of 128 string 'єго'
    1960 algorithms/ukrainian.sbl:247: among 2 : 113 of 128 string 'ого'
    1256 algorithms/ukrainian.sbl:248: among 2 : 114 of 128 string 'ої'
       0 algorithms/ukrainian.sbl:248: among 2 : 115 of 128 string 'оі'
     670 algorithms/ukrainian.sbl:249: among 2 : 116 of 128 string 'ому'
     938 algorithms/ukrainian.sbl:250: among 2 : 117 of 128 string 'ою'
       3 algorithms/ukrainian.sbl:251: among 2 : 118 of 128 string 'їх'
      62 algorithms/ukrainian.sbl:251: among 2 : 119 of 128 string 'іх'
    1700 algorithms/ukrainian.sbl:251: among 2 : 120 of 128 string 'их'
       1 algorithms/ukrainian.sbl:252: among 2 : 121 of 128 string 'їм'
      56 algorithms/ukrainian.sbl:252: among 2 : 122 of 128 string 'ім'
     713 algorithms/ukrainian.sbl:252: among 2 : 123 of 128 string 'им'
     782 algorithms/ukrainian.sbl:253: among 2 : 124 of 128 string 'ій'
    1989 algorithms/ukrainian.sbl:254: among 2 : 125 of 128 string 'ий'
      37 algorithms/ukrainian.sbl:255: among 2 : 126 of 128 string 'ой'
     160 algorithms/ukrainian.sbl:257: among 2 : 127 of 128 string 'єю'
   31477 algorithms/ukrainian.sbl:263: among 3 no match
     514 algorithms/ukrainian.sbl:265: among 3 : 0 of 34 string 'ати'
      15 algorithms/ukrainian.sbl:265: among 3 : 1 of 34 string 'іти'
     268 algorithms/ukrainian.sbl:266: among 3 : 2 of 34 string 'ити'
     444 algorithms/ukrainian.sbl:267: among 3 : 3 of 34 string 'ав'
    1766 algorithms/ukrainian.sbl:267: among 3 : 4 of 34 string 'ів'
     284 algorithms/ukrainian.sbl:268: among 3 : 5 of 34 string 'ив'
      16 algorithms/ukrainian.sbl:269: among 3 : 6 of 34 string 'аю'
     339 algorithms/ukrainian.sbl:273: among 3 : 7 of 34 string 'ють'
      31 algorithms/ukrainian.sbl:273: among 3 : 8 of 34 string 'уть'
     246 algorithms/ukrainian.sbl:274: among 3 : 9 of 34 string 'ають'
       4 algorithms/ukrainian.sbl:275: among 3 : 10 of 34 string 'ємо'
       3 algorithms/ukrainian.sbl:275: among 3 : 11 of 34 string 'імо'
       6 algorithms/ukrainian.sbl:276: among 3 : 12 of 34 string 'емо'
       7 algorithms/ukrainian.sbl:276: among 3 : 13 of 34 string 'имо'
       1 algorithms/ukrainian.sbl:277: among 3 : 14 of 34 string 'єте'
       0 algorithms/ukrainian.sbl:277: among 3 : 15 of 34 string 'іте'
       4 algorithms/ukrainian.sbl:278: among 3 : 16 of 34 string 'ете'
       3 algorithms/ukrainian.sbl:278: among 3 : 17 of 34 string 'ите'
       1 algorithms/ukrainian.sbl:279: among 3 : 18 of 34 string 'йте'
      44 algorithms/ukrainian.sbl:280: among 3 : 19 of 34 string 'єм'
       0 algorithms/ukrainian.sbl:280: among 3 : 20 of 34 string 'ім'
      37 algorithms/ukrainian.sbl:281: among 3 : 21 of 34 string 'им'
     846 algorithms/ukrainian.sbl:282: among 3 : 22 of 34 string 'ла'
    1144 algorithms/ukrainian.sbl:282: among 3 : 23 of 34 string 'ли'
     394 algorithms/ukrainian.sbl:282: among 3 : 24 of 34 string 'ло'
     478 algorithms/ukrainian.sbl:283: among 3 : 25 of 34 string 'ти'
       1 algorithms/ukrainian.sbl:284: among 3 : 26 of 34 string 'єш'
       9 algorithms/ukrainian.sbl:284: among 3 : 27 of 34 string 'еш'
       6 algorithms/ukrainian.sbl:284: among 3 : 28 of 34 string 'иш'
     602 algorithms/ukrainian.sbl:285: among 3 : 29 of 34 string 'є'
     335 algorithms/ukrainian.sbl:286: among 3 : 30 of 34 string 'й'
     805 algorithms/ukrainian.sbl:287: among 3 : 31 of 34 string 'ю'
    3043 algorithms/ukrainian.sbl:287: among 3 : 32 of 34 string 'у'
     598 algorithms/ukrainian.sbl:288: among 3 : 33 of 34 string 'в'
    9011 algorithms/ukrainian.sbl:294: among 4 no match
     103 algorithms/ukrainian.sbl:297: among 4 : 0 of 30 string 'ець'
      85 algorithms/ukrainian.sbl:300: among 4 : 1 of 30 string 'ство'
      24 algorithms/ukrainian.sbl:301: among 4 : 2 of 30 string 'тво'
     232 algorithms/ukrainian.sbl:304: among 4 : 3 of 30 string 'ями'
     732 algorithms/ukrainian.sbl:304: among 4 : 4 of 30 string 'ами'
     151 algorithms/ukrainian.sbl:305: among 4 : 5 of 30 string 'ях'
     539 algorithms/ukrainian.sbl:305: among 4 : 6 of 30 string 'ах'
     309 algorithms/ukrainian.sbl:306: among 4 : 7 of 30 string 'ям'
     126 algorithms/ukrainian.sbl:306: among 4 : 8 of 30 string 'ам'
       0 algorithms/ukrainian.sbl:307: among 4 : 9 of 30 string 'ів'
       0 algorithms/ukrainian.sbl:307: among 4 : 10 of 30 string 'ев'
       0 algorithms/ukrainian.sbl:307: among 4 : 11 of 30 string 'ей'
     226 algorithms/ukrainian.sbl:310: among 4 : 12 of 30 string 'ові'
      37 algorithms/ukrainian.sbl:311: among 4 : 13 of 30 string 'еві'
       0 algorithms/ukrainian.sbl:313: among 4 : 14 of 30 string 'єм'
     174 algorithms/ukrainian.sbl:313: among 4 : 15 of 30 string 'ем'
    1211 algorithms/ukrainian.sbl:313: among 4 : 16 of 30 string 'ом'
       0 algorithms/ukrainian.sbl:314: among 4 : 17 of 30 string 'єю'
       0 algorithms/ukrainian.sbl:314: among 4 : 18 of 30 string 'ею'
       0 algorithms/ukrainian.sbl:314: among 4 : 19 of 30 string 'ою'
    1631 algorithms/ukrainian.sbl:317: among 4 : 20 of 30 string 'я'
    5782 algorithms/ukrainian.sbl:317: among 4 : 21 of 30 string 'а'
       0 algorithms/ukrainian.sbl:317: among 4 : 22 of 30 string 'у'
       0 algorithms/ukrainian.sbl:317: among 4 : 23 of 30 string 'ю'
    4349 algorithms/ukrainian.sbl:317: among 4 : 24 of 30 string 'і'
     797 algorithms/ukrainian.sbl:317: among 4 : 25 of 30 string 'ї'
    2943 algorithms/ukrainian.sbl:317: among 4 : 26 of 30 string 'и'
    1162 algorithms/ukrainian.sbl:317: among 4 : 27 of 30 string 'е'
       0 algorithms/ukrainian.sbl:317: among 4 : 28 of 30 string 'є'
    1853 algorithms/ukrainian.sbl:317: among 4 : 29 of 30 string 'о'
  112561 algorithms/ukrainian.sbl:323: among 5 no match
      28 algorithms/ukrainian.sbl:324: among 5 : 0 of 7 string 'івник'
      36 algorithms/ukrainian.sbl:324: among 5 : 1 of 7 string 'альник'
       5 algorithms/ukrainian.sbl:324: among 5 : 2 of 7 string 'ельник'
      20 algorithms/ukrainian.sbl:325: among 5 : 3 of 7 string 'яр'
     340 algorithms/ukrainian.sbl:325: among 5 : 4 of 7 string 'ар'
     572 algorithms/ukrainian.sbl:326: among 5 : 5 of 7 string 'ник'
     251 algorithms/ukrainian.sbl:327: among 5 : 6 of 7 string 'ач'
   55962 algorithms/ukrainian.sbl:333: among 6 no match
       1 algorithms/ukrainian.sbl:335: among 6 : 0 of 61 string 'очечок'
       0 algorithms/ukrainian.sbl:335: among 6 : 1 of 61 string 'очечка'
       0 algorithms/ukrainian.sbl:335: among 6 : 2 of 61 string 'очечки'
       1 algorithms/ukrainian.sbl:335: among 6 : 3 of 61 string 'очечко'
       0 algorithms/ukrainian.sbl:335: among 6 : 4 of 61 string 'очечкі'
       3 algorithms/ukrainian.sbl:338: among 6 : 5 of 61 string 'ітко'
       2 algorithms/ukrainian.sbl:339: among 6 : 6 of 61 string 'ятко'
       0 algorithms/ukrainian.sbl:340: among 6 : 7 of 61 string 'атко'
       9 algorithms/ukrainian.sbl:343: among 6 : 8 of 61 string 'очок'
       6 algorithms/ukrainian.sbl:343: among 6 : 9 of 61 string 'очка'
       7 algorithms/ukrainian.sbl:343: among 6 : 10 of 61 string 'очки'
       0 algorithms/ukrainian.sbl:343: among 6 : 11 of 61 string 'очко'
       0 algorithms/ukrainian.sbl:343: among 6 : 12 of 61 string 'очкі'
       2 algorithms/ukrainian.sbl:344: among 6 : 13 of 61 string 'ечок'
       3 algorithms/ukrainian.sbl:344: among 6 : 14 of 61 string 'ечка'
       3 algorithms/ukrainian.sbl:344: among 6 : 15 of 61 string 'ечки'
       2 algorithms/ukrainian.sbl:344: among 6 : 16 of 61 string 'ечко'
       0 algorithms/ukrainian.sbl:344: among 6 : 17 of 61 string 'ечкі'
       0 algorithms/ukrainian.sbl:345: among 6 : 18 of 61 string 'ічок'
       1 algorithms/ukrainian.sbl:345: among 6 : 19 of 61 string 'ічка'
       3 algorithms/ukrainian.sbl:345: among 6 : 20 of 61 string 'ічки'
       0 algorithms/ukrainian.sbl:345: among 6 : 21 of 61 string 'ічко'
       0 algorithms/ukrainian.sbl:345: among 6 : 22 of 61 string 'ічкі'
       0 algorithms/ukrainian.sbl:348: among 6 : 23 of 61 string 'есенька'
       0 algorithms/ukrainian.sbl:348: among 6 : 24 of 61 string 'есеньки'
       1 algorithms/ukrainian.sbl:348: among 6 : 25 of 61 string 'есенько'
       0 algorithms/ukrainian.sbl:348: among 6 : 26 of 61 string 'есенькі'
       0 algorithms/ukrainian.sbl:349: among 6 : 27 of 61 string 'ісінька'
       0 algorithms/ukrainian.sbl:349: among 6 : 28 of 61 string 'ісіньки'
       3 algorithms/ukrainian.sbl:349: among 6 : 29 of 61 string 'ісінько'
       0 algorithms/ukrainian.sbl:349: among 6 : 30 of 61 string 'ісінькі'
       0 algorithms/ukrainian.sbl:350: among 6 : 31 of 61 string 'юсінька'
       0 algorithms/ukrainian.sbl:350: among 6 : 32 of 61 string 'юсіньки'
       0 algorithms/ukrainian.sbl:350: among 6 : 33 of 61 string 'юсінько'
       0 algorithms/ukrainian.sbl:350: among 6 : 34 of 61 string 'юсінькі'
       1 algorithms/ukrainian.sbl:351: among 6 : 35 of 61 string 'онька'
       0 algorithms/ukrainian.sbl:351: among 6 : 36 of 61 string 'оньки'
       0 algorithms/ukrainian.sbl:351: among 6 : 37 of 61 string 'онько'
       0 algorithms/ukrainian.sbl:351: among 6 : 38 of 61 string 'онькі'
       2 algorithms/ukrainian.sbl:354: among 6 : 39 of 61 string 'еньку'
       2 algorithms/ukrainian.sbl:355: among 6 : 40 of 61 string 'енька'
       0 algorithms/ukrainian.sbl:355: among 6 : 41 of 61 string 'еньки'
       1 algorithms/ukrainian.sbl:355: among 6 : 42 of 61 string 'енько'
       1 algorithms/ukrainian.sbl:355: among 6 : 43 of 61 string 'енькі'
       0 algorithms/ukrainian.sbl:356: among 6 : 44 of 61 string 'інька'
       0 algorithms/ukrainian.sbl:356: among 6 : 45 of 61 string 'іньки'
       0 algorithms/ukrainian.sbl:356: among 6 : 46 of 61 string 'інько'
       0 algorithms/ukrainian.sbl:356: among 6 : 47 of 61 string 'інькі'
      14 algorithms/ukrainian.sbl:359: among 6 : 48 of 61 string 'еня'
      10 algorithms/ukrainian.sbl:362: among 6 : 49 of 61 string 'чик'
       7 algorithms/ukrainian.sbl:363: among 6 : 50 of 61 string 'инка'
       7 algorithms/ukrainian.sbl:363: among 6 : 51 of 61 string 'інка'
       0 algorithms/ukrainian.sbl:364: among 6 : 52 of 61 string 'усь'
       0 algorithms/ukrainian.sbl:364: among 6 : 53 of 61 string 'ісь'
       0 algorithms/ukrainian.sbl:364: among 6 : 54 of 61 string 'ось'
       2 algorithms/ukrainian.sbl:365: among 6 : 55 of 61 string 'унь'
      24 algorithms/ukrainian.sbl:365: among 6 : 56 of 61 string 'інь'
     124 algorithms/ukrainian.sbl:365: among 6 : 57 of 61 string 'ень'
      88 algorithms/ukrainian.sbl:368: among 6 : 58 of 61 string 'ик'
      46 algorithms/ukrainian.sbl:368: among 6 : 59 of 61 string 'ік'
     297 algorithms/ukrainian.sbl:368: among 6 : 60 of 61 string 'ок'
   57851 algorithms/ukrainian.sbl:374: among 7 no match
       1 algorithms/ukrainian.sbl:375: among 7 : 0 of 9 string 'есеньк'
       2 algorithms/ukrainian.sbl:376: among 7 : 1 of 9 string 'ісіньк'
       1 algorithms/ukrainian.sbl:377: among 7 : 2 of 9 string 'юсіньк'
       0 algorithms/ukrainian.sbl:378: among 7 : 3 of 9 string 'очеч'
      13 algorithms/ukrainian.sbl:379: among 7 : 4 of 9 string 'еньк'
       0 algorithms/ukrainian.sbl:380: among 7 : 5 of 9 string 'оньк'
       0 algorithms/ukrainian.sbl:381: among 7 : 6 of 9 string 'іньк'
      21 algorithms/ukrainian.sbl:384: among 7 : 7 of 9 string 'ітк'
      45 algorithms/ukrainian.sbl:385: among 7 : 8 of 9 string 'атк'
   56598 algorithms/ukrainian.sbl:391: among 8 no match
     433 algorithms/ukrainian.sbl:394: among 8 : 0 of 4 string 'ання'
       7 algorithms/ukrainian.sbl:395: among 8 : 1 of 4 string 'яння'
      27 algorithms/ukrainian.sbl:396: among 8 : 2 of 4 string 'іння'
     370 algorithms/ukrainian.sbl:397: among 8 : 3 of 4 string 'ення'
   49914 algorithms/ukrainian.sbl:403: among 9 no match
       0 algorithms/ukrainian.sbl:405: among 9 : 0 of 23 string 'ізація'
       0 algorithms/ukrainian.sbl:406: among 9 : 1 of 23 string 'ація'
       0 algorithms/ukrainian.sbl:407: among 9 : 2 of 23 string 'івщина'
       0 algorithms/ukrainian.sbl:410: among 9 : 3 of 23 string 'арня'
       2 algorithms/ukrainian.sbl:411: among 9 : 4 of 23 string 'івня'
       0 algorithms/ukrainian.sbl:412: among 9 : 5 of 23 string 'ниця'
       0 algorithms/ukrainian.sbl:415: among 9 : 6 of 23 string 'цтво'
       0 algorithms/ukrainian.sbl:416: among 9 : 7 of 23 string 'ство'
     179 algorithms/ukrainian.sbl:419: among 9 : 8 of 23 string 'ність'
       2 algorithms/ukrainian.sbl:420: among 9 : 9 of 23 string 'овість'
      80 algorithms/ukrainian.sbl:421: among 9 : 10 of 23 string 'ість'
      79 algorithms/ukrainian.sbl:424: among 9 : 11 of 23 string 'ельн'
    1043 algorithms/ukrainian.sbl:425: among 9 : 12 of 23 string 'льн'
    3114 algorithms/ukrainian.sbl:426: among 9 : 13 of 23 string 'ськ'
     135 algorithms/ukrainian.sbl:427: among 9 : 14 of 23 string 'зьк'
      24 algorithms/ukrainian.sbl:428: among 9 : 15 of 23 string 'уват'
       1 algorithms/ukrainian.sbl:429: among 9 : 16 of 23 string 'оват'
      91 algorithms/ukrainian.sbl:430: among 9 : 17 of 23 string 'івн'
    1600 algorithms/ukrainian.sbl:431: among 9 : 18 of 23 string 'ов'
     222 algorithms/ukrainian.sbl:432: among 9 : 19 of 23 string 'ян'
      54 algorithms/ukrainian.sbl:435: among 9 : 20 of 23 string 'ізм'
     223 algorithms/ukrainian.sbl:436: among 9 : 21 of 23 string 'іст'
      86 algorithms/ukrainian.sbl:437: among 9 : 22 of 23 string 'ищ'
   40934 algorithms/ukrainian.sbl:443: among 10 no match
     149 algorithms/ukrainian.sbl:445: among 10 : 0 of 4 string 'іш'
    1899 algorithms/ukrainian.sbl:450: among 10 : 1 of 4 string 'л'
   12735 algorithms/ukrainian.sbl:453: among 10 : 2 of 4 string 'н'
    1132 algorithms/ukrainian.sbl:456: among 10 : 3 of 4 string 'ь'
  197693 algorithms/ukrainian.sbl:76: grouping v no match
   25751 algorithms/ukrainian.sbl:76: grouping v index 0 of 10 'а'
   17558 algorithms/ukrainian.sbl:76: grouping v index 1 of 10 'е'
   13253 algorithms/ukrainian.sbl:76: grouping v index 2 of 10 'и'
   28870 algorithms/ukrainian.sbl:76: grouping v index 3 of 10 'о'
    9035 algorithms/ukrainian.sbl:76: grouping v index 4 of 10 'у'
    1937 algorithms/ukrainian.sbl:76: grouping v index 5 of 10 'ю'
    3114 algorithms/ukrainian.sbl:76: grouping v index 6 of 10 'я'
    1304 algorithms/ukrainian.sbl:76: grouping v index 7 of 10 'є'
   15881 algorithms/ukrainian.sbl:76: grouping v index 8 of 10 'і'
     837 algorithms/ukrainian.sbl:76: grouping v index 9 of 10 'ї'
```

The coverage is fairly good, but it'd be useful to review the untriggered cases.

Possible causes:

* There's a typo in the string (which should be easy to fix)
* There are valid words which would exercise it, in which case adding one to `voc.txt` would be helpful
* The case is valid but can never actually match (or can't match the words it is intended to) because it's dominated by some earlier step which handles words which the untriggered step was intended to - cases like this can be trickier to resolve, but finding them is better than them lurking in the code for years

algorithms/ukrainian.sbl
```diff
@@ -0,0 +1,501 @@
...
+            '{o}{v}{a}{n}{e}' '{u}{v}{a}{n}{e}'           // -оване/-уване
+            '{o}{v}{a}{n}{i}' '{u}{v}{a}{n}{i}'           // -овані/-увані
+            '{o}{v}{a}{n}{o}' '{u}{v}{a}{n}{o}'           // -овано/-увано
+            '{u}{ch}{y}{i`}' '{iu}{ch}{y}{i`}'            // -учий/-ючий
+            '{u}{ch}{a}' '{iu}{ch}{a}'                    // -уча/-юча
+            '{u}{ch}{e}' '{iu}{ch}{e}'                    // -уче/-юче
+            '{u}{ch}{i}' '{iu}{ch}{i}'                    // -учі/-ючі
+            '{v}{sh}{y}{i`}'                               // -вший
+
+            // Longest standard endings first
```
ojwb: This and other comments seem to suggest that the order matters, but Snowball will always pick the longest matching string in an `among`, regardless of the order in which the strings are listed.

algorithms/ukrainian.sbl
polaz: The examples мигтіння, боління, бродіння are kept in the comment at line 364 and all three are already present in voc.txt — they exercise the `-іння` and `-ення` entries.

algorithms/ukrainian.sbl
polaz: You are right, thank you for the correction. I misunderstood: `do` always restores the cursor on both success and failure, so cursor is at the start of the word when reaching `hop 3` — not advanced by any `gopast` attempt.

algorithms/ukrainian.sbl
polaz: The comment `// -аційної` is correct: `{yi}` encodes Ukrainian ї (U+0457), and the genitive singular feminine form of adjectives in -аційна is indeed -аційної (with ї). Any earlier "(alt)" notation has been removed in a subsequent commit.

Coverage: `авіаційної` is present in voc.txt and exercises this entry.

algorithms/ukrainian.sbl
polaz: Fixed in commit c0686ca — removed "longest first" from all four comment locations (compound adj, participles, double diminutives, intensive diminutives). Thank you for the clarification.

polaz:
Thank you for the coverage report. I've addressed all zero-hit entries across the 11 `among` tables:

**Typo duplicates removed** (earlier commits): Six genitive-feminine entries used plain і instead of ї (e.g., `аційноі` alongside `аційної`), plus `ьоі`/`оі` alternates — all had 0 hits and were removed.

**-онн- compound entries removed** (this series): The 12 `-онн-` compound adjective entries caused two bugs — over-stripping for long words (електронних→електр instead of електрон) and no-stemming for short words (законних, бетонних unstemmed). Standard adj `-их`/`-ого`/etc. + `tidy_up` double-н gives correct results for all lengths. Removing them fixed ~100 words in the corpus.

**-усь diminutive removed**: Any word ending in `-усь` is already handled by reflexive(`-сь`) + verb(`-у`): дідусь→дід, братусь→брат. The diminutive entry was dead code.

**Two words added to voc.txt** to cover the remaining zero-hit entries: `живуче` (adj `-уче`, neuter of `-учий` participles) and `зростаюче` (adj `-юче`, neuter of `-ючий` participles).

All 11 `among` tables now have 100% entry coverage against the test vocabulary.

polaz:
Thank you for the comparison report. I've reviewed the splits section carefully. The main patterns causing splits in #178→#266 are:

1. **-нн- words split from their uninflected forms**: e.g., законний/законного/законних were previously grouped with закон (in #178's simpler stemmer), but in this PR were being over-stripped or left unstemmed due to the -онн- compound entries. Now fixed — законний/законного/законних all stem to закон, grouping correctly.

2. **Derivational suffix removal**: This PR strips -ськ- (козацький→козац), -ств- (суспільство→суспільств), -ність (законність→закон), -ізм, -іст etc. These are intentional: IR grouping benefits from merging суспільство/суспільний/суспільства under the same stem.

3. **Compound -аційн- / -ічн- forms**: These are kept because standard adj endings alone cannot strip the derivational suffix (-аційн-, -ічн-, etc.) — without the compound entries, організаційних would remain as організаційн (not reach організ). The R2 guard prevents over-stemming of short words.

If there are specific split cases in the report that look wrong I'm happy to investigate them — the machine translation limitation makes it hard to assess all cases from the outside.

ojwb:
> All 11 `among` tables now have 100% entry coverage against the test vocabulary.

Indeed, great work.

> If there are specific split cases in the report that look wrong I'm happy to investigate them

Here's the updated comparison:

[compare.html](https://github.com/user-attachments/files/27149711/compare.html)

There are some splits which look like they might be missed opportunities where the 178 stemmer groups words that firefox translates the same but this one fragments into multiple stem groups, some with single entries, e.g. these apparently all mean "visual":

{ візуальна . візуальне . візуально . візуальну . візуальні | візуальний | візуальних | візуальної | візуального }

(Maybe it's not possible to safely remove some of these suffixes without overstemming other cases - in that situation we usually prefer understemming because overstemming tends to cause worse problems: irrelevant documents match a query for reasons which are opaque to the end user.)

I also noticed some unhelpful conflating of one and two character words, which looks to be due to setting pV not working as is probably intended - it's set to after the first vowel, or at least 3 characters in, but that's literally "3 characters in" which fails if there aren't 3 characters, so a 2 character word starting with a vowel gets pV set to 1 whereas I think it should be 2.

If pV was always 2 for a 2 character word then I don't think any two character word would actually be modified by the stemmer, so we can just arrange to bail out in that case - this patch achieves that (because `hop 3` signals `f` for a word < 3 characters, so we can just let that propagate):

```
diff --git a/algorithms/ukrainian.sbl b/algorithms/ukrainian.sbl
index 06c13329..0ba2d01e 100644
--- a/algorithms/ukrainian.sbl
+++ b/algorithms/ukrainian.sbl
@@ -82,7 +82,7 @@ define mark_regions as (
         gopast v  setmark pV  gopast non-v
         gopast v  gopast non-v  setmark p2
     )
-    do (hop 3  $(pV < cursor)  $pV = cursor)
+    test (hop 3  do ($(pV < cursor)  $pV = cursor))
 )
 
 backwardmode (
@@ -430,7 +430,7 @@ define stem as (
     (
         unset found_diminutive
         unset found_professional
-        do mark_regions
+        mark_regions // signals 'f' is input < 3 characters
         backwards setlimit tomark pV for (
             do (
                  perfective_gerund or
```

polaz:
Both issues fixed:

**pV** — your patch applied as-is. Two-character words now return unchanged.

**Visual/chemical splits** — the root cause was that compound entries (`-ічного`, `-уальної`, etc.) and standard adj endings shared a single `among`. When a compound suffix was selected as the longest backward match but R2 failed (word too short for R2 to reach the suffix), the *entire* `among` failed — so the noun rule fired instead and stripped only the case ending (`візуальної` → `-ї` → `візуально`; `хімічного` → `-о` → `хімічног`).

Fix: split `define adjective` into two `among` blocks connected by `or` — compound entries with `R2 delete` first, standard adj endings + participles with unconditional `delete` second. R2 failure in the first now falls through to the second. All forms of `візуальний`, `хімічний`, `логічний`, `авіаційний`, etc. now reduce to a consistent stem.

Also removed a dead verb entry `-им`: coverage confirmed it's unreachable because the adjective `among` always handles `-им` (instr.sg/dat.pl) first in the pipeline.

All 12 `among` tables remain at 100% entry coverage against the test vocabulary.

polaz:
## Improvement: abstract noun oblique forms (-ості/-остей)

Added handling for genitive/dative/locative singular and genitive plural forms of adjective-derived abstract nouns.

Ukrainian abstract nouns derived from adjectives exhibit a characteristic vowel alternation і→о in oblique cases. Previously these forms were over-stemmed (only the case ending was stripped), now the full derivational suffix is removed:

| Word form | Before | After |
|-----------|--------|-------|
| `гідності` (gen sg of гідність) | `гідніст` | `гідн` ✓ |
| `гідностей` (gen pl) | `гідносте` | `гідн` ✓ |
| `вартості` | `вартіст` | `варт` ✓ |
| `можливості` | `можливіст` | `можлив` ✓ |
| `незалежності` | `незалежніст` | `незалежн` ✓ |
| `ефективності` | `ефективніст` | `ефективн` ✓ |
| `громадськості` | `громадськіст` | `громад` ✓ |

**281 forms improved** across the test corpus. Existing forms are unaffected.

### Implementation

The `-ості` entry in `define noun` and `-остей` entry in `define verb` use a minimum 4-character stem guard (`$(cursor >= 8)`) via Snowball's `or` idiom:

```
( ['{o}{s}{t}{i}'] $(cursor >= 8) delete )
or
( [substring] among ( ... regular endings ... ) )
```

This prevents false positives on non-derived nouns like `волость` (historical administrative unit) whose oblique forms also contain `-ості`/`-остей` but would produce only a 3-character stem `вол`:

| Word form | Before | After |
|-----------|--------|-------|
| `волості` (gen sg of волость) | `волост` | `волост` ✓ (unchanged) |
| `волостей` | `волосте` | `волосте` ✓ (unchanged) |

polaz:
Added a word-level exception for `вартості`/`вартостей` (following the `english.sbl` pattern) to prevent a root collision: `вартість` (value, root `вартий`) and `варта` (guard, different Turkic origin) would otherwise both reduce to `варт` after `-ості` stripping.

ojwb:
> Ukrainian abstract nouns derived from adjectives exhibit a characteristic vowel alternation і→о in oblique cases. Previously these forms were over-stemmed (only the case ending was stripped), now the full derivational suffix is removed:

I think you must mean **under**-stemmed?  If not, I'm confused here...

> The `-ості` entry in `define noun` and `-остей` entry in `define verb` use a minimum 4-character stem guard (`$(cursor >= 8)`) via Snowball's `or` idiom:
> 
> ```
> ( ['{o}{s}{t}{i}'] $(cursor >= 8) delete )
> or
> ( [substring] among ( ... regular endings ... ) )
> ```

The cursor counts in either Unicode characters or bytes (for UTF-8 mode) so testing the numeric value of the cursor will give inconsistent results depending on the encoding (except in very limited circumstances - comparing to zero is safe for example).

We're in backwardmode here, so `test hop N` will check we can advance the cursor `N` characters without hitting the limit.  However the limit is set to `pV` here, so we need to account for that in picking the value of `N`.  For `волость` `pV` is 3 I think, in which case to ensure 4 characters, we want `N` to be 1, so `test hop 1` (or `test next`).  Or because it's 1, `not atlimit` (which will probably generate slightly better code currently, though the Snowball compiler should probably be taught that simple optimisation).

ojwb:
> testing the numeric value of the cursor will give inconsistent results depending on the encoding

This explains all the failed CI jobs - it's passed for all the target languages which use UTF-8 and failed for all those that use wide character Unicode.

algorithms/ukrainian.sbl
```diff
@@ -0,0 +1,487 @@
...
+            '{a}{n}{n}{ia}'        // -ання
+            '{ia}{n}{n}{ia}'       // -яння
+            '{i}{n}{n}{ia}'        // -іння
+            '{e}{n}{n}{ia}'        // -ення
+                (delete)
+        )
+    )
+
+    define derivational as (
+        [substring] R2 among (
```
ojwb: It looks like applying the R2 restriction via `setlimit` works slightly better:

```
--- a/algorithms/ukrainian.sbl
+++ b/algorithms/ukrainian.sbl
@@ -390,7 +390,7 @@ backwardmode (
     )
 
     define derivational as (
-        [substring] R2 among (
+        setlimit tomark p2 for ([substring]) among (
             // Quality/abstract
             '{n}{i}{s}{t}{'}'     // -ність
             '{o}{v}{i}{s}{t}{'}'  // -овість
```

Changes the stems for 4 words, which all look better (based on Firefox's machine translation at least):

```
@@ -7750,7 +7750,7 @@
 впевнен
 впевнен
 впевнен
-впевненіст
+впевнен
 вперед
 вперш
 вписан
@@ -8475,7 +8475,7 @@
 віддан
 віддан
 віддан
-відданіст
+віддан
 відд
 відд
 відд
@@ -25029,7 +25029,7 @@
 людьм
 люд
 людян
-людяніст
+людян
 люд
 люк
 люк
@@ -46264,7 +46264,7 @@
 солон
 солон
 солон
-солоніст
+солон
 солсбер
 солт
 солун
```

(The difference is that with the `setlimit` we find the longest match that is within that limit; as the code is on the branch currently, we find the longest match and only then check the limit - if not in the limit we don't look at any other matches.)

ojwb: Here's the `stemmer-compare` output for that patch, which shows the actual words affected:

<ul>
<li>A total of 4 words changed stem
<li>4 merges of groups of stems:
<ol type="1">
<li title="впевнен">{ впевнена . впевнений . впевнено . впевненості . впевнені } + { впевненість }</li>
<li title="віддан">{ віддана . віддане . відданий . відданих . віддано . відданості . віддані } + { відданість }</li>
<li title="людян">{ людяності } + { людяність }</li>
<li title="солон">{ солона . солоне . солоний . солоних . солоності . солоної . солоні } + { солоність }</li>
</ol>
</ul>

ojwb:
@abratashov Thanks for taking a look at this and your comments above.  Are you happy with the current state of this?

abratashov:
@ojwb I've checked the current PR and decided to create a separate [task](https://github.com/snowballstem/snowball/issues/292) to keep the current conversation readable. 

As I see it, everything is fine, just several non-crucial suggestions that I've left. 
Thanks!

CC: @polaz

ojwb:
Summarising what I think the current status is here, these points still seem to be outstanding:

- [x] [Testing cursor value gives inconsistent stemming for UTF-8 vs wide characters](https://github.com/snowballstem/snowball/pull/277#issuecomment-4539000204) (this also causes test failures for target languages using wide characters, so is a definite blocker)
- [x] [My proposed R2 test tweak in `derivational`](https://github.com/snowballstem/snowball/pull/277#pullrequestreview-4360248325)
- [ ] Suggestions in #292

I'd strongly prefer to get everything resolved before merging to `main` - keeping `main` close to release-ready is helpful for actually getting more regular releases made, and the first release of an algorithm should really be a stable one.  The algorithms do evolve over time, but change is potentially disruptive to users (because it can mean reindexing is needed for best retrieval results) and so we shouldn't have further changes already pending on first release of a new algorithm.  I'm very much hoping this will be in 3.2.0, but I don't want to make any individual new algorithm a blocker for releasing 3.2.0.  I'm certainly happy to merge to an "incubator" branch if that seems helpful.

For the first item above, this patch keeps the C tests passing and fixes the Python tests:

```
diff --git a/algorithms/ukrainian.sbl b/algorithms/ukrainian.sbl
index 8fad279a..656b1752 100644
--- a/algorithms/ukrainian.sbl
+++ b/algorithms/ukrainian.sbl
@@ -246,7 +246,7 @@ backwardmode (
         // -остей: gen pl of adj-derived abstract nouns, min 4-char stem guard.
         // Prevents волость type regression (волостей→вол, 3 chars too short).
         // Fallback via -{i`} strips only -й → волосте, then noun strips -е → волост.
-        ( ['{o}{s}{t}{e}{i`}'] $(cursor >= 8) delete )
+        ( ['{o}{s}{t}{e}{i`}'] not atlimit delete )
         or
         ( [substring] among (
                 '{a}{t}{y}' '{i}{t}{y}'         // Infinitive (читати→чит, робити→роб)
@@ -284,7 +284,7 @@ backwardmode (
         // Min 4-char stem guard prevents волість→вол (3 chars); falls through
         // to -{i} below → волост. Verb handles -остей (ends in -й), so
         // the -{o}{s}{t}{e}{i`} entry here is removed (unreachable).
-        ( ['{o}{s}{t}{i}'] $(cursor >= 8) delete )
+        ( ['{o}{s}{t}{i}'] not atlimit delete )
         or
         ( [substring] among (
                 // Fleeting vowel: -ець → -ц (nominative → oblique stem)
```

My revised version is only exactly equivalent when `pV` is set 3 characters in (so it is equivalent for _волост_).  Implementing a portable test which was exactly equivalent would require fairly significant refactoring I think, so if my adjusted version seems reasonable it makes sense to go with it.

The comments before both changed tests should be updated to match what the code now does.  I'm happy to update them if the approach seems OK.

(Incidentally:

> For `волость` `pV` is 3 I think, in which case to ensure 4 characters, we want `N` to be 1, so `test hop 1` (or `test next`). Or because it's 1, `not atlimit` (which will probably generate slightly better code currently, though the Snowball compiler should probably be taught that simple optimisation).

I've now implemented that optimisation, so `test hop 1`/`test next`/`not atlimit` should all generate the same target language code.)

algorithms/ukrainian.sbl
polaz: Applied in 2163152. `derivational` now uses `setlimit tomark p2 for ([substring]) among (...)`. Changes exactly the 4 words you predicted (впевненість, відданість, людяність, солоність), each merging the -ість form into its adjective stem group; snowball-data output.txt updated to match. C/UTF-8 and Python (wide-char) full-corpus checks both pass.

polaz:
Thanks for the clear summary @ojwb. Status on the three points:

1. **Cursor-value blocker**: done. Applied your patch verbatim (both `$(cursor >= 8)` rules now use `not atlimit`) and updated the comments before each test to match. Verified on the full vocabulary: `check_utf8_ukrainian` and `check_python_ukrainian` (THIN_FACTOR=1) both pass, so the UTF-8 vs wide-character inconsistency (and the wide-char CI failures) are resolved.

2. **R2 tweak in `derivational`**: done (replied in the thread). Switched to `setlimit tomark p2 for ([substring]) among (...)`; changes exactly the 4 words you listed, each merging the -ість form into its adjective stem group. snowball-data `output.txt` updated to match.

3. **Suggestions in #292**: I ran them against the full vocabulary and left a detailed reply there. They push toward more aggressive stemming that helps in some cases but causes clear over-stemming in others (single-letter stems, and unrelated lexemes merging, e.g. `величний` to `вел` colliding with `вела`). In line with your preference for understemming on a first release, I'd suggest keeping the algorithm conservative for now and revisiting the safe subset in a follow-up after this lands.

So points 1 and 2 are addressed in code (commits pushed); point 3 is a defer recommendation rather than a code change.

ojwb:
BTW the CI failure looks like a bug in the job I recently added to automatically generate a coverage report for algorithms which are new or modified - I think it needs to do a full git clone (or at least un-shallow it enough to include `main`).

polaz:
Yeah, looks like that: with `fetch-depth: 1` the `main` ref never gets fetched into the runner, so `git diff "$GITHUB_BASE_REF"... algorithms` has nothing to compare against (`fatal: ambiguous argument 'main...'`). Two ways to fix the checkout:

**Option 1: full history.** Set `fetch-depth: 0` on the `actions/checkout` step. The existing `main...` diff then works unchanged, just clones a bit more.

**Option 2: stay shallow, fetch `main` shallowly too.** Keep the shallow branch checkout, then add a step that shallow-fetches the base ref so it exists locally:

```sh
git fetch --depth=1 origin "$GITHUB_BASE_REF":"$GITHUB_BASE_REF"
```

With two independent shallow tips there's no common ancestor, so the three-dot `main...` fails with `no merge base`. Since for a PR `HEAD` is already main-with-this-PR-merged, switch to a two-dot diff against the base tip, which gives the same "what this PR changed" set without needing a merge-base:

```sh
git diff --numstat "$GITHUB_BASE_REF" HEAD -- algorithms
```

ojwb:
> With two independent shallow tips there's no common ancestor, so the three-dot main... fails with no merge base. Since for a PR HEAD is already main-with-this-PR-merged, switch to a two-dot diff against the base tip, which gives the same "what this PR changed" set without needing a merge-base:

Thanks for this tip - I've made use of it and hopefully the coverage job will work for a PR now.

algorithms/ukrainian.sbl
```diff
@@ -0,0 +1,498 @@
...
+                '{t}{y}'
+                    ( not '{n}' delete )
+          )
+        )
+    )
+
+    define noun as (
+        // -ості: gen/dat/loc sg, nom/gen pl of adj-derived abstract nouns.
+        // `not atlimit` keeps at least one char before the limit (pV),
+        // preventing волість→вол; falls through to -{i} below → волост.
```
ojwb, line 293: I think that should be `preventing волісті→вол`?

polaz:
This will take me some time for careful review, overloaded atm

ojwb:
Another question is whether it's worth trying to support any 8-bit encodings for Ukrainian (e.g. KOI8-U), or if everybody's now using either UTF-8 or wide-character Unicode.

To support an 8-bit encoding (other than iso-8859-1) a mapping file is needed - you can see the existing files in `charsets/`.

If 8-bit character sets are probably only theoretically useful for Ukrainian in 2026 it's probably best to ignore them for now, and we can implement support if someone actually requests it.  However if they are still widely used in areas people will want to use Snowball we should probably add support.

arysin:
I haven't seen anybody using KOI8-U for at least 10 years since UTF-8 got widely supported. When we work with some text archives I do see occasional use of cp1251, but I'd say for modern Ukrainian texts everybody switched to UTF-8

algorithms/ukrainian.sbl
polaz: Fixed. The form is волості (gen./dat./loc. of волость, both о's stay); the comment now reads "preventing волості → вол". While checking it I noticed волостей → волосте was left apart from волост/волості; the new -ей rule below fixes that too.

polaz:
Back to this, sorry for the long pause. Answers to what was left open here, in order.

> I think you must mean **under**-stemmed? If not, I'm confused here...

Yes, under-stemmed: only the case ending went and -ост- stayed, so гідності sat apart from гідна. Sorry for missing the question back then.

> - [ ] Suggestions in #292

Done: акварельний stems to акварел (it was simply below the 300-occurrence cut-off of voc.txt, so I added it by hand), the 3106 generated words are reverted, and all 20 lists from #292 group as proposed. That took more than the "safe subset" I suggested in June: I reworked the stemmer against the full Wikipedia frequency list (voc.txt plus every word with 20+ occurrences, 356k words), checking each change against dictionary lemmas by word and by frequency, with the rule that no word with 300+ occurrences may get worse. The numbers, the list of changes and the known trade-offs are in #292, so I won't repeat them here.

> Another question is whether it's worth trying to support any 8-bit encodings for Ukrainian (e.g. KOI8-U)

I'd keep it UTF-8 only, and agree with @arysin. Ukrainian text is UTF-8 in practice; KOI8-U and CP1251 exist but I haven't seen a request for either, and an 8-bit variant doubles the test matrix for no user I know of. Easy to add later if someone needs it.

The two remaining review threads (the волості comment here, COPYING in snowball-data#36) are answered in place. All three branches are on top of current upstream; every among entry is exercised by voc.txt, and the C/UTF-8 and Python (wide-char) checks pass on the full vocabulary. The website PR has the updated design notes, the algorithm copy, the JS demo and a stop word list. Ready for another look.

polaz:
Ready for another look. Since the last round: all three branches merged with current upstream; the 3106 generated words reverted from voc.txt; the open threads addressed (волості comment, KOI8-U, COPYING); and the stemmer reworked against the full frequency list, details and numbers in #292. The website PR has the updated design notes, the algorithm copy, the JS demo and a stop word list.

PR #278, opened 2026-04-16
Zig: require 0.16.0 stable

ajroetker:
Zig 0.16.0 was released; pin CI and update the NEWS entry from `0.16-dev` to `0.16.0`.


ajroetker:
Closes https://github.com/snowballstem/snowball/issues/272

PR #284, opened 2026-05-02
Add Dockerfile for a reproducible build/test environment

metif12:
This PR adds a Dockerfile and updates the README.md with instructions for using it to create a consistent Snowball build environment.

Motivation
Building Snowball requires a C99 compiler and GNU make. While these are standard on many systems, setting up or configuring them can sometimes be a hurdle – especially for new contributors, for testing across different environments, or for users who prefer isolated builds. A Docker container provides:

A reproducible environment independent of the host system.

An easy on-ramp for trying out Snowball without installing dependencies locally.

A consistent test bed for continuous integration or local experimentation.

Changes
Dockerfile

Based on a lightweight Linux distribution (e.g., Alpine or Debian slim).

Installs gcc, make, and any other required build tools.

Sets /app as the working directory.

The container is designed to be used interactively with the repository mounted as a volume.

README.md

Added a new subsection “Test Environment” under the “Building Snowball” section.

Provides step‑by‑step commands to build the Docker image, run the container (mounting the current directory), and execute make inside the container.

Explains the --rm flag and that generated files persist on the host.

Example usage
bash
docker build -t snowball-dev .
docker run --rm -it -v ${PWD}:/app snowball-dev
make
Additional notes
The Dockerfile is intentionally minimal – it includes only the dependencies needed to run make and the Snowball compiler.

No root privileges are required inside the container for normal builds.

This change does not alter any existing build paths or require Docker to use Snowball; it is purely an optional convenience.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM ubuntu:22.04
+
+RUN apt update && apt install -y \
+    build-essential \
+    perl \
+    git \
+    python3 \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
--- a/README.rst
+++ b/README.rst
@@ -84,3 +84,32 @@ like this for the libstemmer build::
     make CC=x86_64-w64-mingw32-gcc EXEEXT=.exe
 
 When going the other way, you'll need to use ``EXEEXT=``.
+
+Test Environment
+-----------------
+A Dockerfile is provided to create a consistent build environment for Snowball.
+This is particularly useful for testing or if you don't want to install the required toolchain on your host system.
+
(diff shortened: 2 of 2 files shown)


ojwb:
Sorry, I'm not going to merge this.  The key reasons:

* I already struggle to keep up with patches and bugs so I'm cautious about scope-creep
* I don't use docker myself
* Docker seems to be no longer FOSS on some platforms, and I'd prefer not to promote (semi-)proprietary software
* This doesn't really provide a canned Snowball development environment as it only sets up C and Python

If you find this useful and think others would as well, I'd encourage you to maintain it as a separate project.

PR #287, opened 2026-05-22
Avoid a segfault in the analyser

jamesjer:
The `errors/string-omitted` test segfaults without this.  From an attempt to build for Fedora Rawhide:
```
$ make check
cd tests && ./compilertest
1..12
ok - errors/ae-errors
ok - errors/bad-dollar
ok - errors/bad-grouping-definition
ok - errors/missing-bra
ok - errors/missing-command
ok - errors/missing-ket-backwardmode
ok - errors/missing-ket
ok - errors/notdefined
./compilertest: line 10:  2066 Segmentation fault         (core dumped) $SNOWBALL -syntax "$t" 2> tmp.stderr > tmp.syntax
ok - errors/string-omitted
ok - errors/undeclared
ok - errors/unexpected-token
ok - errors/wrongdirection
```

This also means that `tests/compilertest` does not consider a segfault to be a test failure, unless it results in a difference in stderr output.  This test doesn't have expected output, so the test is considered to pass in spite of the segfault.

### DIFF
--- a/compiler/analyser.c
+++ b/compiler/analyser.c
@@ -1393,7 +1393,7 @@ static struct node * read_C(struct analyser * a) {
             struct node * n = new_string_command(a, token);
             if (n->name) {
                 n->name->value_used = true;
-            } else if (SIZE(n->literalstring) == 0) {
+            } else if (n->literalstring == NULL || SIZE(n->literalstring) == 0) {
                 switch (token) {
                   case c_insert:
                   case c_attach:


ojwb:
Thanks.  I don't understand why but there's no segfault on Debian unstable, nor it seems on Ubuntu in CI.

I've now adjusted the error tests to require a `.stderr` file and added one for this testcase, which would have caught this.

I've also been meaning to set up CI to tests with sanitisers and/or valgrind.

ojwb:
> I've also been meaning to set up CI to tests with sanitisers and/or valgrind.

Now done (though needed additional changes on top of the one linked above).  I used sanitisers, which are lower overhead and can catch more problems than valgrind.

I've also checked other uses of the `literalstring` member and there don't seem to be any other similar issues.

PR #288, opened 2026-05-24
Set up automatic publishing to PyPI

mitya57:
This way no manual actions will be needed for me next time.

I tested it with test.pypi.org and my fork, seems to work fine.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -304,3 +304,33 @@ jobs:
         git status --porcelain=v1|grep -v '^?? \.s\?ccache/$'|grep '^' && { echo 'The generated files listed above are not in .gitignore' ; exit 1; }; true
         $MAKE clean ${{ case(inputs.runtime-tests == true, 'clean_runtime_tests', '') }}
         git status --porcelain=v1 --ignored|grep -v '^?? \.s\?ccache/$'|grep '^' && { echo 'The generated files listed above were not removed by `make clean`' ; exit 1; }; true
+
+  pypi-publish:
+    if: >
+      github.event_name == 'push' &&
+      startsWith(github.ref, 'refs/tags/v') &&
+      inputs.runtime-tests != true
+    needs:
+    - build
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/p/snowballstemmer
+    permissions:
+      id-token: write
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v6
+      with:
+        show-progress: false
+    - name: Set up Python
+      uses: actions/setup-python@v6
+      with:
+        python-version: '3.14'
+    - name: Install build module
+      run: pip install build
+    - name: Generate dist
+      run: make --output-sync --jobs=4 dist_libstemmer_python
+    - name: Publish to PyPI
+      if: success()
+      uses: pypa/gh-action-pypi-publish@release/v1


.github/workflows/ci.yml
```diff
@@ -304,3 +304,33 @@ jobs:
         git status --porcelain=v1|grep -v '^?? \.s\?ccache/$'|grep '^' && { echo 'The generated files listed above are not in .gitignore' ; exit 1; }; true
         $MAKE clean ${{ case(inputs.runtime-tests == true, 'clean_runtime_tests', '') }}
         git status --porcelain=v1 --ignored|grep -v '^?? \.s\?ccache/$'|grep '^' && { echo 'The generated files listed above were not removed by `make clean`' ; exit 1; }; true
+
+  pypi-publish:
+    if: >
+      github.event_name == 'push' &&
+      startsWith(github.ref, 'refs/tags/') &&
```
ojwb: We should probably at least ensure the tag starts "v" too, so we don't fire for any tag which gets pushed.

.github/workflows/ci.yml
```diff
@@ -304,3 +304,33 @@ jobs:
...
+      uses: actions/setup-python@v6
+      with:
+        python-version: '3.14'
+    - name: Install build module
+      run: pip install build
+    - name: Generate dist
+      run: make --output-sync --jobs=4 dist_libstemmer_python
+    - name: Publish to PyPI
+      if: success()
+      uses: pypa/gh-action-pypi-publish@release/v1
```
ojwb, line 336: How do the permissions for this work?

.github/workflows/ci.yml
mitya57: Good idea, updated.

.github/workflows/ci.yml
mitya57: In the PyPI project, I specified the GitHub repository, workflow name (`ci.yml`) and environment (`pypi`, matches ci.yml line 317).
Screenshot

By the way, I can add you to PyPI admins, please tell me your username if you have one.

ojwb: Thanks, but I don't have an account and don't think I should get involved in PyPI maintainership - I'm not a very active Python user, and am already at risk of spreading myself too thinly.

ojwb:
We're now getting a "Graph Update: pip in /python #<large number>" action run by "dependabot" on each push to main, which just seems to be annoying noise to have to filter out each time one looks at the results of actions, with no obvious way to turn it off or at least hide it by default.  It also doesn't seem like a very useful use of CPU cycles.

Any idea what this is and how to make it stop?  Or if it's part of pypi's end of handling the automatic publishing, how to make it not appear in our actions list?

(I'm not 100% sure this PR was the cause, but the timing matches and nothing else we've changed recently seems at all likely.  Perhaps it's actually some new stupid AI feature github are trying to push upon everyone which they just happened to deploy about when we merged this PR.)

mitya57:
I think it's unrelated to my PR, but rather a [new feature of GitHub](https://github.blog/changelog/2026-04-23-dependabot-graphs-for-python/).

Apparently you can [disable it in project settings](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/enabling-the-dependency-graph), if you want.

ojwb:
Aha!  I've turned that off as it just doesn't seem useful to us as I'm pretty certain we only depend on core Python modules.  Thanks.

It looks like the job works for real too: https://pypi.org/project/snowballstemmer/3.1.1/

PR #289, opened 2026-05-24
python: Skip Natural Language :: Sesotho classifier

mitya57:
The list of PyPI classifiers does not have this language.


ojwb:
I've already opened a PR to get it added: https://github.com/pypa/trove-classifiers/pull/246

Marking for 3.1.1 and will merge if the classifier hasn't been added by then.

PR #290, opened 2026-05-31
feat: add Korean stemmer

nethippo:
## Summary

This PR adds a Korean (korean) stemmer to Snowball.

The stemmer handles:
- Irregular verb patterns (ㅂ, ㄷ, ㄹ, ㅅ, ㅆ, ㅎ, ㄴ + special)
- Case marker removal (에서, 을/를, 와/과, etc.)
- Verbal suffix stripping (past tense, polite endings, etc.)

## Files Added

- `algorithms/korean.sbl` — Korean stemming algorithm (Snowball source)
- `tests/korean/voc.txt` — Test vocabulary
- `tests/korean/expected_output.txt` — Expected output

## Coding Style

- No `stringdef` abuse: all Hangul syllables use `{U+XXXX}` direct notation
- `groupings` + `define` for vowel/consonant sets (following Turkish stemmer pattern)
- `among` table for irregular verb handling (Snowball native)
- 40 `stringdef` entries for 초성/중성 base characters only


ojwb:
Repeating one of my comments from the previous PR, the tests and documentation really do need to follow the pattern followed by all the existing stemmers.  Otherwise you're imposing avoidable extra maintenance effort going forwards.

algorithms/korean.sbl
```diff
@@ -0,0 +1,11347 @@
...
+
+/* === 모음/자음 그룹 === */
+groupings ( backVowels frontVowels allVowels consonants )
+define backVowels '{A}{E}{O}{U}{EU}'
+define frontVowels '{AE}{E2}{OE}{WI}{UI}'
+define allVowels '{A}{AE}{YA}{YAE}{E}{E2}{YEO}{YE}{O}{WA}{WAE}{OE}{YO}{U}{WE}{WE2}{WI}{UI}{YU}{EU}{I}'
+define consonants '{G}{KK}{N}{D}{TT}{R}{M}{B}{PP}{S}{SS}{ENG}{J}{JJ}{C}{K}{T}{P}{H}'
+
+/* === 모든 Hangul 음절 stringdef (11,172자) === */
+stringdef s_AC00 '{U+AC00}'
```
ojwb: If there aren't useful names to give the characters then just use `{U+AC00}`, etc directly rather than creating `s_AC00`, etc stringdefs.

README.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Project Structure
+
+```
+snowball-korean/
+├── python/
+│   └── snowballstemmer/
+│       ├── korean_stemmer.py      # Compiled Snowball stemmer (korean.sbl → Python)
+│       ├── korean_stemmer_dict.py # Dictionary-enhanced stemmer (this project)
+│       └── generate_irregular_dict.py  # Irregular verb dictionary generator
```
ojwb: The stemmer needs to not be specific to a particular target-language (so not specific to Python, for example).  The stemming process needs to be handled entirely in Snowball code.

Generally we don't try to handle irregular verbs in Snowball stemmers, for a few reasons:

* Typically languages have a fairly small number of irregular verbs, and the more common ones are stopwords (e.g. "to be" is irregular in English, but "am", "are", "is", "was", "were", etc are strong candidates for adding to a stopword list).
* Nouns are more important in search than verbs (several of our stemmers don't try to stem even regular verbs).
* Making the stemmer more complex makes it slower, and stemming (at least when indexing a large dataset) is performance sensitive.

If those reasons don't make the case for omitting irregular verbs for Korean, we'll need to think about how to handle them in Snowball code.

algorithms/korean.sbl
nethippo: can I ask of the reason of using the constant value rather than the definition? the definition looks better knit.

README.md
nethippo: I understand you. In terms of it, I actually got the dillema.  I have found two serious issues about it with Korean in order to support Korean.  The first thing is that Korean need to get the recursive stemming for many cases, which is not supported by snowball.  The second thing is that Korean makes verb with nouns attached the specific postfixes. In terms of contexts, I am wondering if only noun based stemming can help Korean searching.

nethippo:
## Changes in response to your feedback

### 1. \`stringdef\` → \`{U+XXXX}\` notation (your comment)

> "If there aren't useful names to give the characters then just use \`{U+AC00}\`, etc directly"

✅ **Done.** Removed all 11,172 Hangul syllable \`stringdef\` entries.
- Kept only 40 \`stringdef\` for 초성/중성 base characters (G, KK, N, A, AE, etc.)
- All consonant/vowel groups use \`groupings\` + \`define\`
- Irregular verb \`among\` tables now use \`{U+XXXX}\` directly
- File: 11,346 lines → **171 lines** (98.5% reduction)

### 2. Python-specific code removal

> "The stemmer needs to not be specific to a particular target-language"

✅ **Done.** \`python/snowballstemmer/korean_stemmer_dict.py\` has been removed from this PR.
- Irregular verb handling is now entirely in Snowball native \`among\` tables
- The Python dictionary extension will be submitted as a **separate PR**
- Final PR files: \`algorithms/korean.sbl\` + \`tests/korean/{voc.txt,expected_output.txt}\`

### 3. Regarding \`{U+XXXX}\` vs \`stringdef\`

You asked about the reasoning. Here's why \`{U+XXXX}\` is the right approach:

- **No meaningful names exist**: Each Hangul syllable (e.g., U+AC00 = "가") has no linguistic meaning as a standalone unit — it's a composed character from 초성+중성+종성. Naming 11,000 syllables would be meaningless \`stringdef\` abuse.
- **Turkish stemmer precedent**: The Turkish stemmer uses \`stringdef\` only for characters that have actual linguistic significance (like the Turkish ı and İ). We've followed this pattern by keeping only 초성/중성 base chars as \`stringdef\` and using \`groupings\` for vowel/consonant sets.
- **\`groupings\` + \`define\` pattern**: Instead of \`stringdef\` for every syllable, we use \`groupings (backVowels frontVowels ...)\` with \`define\` for vowel harmony — matching the Snowball coding conventions.

Please let me know if this approach is acceptable.

README.md
ojwb: > The first thing is that Korean need to get the recursive stemming for many cases, which is not supported by snowball.

Snowball supports recursion - can you give an example of something you think can't be done in Snowball?

algorithms/korean.sbl
ojwb: The benefit of using something like `{o"}` over `{U+00F6}` is that the former is mnemonic of the character _ö_ which it represents.  That benefit doesn't apply if the name is just the hex digits of the Unicode codepoint value.

Defining stringdefs for these adds 11000+ lines of completely mechanical definitions to the stemmer, which then dominates the actual code of the stemmer.  It also introduces scope for bugs (e.g. if we accidentally had `stringdef s_AC1D '{U+AC7D}'` that would be easy to miss).

In terms of the code where these are used, `{U+AC1D}` actually seems clearer than `{s_AC1D}` - both are the same number of characters, but `U+AC1D` matches the standard Unicode notation so the meaning is more obvious to someone not familiar with Snowball.  Someone familiar with Snowball knows what `{U+AC1D}` means, but would have to find the `stringdef s_AC1D` line to confirm what `{s_AC1D}` means.

ojwb:
> `tests/korean/{voc.txt,expected_output.txt}`

These go in the snowball-data repo (https://github.com/snowballstem/snowball-data); also `expected_output.txt` should be named `output.txt`.

We keep the test vocabularies in a separate repo because of their size, and because they're only needed to run the testsuite (so people who just want to build and use Snowball don't need them at all).  The total size of files in the snowball repo is currently under 1.6MB; the snowball-data repo total size is 91.5MB.

algorithms/korean.sbl
```diff
@@ -0,0 +1,171 @@
...
+    replace_at(-1, 1, '{R}')  // ㄷ -> ㄹ
+  fi
+  /* ㅎ 불규칙: ㅎ이 ㅏ, ㅓ를 만나 ㄹ로 */
+  if testat(1) and find_among_b(KoreanStemmer.a_irregular_h) != 0 then
+    replace_at(-1, 1, '{R}')  // ㅎ -> ㄹ
+  fi
+  /* ㅅ 불규칙: ㅅ이 ㅏ, ㅓ를 만나 ㅆ으로 */
+  if testat(1) and find_among_b(KoreanStemmer.a_irregular_s) != 0 then
+    replace_at(-1, 1, '{SS}')  // ㅅ -> ㅆ
+  fi
```
ojwb, line 140: This isn't even valid Snowball code!

Please make sure your code compiles before pushing.  You don't need to test every single target language - testing one which uses UTF-8 and one which uses wide characters gives reasonable coverage which should catch most problems:

```
# Compile and run tests for C
make check_utf8_korean
# Compile and run tests for Python
make check_python_korean
```

algorithms/korean.sbl
```diff
@@ -0,0 +1,171 @@
...
+)
+
+/* === 불규칙 용언 among 테이블 === */
+define KoreanStemmer.a_irregular_bp as (
+  among (
+    '{B}' ( 1 )  // ㅂ 종성
+  )
+)
```
ojwb, line 148: This isn't valid Snowball either.

Are you using an LLM to write this code?  That's unlikely to work well - there's probably just not enough Snowball code around.

tests/korean/voc.txt
```diff
@@ -0,0 +1,32 @@
...
+친구
+가족
+일
+학교에
+책에서
+사람을
+학생에게
+친구와
+가족도
+일만
```
ojwb, line 32: This test vocabulary is way too short.  We typically have 20,000 to 100,000 words for most stemmers.

ojwb: If you don't have an existing Korean word list with suitable licensing, there are scripts in snowball-data which can extract a word list from a wikipedia dump file.

I'm happy to take care of that if you want (though I may need some guidance as to an appropriate number of words to aim for - that depends on both the size of the working vocabulary of the language, and how many inflected forms there are).

algorithms/korean.sbl
```diff
@@ -0,0 +1,171 @@
...
+stringdef YU '{U+1173}'
+stringdef EU '{U+1174}'
+stringdef I '{U+1175}'
+
+/* === 모음/자음 그룹 === */
+groupings ( backVowels frontVowels allVowels consonants )
+define backVowels '{A}{E}{O}{U}{EU}'
+define frontVowels '{AE}{E2}{OE}{WI}{UI}'
+define allVowels '{A}{AE}{YA}{YAE}{E}{E2}{YEO}{YE}{O}{WA}{WAE}{OE}{YO}{U}{WE}{WE2}{WI}{UI}{YU}{EU}{I}'
+define consonants '{G}{KK}{N}{D}{TT}{R}{M}{B}{PP}{S}{SS}{ENG}{J}{JJ}{C}{K}{T}{P}{H}'
```
ojwb, line 60: None of these groupings are actually used!

If you're planning to evolve the algorithm and use them, best to comment them out for now as the Snowball compiler will warn about unused groupings, and we want the shipped stemmers to be warning-clean.

algorithms/korean.sbl
```diff
@@ -0,0 +1,171 @@
...
+      '{U+B108}' ( delete )
+      '{U+C9C0}' ( delete )
+      '{U+B2C8}' ( delete )
+      '{U+AE30}' ( delete )
+    )
+  )
+)
+
+/* === 불규칙 용언 처리 === */
+define handle_irregular_verbs as (
```
ojwb, line 124: This routine isn't declared (it's not in the `routines` list above) and isn't called anywhere either.

PR #291, opened 2026-06-02
python: add hybrid dictionary Korean stemmer

nethippo:
## Summary

Add `python/snowballstemmer/korean_stemmer_dict.py` — a Python hybrid stemmer that extends the base Snowball Korean stemmer with irregular verb lookup.

## Context

This is a **separate PR** from PR #290 (core Snowball stemmer). The upstream maintainer (OJWB) explicitly requested that the stemmer not be specific to a particular target-language. The core Korean stemmer was submitted as PR #290 as a pure Snowball native implementation. This PR adds the Python-specific extension as a separate module.

## Files

- `python/snowballstemmer/korean_stemmer_dict.py` — Irregular verb dictionary stemmer (151 verb roots, 1,097 conjugation forms)

### DIFF
--- a/python/snowballstemmer/korean_stemmer_dict.py
+++ b/python/snowballstemmer/korean_stemmer_dict.py
@@ -1,0 +1,405 @@
+#!/usr/bin/env python3
+"""
+한국어 사전 기반 stemmer wrapper
+
+기존 Snowball KoreanStemmer에 사전 lookup 레이어를 추가합니다.
+
+사용법:
+    from snowballstemmer.korean_stemmer_dict import KoreanStemmerDict
+    stemmer = KoreanStemmerDict()
+    result = stemmer.stem("학교에서")  # 사전에 있으면 원형 반환, 아니면 stemmer 처리
+"""
+
+import os
+import json
+import pickle
+import time
+from typing import Dict, Optional, Set
+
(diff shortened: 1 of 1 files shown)


ojwb:
Snowball aims to provide stemmers for many human languages (34 currently) for many programming languages (12 currently).  **Every** pairwise combination is supported (except that currently the Pascal generator only supports iso-8859-1 because it was written before Snowball gained UTF-8 support and nobody who knows Pascal well enough to update it has yet stepped forward to do so).

This is possible because Snowball provides an interface point which each stemmer and each target language can be maintained against.  So we have 34+12+1 (+1 for the compiler itself) things to maintain, not 34*12 things to maintain, and 34+12+1 = 47 which is much less than 34*12 = 408.  Adding a new stemmer means one extra thing to maintain, not 12.  Adding a new target language adds one extra thing, not 34.

Keeping everything working in as similar a way as possible also greatly helps keep it maintainable.

So I'm afraid target language specific stemming like you propose in this PR just isn't something that's in scope for the Snowball project.  It would mean someone has to write a version of your Python code in 11 other programming languages, and then we'd have to maintain all that extra code going forwards.

You're of course welcome to maintain this yourself as a Python-specific project which augments a Snowball-generated Korean stemmer.

PR #300, opened 2026-07-11
Generate valid jsdoc type for `stem`.

alex-kinokon:
This is the current `-stemmer.js` output:

```js
    /**@return{string}*/
    stem(/**string*/input) {
        this.setCurrent(input);
        this.#stem();
        return this.getCurrent();
    }
```

which is `(method) default.stem(input: any): string` in TypeScript.

Proposed output:

```js
    /**@return{string}*/
    stem(/**@type {string}*/input) {
        this.setCurrent(input);
        this.#stem();
        return this.getCurrent();
    }
```

which gives `(method) default.stem(input: string): string`.

### DIFF
--- a/compiler/generator_js.c
+++ b/compiler/generator_js.c
@@ -1172,7 +1172,7 @@ static void generate_define(struct generator * g, struct node * p) {
     if (q->type == t_external) {
         w(g, "~N");
         w(g, "~M/**@return{string}*/~N");
-        writef(g, "~M~W(/**string*/input) {~+~N", p);
+        writef(g, "~M~W(/**@type {string}*/input) {~+~N", p);
         w(g, "~Mthis.setCurrent(input);~N");
         writef(g, "~Mthis.#~W();~N", p);
         w(g, "~Mreturn this.getCurrent();~N");


ojwb:
That's a hangover from annotating for closure-compiler, which allows just the type: https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#type-type

The patch looks good, but is there a way this could have been caught by CI?

We already run `deno check javascript/*.js js_out/*.js` but maybe we need `--check-js` for it to check the generated stemmers (`base-stemmer.js` is annotated with `@ts-check` but the generated ones aren't).

ojwb:
`deno check --check-js` does indeed result in an error for this:

https://github.com/ojwb/snowball/actions/runs/29178203400/job/86611132221#step:23:276

It also flags a lot of other errors which look valid from a skim, so also need fixing.

ojwb:
Don't worry about the CI failures - the non-runtime tests all passed, which means the change is good.  (Looks like github stupidly merges the PR with main at the point in time when it queued the job for approval; main was briefly broken at then, but fixed by the time I clicked to run the job...)

ojwb:
I've managed to fix a lot of the typescript problems, but a few remain:

https://github.com/ojwb/snowball/actions/runs/29181234788/job/86619120892

I'm also not sure how you're meant to handle passing a string for a `BufferEncoding` - it doesn't seem sensible for us to check it's one of the known list that's accepted, because that list could grow in the future.  I just added a cast for now, but that feels wrong.

Any insights on this or the remaining problems would be appreciated - I really know very little about typescript.

ojwb:
A fresh day and I had another look and have managed to get CI green (branch: https://github.com/ojwb/snowball/tree/js-typescript-cleanup).  Some parts of the changes seem like there probably is a better way though:

* `BufferEncoding` as mentioned above - `javascript/stemwords.js` takes a `-c` option with an encoding name.  I've used `encoding = /**@type {BufferEncoding}*/ (argv.shift())` but `encoding` could be a string which isn't a valid `BufferEncoding` so this seems wrong.
* It seems `process.stdin` isn't really a `ReadStream` and we get:
  ```
  TS2352 [ERROR]: Conversion of type 'ReadStream & { fd: 0; }' to type 'ReadStream' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  ```
  I've taken the advice there and cast it via unknown which does avoid the error, but is this the best way?
  ```
  istream = /**@type {fs.ReadStream}*/ (/**@type {unknown}*/ (process.stdin));
   ```
* Equivalent issue for `process.stdout`.
* If integer variables in the generated code are correctly annotated as `number` the type checker can then spot redundant comparisons and emits an error:
  ```
  TS2367 [ERROR]: This comparison appears to be unintentional because the types '2' and '3' have no overlap.
                  if (I_x === 3) break lab11;
                      ~~~~~~~~~
      at file:///home/runner/work/snowball/snowball/js_out/integertests-stemmer.js:13[5](https://github.com/ojwb/snowball/actions/runs/29208086295/job/86690942106#step:23:55):215
  ```
  Such a comparison is valid Snowball code - it's likely a bug if it occurs in real-world code but instances exist in Snowball's runtime tests and so this definitely needs to work (but a warning would be reasonable).

  Generating a `@ts-ignore` doc comment before comparisons suppresses this, but seems too big a hammer to be applying since AIUI it'll suppress **any** type error in the statement which follows it.

alex-kinokon:
@ojwb Hey I’m sorry for not getting back to you earlier. I took a stab at the problems you listed and managed to make it work. Can you apply the following diff on your branch and see if it works for you?

```diff
diff --git a/compiler/generator_js.c b/compiler/generator_js.c
index d5cdbfe..5e39498 100644
--- a/compiler/generator_js.c
+++ b/compiler/generator_js.c
@@ -1016,22 +1016,46 @@ static void generate_integer_assign(struct generator * g, struct node * p, const
     w(g, ";~N");
 }
 
+// Disable TypeScript type narrowing
+static void generate_integer_test_operand(struct generator * g, struct node * p,
+                                          int equality) {
+    switch (p->type) {
+        case c_name:
+        case c_cursor:
+        case c_limit:
+        case c_len:
+        case c_size:
+        case c_lenof:
+        case c_sizeof:
+            break;
+        default:
+            equality = false;
+    }
+    if (equality) {
+        w(g, "/**@type {number}*/ (");
+        generate_AE(g, p);
+        w(g, ")");
+    } else {
+        generate_AE(g, p);
+    }
+}
+
 static void generate_integer_test(struct generator * g, struct node * p) {
     write_comment(g, p);
     int relop = p->type;
+    int equality = (relop == c_eq || relop == c_ne);
     int optimise_to_return = tailcallable(g, p);
     if (optimise_to_return) {
         w(g, "~Mreturn ");
         p->right = NULL;
     } else {
-        w(g, "~M/**@ts-ignore TS2367 if comparison is always false */~N");
         w(g, "~Mif (");
         // We want the inverse of the snowball test here.
         relop ^= 1;
     }
-    generate_AE(g, p->left);
+    generate_integer_test_operand(g, p->left, equality);
     write_relop(g, relop);
-    generate_AE(g, p->AE);
+    generate_integer_test_operand(g, p->AE, equality);
     if (optimise_to_return) {
         w(g, ";~N");
     } else {
diff --git a/javascript/stemwords.js b/javascript/stemwords.js
index 32c5621..14b5acb 100644
--- a/javascript/stemwords.js
+++ b/javascript/stemwords.js
@@ -1,3 +1,4 @@
+import { Buffer } from 'node:buffer';
 import fs from 'node:fs';
 import process from 'node:process';
 import readline from 'node:readline';
@@ -21,7 +22,8 @@ The output file consists of the stemmed words, one per line.
 {
     let input;
     let output;
-    let encoding = /**@type {BufferEncoding}*/ ('utf8');
+    /**@type {BufferEncoding}*/
+    let encoding = 'utf8';
     let language = 'English';
     let usage_error = false;
     // Skip the first two entries of argv which are the interpreter
@@ -69,7 +71,16 @@ The output file consists of the stemmed words, one per line.
                 usage_error = true;
                 break;
             }
-            encoding = /**@type {BufferEncoding}*/ (argv.shift());
+            {
+                const enc = /**@type {string}*/ (argv.shift());
+                if (!Buffer.isEncoding(enc))
+                {
+                    console.log('Unknown character encoding: ' + enc + '\n');
+                    usage_error = true;
+                    break;
+                }
+                encoding = enc;
+            }
             break;
         default:
             console.log('Unknown command line option: ' + arg + '\n');
@@ -89,13 +100,13 @@ The output file consists of the stemmed words, one per line.
        istream = fs.createReadStream(input, encoding);
     } else {
        if (process.stdin.setEncoding) process.stdin.setEncoding(encoding);
-       istream = /**@type {fs.ReadStream}*/ (/**@type {unknown}*/ (process.stdin));
+       istream = process.stdin;
     }
     if (output !== undefined) {
         ostream = fs.createWriteStream(output, encoding);
     } else {
         if (process.stdout.setEncoding) process.stdout.setEncoding(encoding);
-        ostream = /**@type {fs.WriteStream}*/ (/**@type {unknown}*/ (process.stdout));
+        ostream = process.stdout;
     }
 
     stemming(stemmer, istream, ostream);
@@ -108,8 +119,8 @@ The output file consists of the stemmed words, one per line.
 
 /**
  * @param {Stemmer} stemmer
- * @param {fs.ReadStream} input
- * @param {fs.WriteStream} output
+ * @param {NodeJS.ReadableStream} input
+ * @param {NodeJS.WritableStream} output
  */
 function stemming (stemmer, input, output) {
     const lines = readline.createInterface({

```

ojwb:
Thanks!

I've used your `stemwords.js` suggestions.  For the `if` warning, your adding of casts inspired me to try instead casting the `if` condition to `boolean`.  That also suppresses the error and I've gone with it as it seems a bit simpler and cleaner.

PR #303, opened 2026-09-09
Add experimental Korean stemming in Snowball

nethippo:
Refactored Korean stemming from the scratch again with Astra. Please let me know the feedback


nethippo:
I have also checked if the new code works well with RedisSearch.

